PR #84, opened 2020-06-16
Non-persistent LDAP connections

JeltevanBoheemen:
This replaces the persistent LDAP connection to fetch users with an on-demand connection.
I'm not 100% sure if this works correctly with the Select2 paging thing. Haven't been able to test this without ldap access.
@tymees Could you take a look?


tymees:
Code looks good, I don't really have anything to add.

I tested it on the test server, and it was working fine. So I think it's okay to merge!

PR #85, opened 2020-06-18
Feature/multi secretary

JeltevanBoheemen:
Allows for multiple users in the 'Secretaris' group. They can edit and view each-others reviews and decisions. It's hacky, but it works (I believe).

Implemented as an additional group for the primary secretary (please suggest a better name if you can think of it). This is much simpler than making an additional group for extra secretaries, as most infrastructure remains intact.

Additionally overwrites the get_secretary() email field by the general email address.

### DIFF
--- a/core/utils.py
+++ b/core/utils.py
@@ -20,7 +20,7 @@ def get_secretary():
     """
     Returns the Head secretary. We limit this to one user.
     """
-    obj = get_all_secretaries().first()
+    obj = get_user_model().objects.filter(groups__name=settings.GROUP_PRIMARY_SECRETARY).first()
     obj.email = settings.EMAIL_FROM
     return obj
 
@@ -28,7 +28,7 @@ def get_all_secretaries():
     """
     Return all users in the 'Secretary' group.
     """
-    return get_user_model().objects.filter(groups__name=settings.GROUP_PRIMARY_SECRETARY).all()
+    return get_user_model().objects.filter(groups__name=settings.GROUP_SECRETARY).all()
 
 def is_secretary(user):
     """
--- a/core/utils.py
+++ b/core/utils.py
@@ -18,16 +18,21 @@ class AvailableURL(object):
 
 def get_secretary():
     """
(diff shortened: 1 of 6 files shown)


core/utils.py
```diff
@@ -17,10 +18,23 @@ def __init__(self, title, margin=0, url=None, is_title=False):
...
     """
-    Returns the Secretary. We limit this to one user.
+    Returns the Head secretary. We limit this to one user.
     """
-    return get_user_model().objects.filter(groups__name=settings.GROUP_SECRETARY)[0]
+    obj = get_all_secretaries().first()
+    obj.email = settings.EMAIL_FROM
+    return obj
 
+def get_all_secretaries():
```
tymees, line 27: Shouldn't this filter be filtering on GROUP_SECRETARY?

From what I can gather the 'only one person' rule now applies to the GROUP_PRIMARY_SECRETARY, so this method will be the same as get_secretary, except in a QS container. 

At the least the method name is misleading at the moment.

tymees:
I haven't pulled and tested it myself yet, I'll do that after this comment.

I do have a general question first about the secretary mail override:
This will also make most mails go to the fetc e-mail instead of the personal e-mail. Does Desiree know about this change? I can imagine it's actually preferable this way, but in my experience this should be announced beforehand.

tymees:
Seems to work! Once the two questions I had are adressed it can be merged in my opinion :)

core/utils.py
JeltevanBoheemen: Yeah this makes no sense.

JeltevanBoheemen:
> I haven't pulled and tested it myself yet, I'll do that after this comment.
> 
> I do have a general question first about the secretary mail override:
> This will also make most mails go to the fetc e-mail instead of the personal e-mail. Does Desiree know about this change? I can imagine it's actually preferable this way, but in my experience this should be announced beforehand.

Good that you checked that, this was indeed requested by Desiree.

PR #101, opened 2020-12-11
Feature/route visibility

miggol:
Feature should be implemented correctly. Some tests fail because of a missing table "tasks_registration__old", which appears to be unrelated.


core/urls.py
```diff
@@ -15,5 +15,5 @@
     url(r'^user_search/$', UserSearchView.as_view(), name='user_search'),
 
     # Checks on conditional fields
-    url(r'^check_requires/$', check_requires, name='check_requires'),
+    url(r'^check_requires/$', check_requires, name='check_requires')
```
tymees, line 18: Deze comma moet er eigenlijk weer terug komen. 

We gebruiken overal trailing comma's, om git diffs duidelijk te houden

proposals/templates/proposals/proposal_list.html
```diff
@@ -118,6 +124,8 @@ <h3>{% trans "Uitleg" %}</h3>
                             <img src="{{ img_hide }}" title="{% trans 'Verberg' %}">
                         </a>
                     {% endif %}
+                    {% if proposal.ALLOW_DETAIL_VIEW %}
```
tymees, line 127: Waar is dit voor?

Kan dit weg?

tymees:
> Some tests fail because of a missing table "tasks_registration__old", which appears to be unrelated.

This is an incompatibility issue with newer SQLite versions and Django 1.11. Should be fixed when we migrate to 2.2

tymees:
Your solution is fine. 

However, I would suggest moving the code that assigns the human readable text to the route from the view code to a new method in the ``Review`` class. 

This way, we can re-use that code elsewhere later on. 

In addition, we might want to look at using ``.annotate()`` (see the Django documentation) on the QS instead. That way we could annotate the proposal object with the latest review object, and call the method I described earlier in the view instead. (And ``get_queryset`` could then actually return a QS, which is preferable).
That could be quite hard, so ignore this if you deem it so.

proposals/views/proposal_views.py
```diff
@@ -45,6 +45,21 @@ def get_queryset(self):
                                        status_review=True,
                                        in_archive=True,
                                        public=True)
+    
+    def add_route_info(self, p):
+        "Adds human-readable route info to the given proposal"
+        last_review = p.review_set.last()
+        
+        try: route = last_review.short_route
```
tymees, line 53: We prefer the try: and catch: on separate lines :)

proposals/templates/proposals/proposal_list.html
miggol: Onbedoeld restant van toen ik overwoog de detailpagina zichtbaar te maken voor supervisors. Is nu weg.

core/urls.py
miggol: Logisch!

miggol:
Comments duly noted. I've applied changes, and have added some help-text reflecting the desire to move route info into the Review class in the future. I think it's best to get this one out the door for now, and start applying annotate() on my next pass.

tymees:
Fine with me!
Merge when you can!

Could you make an issue for moving the code? Before we forget to do it....

Issue #93, opened 2020-12-03
Allow all committee members to see currently open proposals


tymees:
The best way to do this (I think) would be to clone `AllProposalReviewsView`, and change that view such that:

1. It is viewable on the basis of committee-group membership instead of secretary membership
2. It selects all Review objects that are open

The template used in that view should already restrict actions, such that only the secretary can do things. 

I would recommend checking if the same thing holds true for the details page!

miggol:
Implemented in #102

PR #102, opened 2020-12-11
Feature/members see all reviews

miggol:
Implements #93 and additionally allows committee members to view the details page of all reviews in the chamber they are a member of.

### DIFF
--- a/reviews/views.py
+++ b/reviews/views.py
@@ -13,7 +13,8 @@ from proposals.models import Proposal
 from .forms import (DecisionForm, ReviewAssignForm, ReviewCloseForm,
                     ChangeChamberForm)
 from .mixins import (AutoReviewMixin, UserAllowedMixin,
-                     CommitteeMixin, UserOrSecretaryAllowedMixin)
+                     CommitteeMixin, UserOrSecretaryAllowedMixin,
+                     UsersOrGroupsAllowedMixin)
 from .models import Decision, Review
 from .utils import notify_secretary, start_review_route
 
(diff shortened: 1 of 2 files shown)


reviews/views.py
```diff
@@ -327,12 +339,20 @@ def get_queryset(self):
...
+                       UsersOrGroupsAllowedMixin, generic.DetailView):
     """
     Shows the Decisions for a Review
     """
     model = Review
+    
+    def get_group_required(self):
+        
+        obj = self.get_object()
+        allowed_groups = [ settings.GROUP_SECRETARY ]
```
tymees, line 352: Kan dit niet beter een oneliner zijn?

``return [ settings.GROUP_SECRETARY, obj.proposal.reviewing_commitee.name ]``

miggol: Ja, en ik even de variabelenaam consistent gemaakt met de soortgelijke functies. 

Soms vind ik de eerste stijl wel lekker idiomatisch, dat je in íeder geval de secretaris toelaat en daar nog iets aan toevoegt. Maar met maar twee stappen is dat een beetje lelijk.

miggol:
Before merging I want to double-check if it's safe to assume that proposal.reviewing_commitee.name will always reference a group. I think we do that elsewhere too but maybe there's a better way I don't know of.

tymees:
A doublecheck is always good, but I think in this case it should always be the case in production.

This attribute is set on the very first page of the form, and is hardcoded against the default groups. The only reason I see that this won't happen is:
1. The groups aren't present at all (in which case we fucked up royally anyway)
2. A proposal was created outside the form (in which case we probably fucked up royally again)

PR #108, opened 2020-12-21
Bugfix for missing METC decision file

miggol:
It is currently possible to submit an METC-positive study without an METC decision PDF. When this happens, an error occurs in PDF generation and the user cannot continue. Modifying the proposal PDF template with a conditional fixes this.

### DIFF
--- a/proposals/templates/proposals/proposal_pdf.html
+++ b/proposals/templates/proposals/proposal_pdf.html
@@ -163,9 +163,14 @@ Page <pdf:pagenumber /> of <pdf:pagecount />
             <tr>
                 <th>{% get_verbose_field_name "proposals" "wmo" "metc_decision" %}</th><td>{{ wmo.metc_decision|yesno:_("ja,nee") }}</td>
             </tr>
-            {% if not proposal.is_practice %}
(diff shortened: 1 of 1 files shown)


tymees:
I'm pretty sure this fixes a symptom instead of the actual problem. 

The actual problem is that you can submit a study that needs approval from the METC without their approval. Reading through the procedures of the FETC I don't see a situation where that should be possible.

tymees:
Did some digging, and I think I see how the document uploading can be bypassed. While the form itself will block progress, you can still progress though the navigation. 

This particular bypass will be patched in the new layout update. (Navigation has been changed to save the form before progressing to the indicated page, which will block the progressing in this case). However, I do think we need a second check added to be sure.

We have a method that validates the entire proposal, which is used to block submitting if any of the forms still have errors in them. The ``WmoApplicationForm`` isn't checked however. 

If we add that form to ``_build_forms(1)`` in ``proposals/utils/validate_proposal.py``, we should patch this problem completely and in a user friendly manner.

tymees:
(Also, I think having a check in the PDF doesn't hurt, so we can keep that change)

PR #110, opened 2021-01-12
Proposal diff will now correctly compare studies. Fixes #97

### LINKED ISSUE #97: Comparing proposals with multiple trajactories don't preserve order
When you compare them, Trajectory 1 of the original proposal is trajectory 2 of the new proposal. This results in a lot of false-positives in the comparison view.

tymees:


tymees:
A classic case of 'this never worked how was this not noticed for so long'

tymees:
(Also fixed a bug in which observation studies couldn't be compared because of a faulty if statement

Issue #96, opened 2020-12-03
Make 'show in archive' conditional


tymees:
Right now, when the secretary concludes a review he/she is always asked if the proposal should be included in the archive.

This should be limited to only show the question if the proposal is accepted. 

I would suggest looking at how conditional questions are implemented elsewhere, as this can be implemented the same way.

miggol:
Fixed in #113

Issue #95, opened 2020-12-03
Add if a proposal is a revision to the mail to reviewers


tymees:
The e-mail template needs a conditional for some extra text if it's a revision. 

The code that sends out the mail can be found in `start_review_route` function in `reviews/utils.py`. 

While you're there, please add the reference number to the mail as well. I don't know why it isn't included there.

miggol:
Fixed in #113

PR #113, opened 2021-01-14
2.2/small fixes for #95 and #96

miggol:
Solves #95 and #96. Note that this is a pull without migrations, because the only thing that had to be migrated is the `related_name="children"` line in models/propoals.py that should already be in the code. I will make sure that line is not duplicated after automatically merging.


tymees:
Looks good.

One small note:
You're using translations for new/revised in the mail... Considering the mail isn't translated at all, that's not really necessary.
Additionally, making ``new_or_revised`` a boolean and using an if/else in the template would keep the text in the template.

PR #115, opened 2021-01-18
Allow a maximum of 10 trajectories. Fixes #114

### LINKED ISSUE #114: Allow more trajectories
For the first time since this portal exists, somebody needs more than 5 trajectories.

Most of the portal is written such that it can acommodate any number of studies, but the ``StudyStartForm`` is the exception.

As we have to define questions statically (for the most part), we have to add a question for every possible trajectory. So the easiest solution is to increase the amount of trajectory-name questions in ``StudyStartForm`` to 10.

tymees:


miggol (approved):
I'm still somewhat confused about the fixed amount, but I'll get there

tymees:
It's a very basic workaround for the fact that you need to define all fields in the form class. 

While you could dynamically add questions in ``__init__`` if you know how many there are, but we don't know that in this instance (as this is determined in the same form). Thus, the original programmer just defined 5 name fields and used JS to show the required number of fields. 

A better fix would be to split up the pages into 'how many trajectories do you have' and 'what are their names'.  That way you know how many names you need to collect before building the form, and dynamically add a field for every trajectory. (Or use mutli-form form, like the documents page).

(Alternatively, one could even just put the name field in the first page of the trajectory itself).

That would effectively allow an unlimited amount of trajectories, but would be more work than just copy pasting like this. Also, an unlimited number of trajectories might be a bit much work for reviewers....

mhkuu:
Original programmer here! I confirm this was (and is) a hacky workaround. Feel free to @ me if you come across more seemingly inexplainable hacks. BTW, really like the new lay-out! 

Also, 5 trajectories ought to be enough for everybody, right ;-)

tymees:
Hey Martijn, nice to see you here :P

Well 5 was enough for years, until somebody became ambitious and needed 6 ;)
(Poor, poor participants).

PR #117, opened 2021-01-21
Added paragraph in template

miggol:
This is the current state of things:

image

At first I considered only adding this paragraph if the proposal was a revision. However I feel that if a submitter got their proposal back for revision they would only see this page right when they were ready to upload it, at which point it would be of little use. The same goes for my "tip". I also fleshed out the text to fill it up.

Comments welcome


tymees (approved):

tymees:
No comments, I think this is exactly what was requested.

Merged

PR #121, opened 2021-01-25
2.2/filename generation

### LINKED ISSUE #98: Automatically rename uploaded files
Uploaded documents should be named according to the following format:
`{refnum}-{trajectory}-{type}`, where:

* refnum, the reference number of the proposal
* trajectory, the trajectory number. (Ideally prefixed with 'trajectory'). Note: some documents are regarded as 'extra' and don't have a trajectory. In which case it should say 'extra'
* type, the type of document. This should be one of the following: 
  * IB - for 'informatiebrieven' (including for the parents)
  * TV - for 'toestemmingsformulieren`
  * IB-SL - for 'informatiebrieven schoolleiding`

You can implement this using a callable, see this documention:
https://docs.djangoproject.com/en/3.1/ref/models/fields/#django.db.models.FileField.upload_to

Also, the documents aren't attached to the proposal directly, rather through an intermediate `Documents` model

miggol:
Fixes #98. According to my testing this should not affect existing files that don't match the filename schema. They simply remain as they were.


proposals/utils/proposal_utils.py
```diff
@@ -373,3 +377,60 @@ def notify_local_staff(proposal):
...
+        from proposals.models import Proposal
+        
+        if isinstance(instance, Proposal):
+            proposal = instance
+            trajectory = None
+        else:
+            # In case of Documents or Study objects
+            proposal = instance.proposal
+            try:
+                trajectory = 'T' + str(instance.study.id)
```
tymees, line 404: This should be ``instance.study.order``

proposals/utils/proposal_utils.py
```diff
@@ -373,3 +377,60 @@ def notify_local_staff(proposal):
...
+        refnum = proposal.reference_number
+        
+        extension = '.' + original_fn.split('.')[-1][-7:] # At most 7 chars seems reasonable
+        
+        fn_parts = [ p for p in ['FETC',
+                                 refnum,
+                                 lastname,
+                                 trajectory,
+                                 self.document_type,
+                                 ] if p != None or '' ]
```
tymees, line 418: Isn't it better to filter out None's instead?

proposals/utils/proposal_utils.py
tymees: (ID results in large numbers that don't make sense for the reviewer, as it's the PK. ``order`` is the trajectory number specific for this proposal)

miggol: Ah yes, thank you for that. Locally I only had at most a couple proposals going and therefore didn't notice

proposals/utils/proposal_utils.py
miggol: Some Django fields will return an empty string yet not be equivalent to None. I have now used filter() because it's more obvious than a comprehension.

proposals/utils/proposal_utils.py
```diff
@@ -373,3 +377,69 @@ def notify_local_staff(proposal):
...
+        if isinstance(instance, Proposal):
+            proposal = instance
+            trajectory = None
+        else:
+            # In case of Documents or Study objects
+            proposal = instance.proposal
+            try:
+                trajectory = 'T' + str(instance.study.order)
+            except AttributeError:
+                trajectory = None
```
tymees, line 406: Just realised, if a researcher adds more than 1 extra document of the same type, this won't give both of them an unique name.

tymees: As for a fix, maybe do a QS of extra document objects and see which index this object is in that QS?

miggol: Good tip. I probably should have caught this, too.

tymees: That's why I wanted to do these reviews. I make these kinds of oversights all the time...

PR #131, opened 2021-02-02
Target UiL Core v1.1 stable + updated dependencies

tymees:
This change makes sure we target the stable version of UiL Django Core v1.1.

There are no differences between this version and the old beta-11 version, but as we're releasing, I prefer to target the stable tag instead of the beta one.

In addition, other depenendencies have received a version bumb. Should have no breaking changes


tymees:
Eh, hold off on merging this for now.

Figured out a bug in this version.

tymees:
Okay, we should be good now.

As an additional bonus, now the datepicker looks more in line with the rest of the app

miggol (approved):

miggol:
Can't wait for that new datepicker

PR #134, opened 2021-02-05
Django 2.2

miggol:
It has finally come to this. A django-1.11 branch has been made to keep a working copy of the old version.


tymees:
>  A django-1.11 branch has been made to keep a working copy of the old version.

Wouldn't a git tag be a better way to do this?

miggol:
I have changed the common.yaml of im-linux-etcl to pull from the new django-1.11 branch.

tymees:
Fair point

Still would like a tag, so we can remove that branch when we decommission the old servers

miggol:
> Fair point
> 
> Still would like a tag, so we can remove that branch when we decommission the old servers

What do you mean exactly by this? A tag on the dango-1.11 branch to remind us to remove it?

tymees:
> What do you mean exactly by this? A tag on the dango-1.11 branch to remind us to remove it?

A tag on master before we pull in the django 2.2 code. Otherwise the tag would be removed if the django-1.11 branch is removed

PR #138, opened 2021-02-15
Merge tst into main django-2.2 branch

### LINKED ISSUE #136: List of proposals should be sorted newest first
As in the title. Letting users choose what  the list sorts by would of course be even better, but the default should definitely be newest first rather than oldest first, as it is now.

miggol:
This fixes #136 and updates some content and translations. Also a new banner image from the UU beeldbank is included.


miggol:
Just realized that this should be `proposal__date_confirmed` not `date_created`, will fix

miggol:
It was already different in 03bfde0 ;)

But I had accidentally ordered a poor quality version from the beeldbank, this one is nice and sharp

tymees:
Oh, should've looked closer then. Oops

PR #140, opened 2021-02-16
Fixed an issue in which proposals concluded as 'needs revision' aren'…

tymees:
…t closed properly


tymees:
Derp

miggol:
Big whoops. But good on you for fixing it so quickly!

Issue #125, opened 2021-01-28
High task count hangs server


miggol:
It is currently possible to enter a very high amount of tasks or sessions in the study form. Doing so hangs the server until it can process this ridiculous SQL request.

Might be better in production with an actual SQL server than with SQLite, but still a nasty bug.

tymees:
Do you have the debug toolbar enabled?

That has a tendency of adding 300% overhead on SQL queries, on top of the innefficencies of SQLite. 

That being said, optimizing the querysets isn't a bad idea. Often you can reduce the time it takes by simply instructing Django to prefetch data in the initial query. (Normally it's queried when needed, but that increases the amount of queries exponantially).

miggol:
Good to know! I don't really know how I would optimize the querysets. My plan was to just limit the amount of tasks that can be added to a proposal. Something reasonable like 20 million. (I kid, but there's actually no limit now)

tymees:
Or we could just trust our researchers to not go overboard and subject the participants to ~~torture~~ insanely complicated tasks

miggol:
Fixed in develop

miggol:
Fixed in develop

PR #147, opened 2021-02-26
Fix/content and translations

### LINKED ISSUE #146: HTML is escaped in form.as_table descriptions
Thought this would be a quick fix but I'm kind of stuck, so I'm paging @tymees for this one. An example can be seen `ProposalUpdate`:

> Zijn er nog andere onderzoekers bij deze studie betrokken die \<strong>niet\</strong> geaffilieerd zijn aan een van de onderzoeksinstituten van de Faculteit Geestwetenschappen van de UU? :

The text itself is defined in the model, where `Django.utils.safestring.mark_safe()` has no effect on it, either inside or outside of the translation hook. Using the safe filter in the template also has no effect, meaning this string is explicitly escaped somewhere inbetween. Any ideas?

Update: tried `mark_safe_lazy()` as well to no avail

miggol:
Fixes #146, and also #125 because I inadvertently left that commit in.


proposals/models.py
```diff
@@ -7,6 +7,10 @@
 from django.db import models
 from django.utils.translation import ugettext_lazy as _
 
+from django.utils.functional import lazy
+from django.utils.safestring import mark_safe
+mark_safe_lazy = lazy(mark_safe, str)
```
tymees, line 12: Any reason we need to use the lazy mark_safe here?

studies/forms.py
```diff
@@ -279,3 +279,18 @@ def __init__(self, *args, **kwargs):
         """
         super(SessionStartForm, self).__init__(*args, **kwargs)
         self.fields['sessions_number'].required = True
+    
+    def clean(self):
```
tymees, line 283: Wouldn't it be better to use a ``MaxValueValidator`` (either  in the form or on the actual model field itself)

tasks/forms.py
```diff
@@ -82,7 +85,19 @@ def clean(self):
             # Prevent double required errors
             if 'tasks_number' not in self.errors:
                 self.add_error('tasks_number', forms.ValidationError(_('Dit veld is verplicht.'), code='required'))
-
+
```
tymees, line 88: Same comment as ``sessions_number``

studies/forms.py
miggol: Yes, it would. I've moved this requirement to the model now with a validator.

proposals/models.py
miggol: We used mark_safe_lazy with lazy translations elsewhere, so I was under under the impression that it was unwise to mix the reqular mark_safe with lazy translation. But I can't find specific references to this in the Django docs now that I'm looking for it.

We tend to use ugettext_lazy everywhere, I suppose for convenience, even when we don't strictly need it. So I thought this would be fine too. But marking safe has security implications so I'm fine with either.

tymees: I tend to use mark_safe over mark_safe_lazy, mostly because I'm lazy and using mark_safe is easier. I suppose it doesn't really matter, I was just wondering. 

And year, ugettext_lazy is preferable imo as it almost always works, which can't be said for the non-lazy variant

PR #148, opened 2021-03-01
Update master from develop

miggol:
This merge contains content updates, translations, and fixes for escaped HTML in form labels and #125 . Tested on ACC and all seems well.


miggol:
This now also contains fixes for practice proposals. Merging now as a hotfix because students are using the practice functionality for courses.

tymees:
Small warning: I see the MaxValueValidator added here (probably done elsewhere), but I can't see any migration related to that in this PR.

This probably means you either forgot to add it to git, or to create it at all. While this validator enforces this contraint in Python, it also generates a migration. (I think it also adds a DB contraint as well).

Might be worth looking into

miggol:
Ah yes, well spotted. I have migrations in my working dir, so I think I forgot to add them to the commit. I'm guessing that the deploy script does not `makemigrations`, but only migrates existing ones. I will add the migrations to develop and immediately merge to resolve this before more development takes place.

tymees:
> I'm guessing that the deploy script does not makemigrations, but only migrates existing ones.
Nope, that would create a ton of problems if it did.

> I will add the migrations to develop and immediately merge to resolve this before more development takes place.
Awesome!

Issue #154, opened 2021-03-09
Practice applications are neglected


miggol:
I can think of a couple of improvements:

* Templates in the practice form should reflect the fact that you are editing a practice proposal
* Practice applications should either not be submittable at all or have a custom submission page that explains what will happen. Currently practice applications use the default submission success page, which is misleading because practice applications don't do anything on submission.
* There is no link to create a new practice proposal in the menu or on the home page

miggol:
Largely fixed in #155 but leaving this open while I still add links to the front page to create practice proposals. I think a main menu entry is a bit much though, so I'm leaving that out.

tymees:
I actually think adding it to the main menu would actually be a good idea.

That way, it might actually get used (people tend to ignore everything on the frontpage) and it might actually convince a certain VIP the practice portal is actually the normal portal and not the testserver...

miggol:
Alright, it's on the list!

I've also just noticed that practice proposals are explicitely excluded from validation and therefore don't show any errors, which seems counterproductive to actual practice. It's coded into both the template and clean() function. Do you know the intentions of this?

tymees:
Hmmm... If I were to guess it's because previously one had to answer everything before you could go to the next page. Which might be annoying if practicing and you're not really interested in what page 1 needs. 

Can't really think of a different reason, as I agree that it feels counter-productive

miggol:
Status since #160 is that things are better, though I still need to:

* Add a menu item
* Re-enable validation

miggol:
Practice proposals now work pretty well in master

PR #155, opened 2021-03-09
Improvement/practice portal

miggol:
Starting to address #154


proposals/views/proposal_views.py
```diff
@@ -249,6 +249,13 @@ def get_next_url(self):
         else:
             return reverse('proposals:wmo_create', args=(proposal.pk,))
 
+class ProposalContextMixin:
```
tymees, line 252: Might it be preferable to keep mixins in a ``mixins.py``?

PR #160, opened 2021-03-12
Update production from develop

miggol:
I've tested everything relevant to the commits, but might be best to deploy this after the weekend in case anything breaks.


proposals/templates/proposals/proposal_submit.html
```diff
@@ -121,7 +121,12 @@ <h2>
                 </div>
             {% endif %}{% csrf_token %}
                 <table>{{ form.as_table }}</table>
-                {% if troublesome_pages %}
+                {% if is_practice %}
+                <p>
+                    {% trans "U bewerkt op het moment een oefenstudie. Deze kan niet ter beoordeling door de FETC-GW worden ingediend." %}
```
tymees, line 126: It might be better to wrap this is a ``<div class="info"`` for better visibility

reviews/models.py
```diff
@@ -88,6 +88,8 @@ def update_go(self):
                 # On NO-GO, reset the Proposal status
                 # TODO: also send e-mail?
                 else:
+                    from .utils import notify_supervisor_nogo
```
tymees, line 91: Why the local import?

reviews/utils.py
```diff
@@ -253,6 +253,27 @@ def notify_secretary(decision):
...
+        'decision': decision,
+        'supervisor': supervisor,
+    }
+    
+    for applicant in receivers:
+        params['applicant'] = applicant
+        msg_plain = render_to_string('mail/supervisor_decision.txt', params)
+        send_mail(subject, msg_plain, settings.EMAIL_FROM, [applicant.email])
+    
+    print('\n\n',receivers, decision,'\n\n')
```
tymees, line 273: Debug statement?

reviews/models.py
```diff
@@ -88,6 +88,8 @@ def update_go(self):
                 # On NO-GO, reset the Proposal status
                 # TODO: also send e-mail?
```
tymees, line 89: I think this TODO is done now? :P

reviews/utils.py
```diff
@@ -253,6 +253,27 @@ def notify_secretary(decision):
     send_mail(subject, msg_plain, settings.EMAIL_FROM, [secretary.email])
 
 
+def notify_supervisor_nogo(proposal, decision):
```
tymees, line 256: The proposal parameter is a bit redundant. You can access that through ``decision.review.proposal``

It's not bad or anything, but it's more consistant to use that method, as it's also used for the ``notify_secretary`` function (in the template)

reviews/models.py
miggol: This prevents a circular. I'd rather import models at the top layer in `utils.py` than import a utility at top level elsewhere. Models make more sense to have available everywhere.

tymees: Right.

Friday 5PM  strikes again, should've spotted that

PR #163, opened 2021-03-15
Some compare proposal fixes

tymees:
Three fixes:

* Some word documents weren't seen as such, due to a missing check for the 'application/msword' mimetype
* Sessions weren't compared in proposals that have METC decision attached
* Tasks and sessions were improperly compared. (Same issue as with trajectories we had before)

### DIFF
--- a/proposals/templates/proposals/proposal_diff.html
+++ b/proposals/templates/proposals/proposal_diff.html
@@ -211,7 +211,7 @@ $(function() {
             </table>
 
             {% if proposal.wmo.status == proposal.wmo.NO_WMO or proposal.wmo.status == proposal.wmo.JUDGED %}
-            {% for study, p_study in zipped_studies %}
+            {% for study, p_study in proposal.study_set.all|zip:p_proposal.study_set.all %}
             <h2 class="mt-5 mb-1">{% trans "De deelnemers" %}</h2>
             {% include "studies/study_title.html" %}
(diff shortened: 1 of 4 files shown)


tymees:
One problem reported resulted in 2 extra problems getting fixed. That's what I call efficient

miggol:
It took me way to long to actually identify the problem, even with your commits. Nice work on finding the other two!

proposals/templates/proposals/proposal_diff.html
```diff
@@ -209,7 +210,7 @@ <h2 class="mt-5 mb-1">{% trans "Eén of meerdere trajecten?" %}</h2>
                 {% endif %}
             </table>
 
-            {% if proposal.wmo.status == proposal.wmo.NO_WMO %}
+            {% if proposal.wmo.status == proposal.wmo.NO_WMO or proposal.wmo.status == proposal.wmo.JUDGED %}
             {% for study, p_study in zipped_studies %}
```
miggol, line 214: I like the filter solution more than the context var, could we use it here as well?

miggol (approved):

proposals/templates/proposals/proposal_diff.html
tymees: Yes we can, and now we do!

Good spot, missed it yesterday.

PR #176, opened 2021-04-01
Improvement/documents list

miggol:
This is a rewrite of the detail sidebar, which now appears on the assign page as well. I wouldn't exactly call this production-ready, it does need some serious testing on develop and there's some things I want an opinion on.

- Where should things like `documents_list` and `simple_compare_link` reside? They can go in different places. But main didn't feel like the best place.
- How should we handle deprecated observations' `approval_documents`? Currently `simple_compare_link` can deal with them, but `documents_list` doesn't look for them. It wouldn't be too hard to check for them explicitly but it felt unnecessary as we'll never have a review with that case ever again.

Any and all comments welcome


tymees:
I've added a commit, most of which are style related. (Mostly related to stuff I had difficulty reading due to density). 

Addional changes:
* Changed the if statements of ``simple_compare_link`` to a if/elif/else, and return an empty dict if the given object type isn't recognised.
* Replaced the 3 with statements in ``document_list`` with a simple unpacking for loop
* Added more detailed comments on the documents QS to explain what that QS actually tries to do

One final note:
``documents_list.html`` and ``simple_compare_link.html`` aren't in a namespace folder (e.g. ``reviews/x.html``), is this on purpose or can these files be moved into a namespace folder?

reviews/templates/documents_list.html
```diff
@@ -0,0 +1,41 @@
...
+       {{header}}
+    </h5>
+
+    <ul>
+    {% for name, file, obj in items %}
+        <li>
+            <a href="{{ file.url }}" target="_blank">
+                {{name}}
+            </a>
+            {% if file.field.name != "pdf" %}
```
tymees, line 21: Why are we filtering PDF files here?

miggol:
On the location of the files: I simply started coding not really knowing where they would go in the end. The sidebar is review-specific but `documents_list` and `simple_compare_link` could be useful elsewhere. I just didn't know where to put them and was wondering about your input. 

My opinion would be to make a namespace for inclusion tag templates such as `inclusion_templates` in `reviews/templates` and put the templates there. And then keep `documents_list.py` in the reviews app for the time being.

What do you think about no longer handling observation files on the review detail page?

tymees:
> On the location of the files: I simply started coding not really knowing where they would go in the end. The sidebar is review-specific but documents_list and simple_compare_link could be useful elsewhere. I just didn't know where to put them and was wondering about your input.
My opinion would be to make a namespace for inclusion tag templates such as inclusion_templates in reviews/templates and put the templates there. And then keep documents_list.py in the reviews app for the time being.

The namespace is often just the name of the app that 'owns' it, with any other grouping happening below that. For example, the proposals template dir has 'diff' and 'pdf' as subdirs to group those subtemplates.

 The reason why it's an app name is because you can 'override' a template from a different app above you in the loading order. Thus, by explicitly using the app name, you at least know that one is overriding that app. For example, we override the ``registration`` app templates to provide our own login and logged out pages. 

(and yes, we are actually doing this wrong  with the ``mail`` folder in the reviews app, that should be a subdir of ``reviews``....)

Thus, I would move it into the ``templates\reviews`` folder as long as the corresponding Python code is also in that app. An additional ``inclusion_templates`` subdir might be best. To be honest, if one of us has time and motivation, we probably should group templates more. (At least the proposal templates can be grouped... I hope). 

> What do you think about no longer handling observation files on the review detail page?

I'd rather keep it in there, otherwise all studies before 2018 don't give access to those documents. You might ask if those are still relevant, to which I would respond: probably not, but life has a tendency to surprise.

miggol:
Ah yes, thank you for the explanation of the namespaces. I'll at least be moving them into `templates/reviews` then.

So my next question on the observations is if you know of a reasonable way of making a local "deprecated" proposal with the old-style observation so I can test while developing? Even by shell I wouldn't really know where to start.

If not I guess I'll just work blind based on how they were handled before and cross my fingers when we merge to master.

miggol (commented):
> I've added a commit, most of which are style related. (Mostly related to stuff I had difficulty reading due to density).
> 
> Addional changes:
> 
>     * Changed the if statements of `simple_compare_link` to a if/elif/else, and return an empty dict if the given object type isn't recognised.
> 
>     * Replaced the 3 with statements in `document_list` with a simple unpacking for loop

Nice changes by the way. IMO this is now ready for merging into develop. I'll make sure to be ready to test an old proposal the day it goes to master.

reviews/templates/documents_list.html
tymees: Oh it's the proposal PDF! I read it as filterring pdf's in general. This makes more sense. 
(Maybe we should add a comment what it's actually filtering)

tymees (approved):

Issue #185, opened 2021-04-12
Secretary should no longer be a required reviewer


miggol:
Currently every review must have at least one secretary user assigned to it. The origin of this rule might have been technical or procedural, but currently the secretary would prefer it didn't exist. This would save a step in reviews where the secretary isn't interested in reviewing a proposal themselves. A secretary has to make the final call anyway.

After removing this rule in `ReviewAssignForm`, I've run into the following:

- If a secretary user visits the close form for a review they are not a part of, they are auto-added back to the list of reviewers.
- The "decide" button goes straight to `ReviewCloseForm`, which is quite similar to the decide form. There is no warning if there are still open decisions for this review. So a (temporary) secretary could easily get confused and close a review thinking they were simply adding a comment.
- The template link somehow bugs out if a secretary user tries to go to "decide" when they're not in the reviewing users `NoReverseMatch at /reviews/decide/new/39)/` (sic, with parenthesis)

These are issues that will likely cause problems when there is more than one secretary, so I'm going to get this fixed before the temp secretaries take over.

miggol:
This requirement has now been lifted and it's live in production.

PR #188, opened 2021-04-19
Fix/multiple secretaries

miggol:
This addresses #185 so that multiple secretaries can work at the same time.

`CreateDecisionRedirectView` is now unused, and I've chosen to comment out rather than remove code relating to the "always have a secretary as reviewer" rule. We might need it again in the future.


tymees:
Looks good.

However, I think commenting out, or placing a return statement before that code, is not really the way to go. The beauty of Git is that we can recover any changes we made, so why not use that? We could always leave a comment somewhere that we used to have that functionality, but it was removed the date of the commit. 

It's a bit more hidden that way, but that also means cleaner code.  (And less accident prone)

miggol:
Good point, I agree. The only part I still left in is `CreateDecisionRedirectView`, because I can see some use for it in the short term now that secretaries are no longer required reviewers. Something like a quick "add decision for me" button.

PR #199, opened 2021-05-17
Some UFL bug fixes

tymees:
This PR fixes the following:
- Sort on reference number not working on decision list pages
- Conclusion showing as a sort option on decision pages (now removed, no pages benefit from this)
- Conclusion showing as a sort option on review pages (now a filter)
- get_continuation_display being a bit eager in predicting the result of a review before it has been concluded
- Inconistant usage of 'stage' vs 'phase'

Lastly, this branch targets the release version of UiL Core, which has some additional bug fixes. The only relevant fix is the search bar no longer crashing and burning if it can't find a specific attribute.


tymees:
Also changed Frans' email adress to the right one. I added this here, as merging locale files is annoying and this avoids that whole thing.

miggol (approved):
Looks good!

PR #201, opened 2021-05-18
 [HOTFIX] Fixed edit document link only linking to the first trajectory to develop

tymees:
The previous document edit link only allowed editting the first trajectory (well, the first document object).

This PR fixes that, by adding an edit link for every documents object.

Also changed the compare link logic a bit.


miggol:
I think when I wrote it I was somehow convinced that studies shared a Documents item. Having a single link for the attachments page would never have worked without rewriting that as well. Quite an educational hotfix for me!

PR #202, opened 2021-05-18
Instruct Travis to use Python 3.6

tymees:
Title pretty much says it all.

Should fix Travis erroring, making it usefull again


tymees:
Okay Travis works again, showing us we have some errors when running the tests.

Probably should investigate those errors at some point

Issue #126, opened 2021-01-28
Invalid fields are not saved


miggol:
If a submitter changes a field value to something invalid, for example by removing the proposal's description on page 1, they can submit that page or navigate away without that field actually being saved.

If a field is left blank from the start, this behaviour is not observed. 

This can lead to problems if a user intends to leave a field blank. The application does not warn them about the missing information but keeps the old value. That old value may be not at all what the user wanted to submit.

miggol:
The desired behaviour is already present for simple model fields with `blank=True` and soft validation, except for Create forms which really do need a value to go to the next step. That is good. I currently believe the bug comes from `ConditionalModelForm` and `check_dependency()`, which bypasses soft validation . I'm looking into a neat way of fixing this.

EdoStorm96:
I think this is still happening? I noticed it again today ...

Issue #123, opened 2021-01-28
Unable to navigate away from required questions


miggol:
This is an issue similar to the accidental submission in #120, and can be seen on the first METC page. Possible fix could be SoftValidationMixin, because in this case we actually DO want form submission so other answers are saved.

PR #205, opened 2021-05-21
Added "avg_understood" field, and DMP file upload

miggol:
Changes are reflected in the review detail sidebar. Also fixed a bug where supervisors were not allowed to compare items. Migrations and translations included.


miggol:
Just realised I didn't change the text on the DMP page yet, will add that before merge.

proposals/forms.py
```diff
@@ -523,10 +523,12 @@ def clean(self):
                     self.add_error(study_name, _('Dit veld is verplicht.'))
 
 
-class ProposalDataManagementForm(forms.ModelForm):
+class ProposalDataManagementForm(SoftValidationMixin, forms.ModelForm):
```
tymees, line 526: If you're going to use SoftValidationMixin, you'll probably want to add the form to ``_build_forms`` in ``proposals/utils/validate_proposal.py`` as well, as otherwise a user won't be stopped properly ;)

tymees: Derp, you did that. Ignore me

proposals/validators.py
```diff
@@ -1,21 +1,33 @@
...
         if self.proposal:
             qs = qs.exclude(pk=self.proposal.pk)
 
         if qs.exists():
             raise forms.ValidationError(_('Er bestaat al een studie met deze '
-                                          'titel.'), code='unique')
\ No newline at end of file
+                                          'titel.'), code='unique')
+
+def AVGUnderstoodValidator(value):
```
tymees, line 28: Any reason to use a whole validator over just marking the DB field as required?

proposals/models.py
```diff
@@ -287,6 +289,22 @@ class Proposal(models.Model):
         null=True,
         blank=True,
     )
+    
+    avg_understood = models.BooleanField(
```
tymees, line 293: This _technicaly_ is mixing Dutch and English, as it should be GPDR in English.

(You could defend it with the argument that the AVG is a specific Dutch law based on the generic GDPR and thus a valid name in English as well, if you feel changing it is too much work ;) )

locale/en/LC_MESSAGES/django.po
```diff
@@ -1597,36 +1597,50 @@ msgstr ""
...
+#: proposals/models.py:276
 msgid "Hoeveel verschillende trajecten zijn er?"
 msgstr "How many different trajectories are there?"
 
-#: proposals/models.py:305
+#: proposals/models.py:294
+msgid ""
+"Ik heb kennis genomen van het bovenstaande en begrijp mijn "
+"verantwoordelijkheden ten opzichte van de AVG."
+msgstr ""
```
tymees, line 1608: Missing translation?

proposals/validators.py
miggol: So far, this is the only way I've found to work around both #126 and #123 (very much pet peeves of mine) in the current proposal form. Either you can't navigate away from the page, or you use SoftValidationMixin and accept that an invalid value will never be saved because of a DB constraint.

The latter admittedly only comes in to play with a user checking the box, saving the form, and then later trying to uncheck it, which is kind of their own fault for trying to un-read something. But this feels like the kind of question where there should be no possibility of the user "unintentionally" having this box checked.

proposals/models.py
miggol: As I understand it they're not entirely interchangeable. Data collected in NL must comply with the AVG, and if it does, it complies with the GDPR directive as well. But the AVG implementation is more protective in certain areas than strictly necessary through the GDPR. AVG implies GDPR, but not vice versa. 

If we imagine an international researcher used to say, the French implementation of GDPR, they might break the AVG while still technically complying with GDPR. So we ask them specifically to comply with AVG. Thankfully, neither GDPR or AVG say anything about intercourse with ants.

locale/en/LC_MESSAGES/django.po
miggol: Added!

proposals/models.py
tymees: That's entirely true from what I believe as well, which is why I gave that defense in my comment as well :p (although I also believe the gdpr implementations of other eu states should be accepted universally... But that doesn't affect us, as that only comes into play if someone shops at a German site for example).

The thing is, our beloved bureaucrats at the government really like to be confusing. The official dutch translation of avg is actually also gdpr. (Even though the gdpr and the gdpr are indeed different). See also the English site of the AP for example.

Anyway, let's just leave it the way it is then. We have enough ants in the vending machine at work already.

PR #208, opened 2021-05-28
Simple system message system to master

tymees:
This PR adds a simple way to display system messages (like scheduled downtime) for a set period of time. 

I've targetted master as develop contains some code that isn't ready yet, and I prefer to have this online today if possible, for the downtime next weekend


miggol (approved):
Looks good! Does adding the field in `translation.py` mean there's an field for both locales? I'll read up on that part next week either way.

tymees:
Yes it does!

It's part of the modeltranslation library, which dynamically adds extra fields to the model and the admin to provide localized values.

PR #227, opened 2021-07-30
back-merge hotfix from master

tymees:
This PR merges back a hotfix in master which caused the LK public archive fail to load. 

It also adds a cache on both public archives; this could be done straight to master as it won't be a problem if the list is an hour or 2 out of date. (In contrast to the private archive, in which it's actually pretty bad when that happens).


miggol:
If I understand correctly, the non-inline `ProposalSerializer` exists for where we need a `continue_url`, and otherwise we use the inline version internally. (Maybe I'm missing something)

Is this just an optimisation or does a method like `continue_url` in the serializer fields break stuff?

Top job on the fix otherwise

tymees:
It's a fix for a serializing problem, but it's not related to ``continue_url``;

The non-inline version is the main serializer for a Proposal. As a proposal may have a parent proposal, which we sometimes need, we need to add a ``parent`` field. Normally you could just use the same serializer (which we accidentally did before this fix); however, some (older) proposals are kinda f-ed, and either directly or indirectly reference eachother. (In other words, two proposals have eachother as a parent). This causes an infinite recursion, thus the need for a second serializer which doesn't serialize it's parent.

As we don't really need to access the parent of a parent (yet), this actually also saves some processing as we limit the amount of objects to be serialized. So it's also optimization.

The ``continue_url`` field isn't part of the inline version mostly due to me not realizing I didn't add it, but I'll take your optimization explaination as it actually holds true :P It's an expensive call, and it's not needed for the parent.

miggol:
Thanks for the explanation!

PR #228, opened 2021-08-03
Improvement/new continuation

miggol:
New continuation which TERMINATES reviews. Also an action framework which I'm looking to adapt to the action icons in ufl lists in the future.

### DIFF
--- a/reviews/models.py
+++ b/reviews/models.py
@@ -28,6 +28,7 @@ class Review(models.Model):
     METC = 4
     GO_POST_HOC = 5
     NO_GO_POST_HOC = 6
+    UNSUBMITTED = 7
     CONTINUATIONS = (
         (GO, _('Goedkeuring door FETC-GW')),
         (REVISION, _('Revisie noodzakelijk')),
@@ -36,6 +37,7 @@ class Review(models.Model):
         (METC, _('Laat opnieuw beoordelen door METC')),
         (GO_POST_HOC, _('Positief advies van FETC-GW, post-hoc')),
         (NO_GO_POST_HOC, _('Negatief advies van FETC-GW, post-hoc')),
+        (CLOSED, _('Niet verder in behandeling genomen door de FETC-GW')),
     )
 
     stage = models.PositiveIntegerField(choices=STAGES, default=SUPERVISOR)
--- a/reviews/models.py
+++ b/reviews/models.py
@@ -37,7 +37,7 @@ class Review(models.Model):
         (METC, _('Laat opnieuw beoordelen door METC')),
         (GO_POST_HOC, _('Positief advies van FETC-GW, post-hoc')),
         (NO_GO_POST_HOC, _('Negatief advies van FETC-GW, post-hoc')),
-        (CLOSED, _('Niet verder in behandeling genomen door de FETC-GW')),
+        (UNSUBMITTED, _('Niet verder in behandeling genomen door de FETC-GW')),
     )
 
     stage = models.PositiveIntegerField(choices=STAGES, default=SUPERVISOR)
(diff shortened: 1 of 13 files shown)


reviews/models.py
```diff
@@ -36,6 +37,7 @@ class Review(models.Model):
         (METC, _('Laat opnieuw beoordelen door METC')),
         (GO_POST_HOC, _('Positief advies van FETC-GW, post-hoc')),
         (NO_GO_POST_HOC, _('Negatief advies van FETC-GW, post-hoc')),
+        (UNSUBMITTED, _('Niet verder in behandeling genomen door de FETC-GW')),
```
tymees, line 40: Unsubmitted isn't that clear. With this definition it makes sense, but as I read the word unsubmitted in a different file before I saw this, I thought it meant 'not submitted yet'. 

Maybe change it to 'NOT_CONSIDERED'?

reviews/utils/review_utils.py
```diff
@@ -11,7 +11,7 @@
 from proposals.models import Proposal
 from tasks.models import Task
 from proposals.utils import notify_local_staff
-from .models import Review, Decision
+from reviews.models import Review, Decision
```
tymees, line 14: Any reason for this change?

reviews/views.py
```diff
@@ -9,14 +9,15 @@
...
 from proposals.models import Proposal
 from .forms import (DecisionForm, ReviewAssignForm, ReviewCloseForm,
-                    ChangeChamberForm)
+                    ChangeChamberForm, ReviewUnsubmitForm)
 from .mixins import (AutoReviewMixin, UserAllowedMixin,
                      CommitteeMixin,
                      UsersOrGroupsAllowedMixin)
 from .models import Decision, Review
-from .utils import notify_secretary, start_review_route
+from reviews.utils.review_utils import notify_secretary, start_review_route
```
tymees, line 20: The imports are inconsistent here;
.utils.review_actions is an relative import, but reviews.utils.review_utils is an absolute import.

reviews/views.py
```diff
@@ -246,6 +258,36 @@ def form_valid(self, form):
...
+    template_name = 'reviews/review_unsubmit_form.html'
+    group_required = settings.GROUP_SECRETARY
+
+    def get_success_url(self):
+        'Return to the detail view after unsubmission'
+        committee = self.object.proposal.reviewing_committee.name
+        return reverse('reviews:detail', args=[self.object.pk])
+
+    def form_valid(self, form):
+        'Sets the unsubmitted continuation on the review'
```
tymees, line 273: Maybe move most of this logic in to a helper function?

tymees:
Looks good; for some of the small issues I've added a comment, but I got one more that I didn't really know where to add;

First off, splitting off the utils into seperate packages is good; no problem there. However, personally I like to add an __init__.py to the utils package, which imports all the methods from the other files. This in turn allows one to import it from the ``app.utils`` package, which is a bit easier in cases you don't realy remember which sub-package it's actually located in.

It's just personal preference, feel free to ignore ;)

reviews/views.py
miggol: Bit messy on my part, must have just set it quickly and forgotten it

reviews/utils/review_utils.py
miggol: Can't remember how it got there

reviews/models.py
miggol: How about "discontinued"? Or maybe just "removed". I feel like "not_considered" implies that it never was considered.

Coming up with a good word for this has been the hardest part of the entire branch

tymees: > How about "discontinued"? Or maybe just "removed". I feel like "not_considered" implies that it never was considered.

I'd be fine with discontinued; it still doesn't fit perfectely, but I don't think it would be confusing. I'm not a big fan of removed, as it's a bit too generic. (In other words, it doesn't really tell what happend). 

> Coming up with a good word for this has been the hardest part of the entire branch

Naming things is always much harder than it would seem, sadly.

PR #229, opened 2021-08-10
Implemented new 'all running reviews' view

tymees:
As requested, a view that views all reviews except those who've been closed.


miggol (approved):
Prachtig!

PR #244, opened 2021-08-27
Hotfix/broken import 2.0

tymees:
Same fix as the previous PR to master, but merged with develop to override the previous attempt to fix the problem.

Apperantly the import fix from Tuesday works fine everywhere *except the uWSGI environment*. An absolute import should fix this once and for all.

Additionally, this PR seems to pull in some other fixes that aren't in develop?


miggol:
7942b08e67cae2850b2e3c16f9c430738cc9d072 was an absolute massacre. I don't know what crazy emacs function I accidentally tapped to invoke such wrath, but I'll be sure to double check my diffs on Github as well from now on.

miggol (approved):

PR #246, opened 2021-09-02
Added UiL-OTS fork of docx2txt to support Word online documents

miggol:
This has brought in a lot of updates to libraries which weren't pinned to a specific version. Nothing broke for me, so fingers crossed nothing will break on ACC.

Big thanks to Github user schmamps for the improvements to docx2txt


tymees (approved):
Looks good.

And yeah, we probably didn't run a thorough dependency update in a while :P

Issue #211, opened 2021-06-01
Mention current documents on comparison page


miggol:
As per secretary request, should be a quick one.

miggol:
Fixed in improvement/show-compared-files

PR #248, opened 2021-09-06
Improvement/show compared files

miggol:
Old branch that never got merged fixing #211


studies/templatetags/describe_documents.py
```diff
@@ -0,0 +1,83 @@
+from django import template
+from studies.models import Documents, Study
```
tymees, line 2: It's customary to order imports thusly:
standard library
<whitespace>
libraries
<whitespace>
in-project code

Not really a big issue, but I noticed I was a tad confused when trying to find the source of give_name

miggol: Good point, I usually appreciate the ordering as well! This was just laziness / forgetfulness on my part

PR #253, opened 2021-09-10
More textual changes 2: Electric Boogaloo

tymees:
Draft for now, as there might be additional changes incoming


tymees:
@miggol Should be ready now;

Could you please check my translations? Not really sure I managed to get them right

PR #255, opened 2021-09-13
Also show refnum when copying a proposal

tymees:
This should make it easier to select the right version to copy if there
are several proposals with similar names


miggol (approved):
What's your opinion on condensed code like

```python3
'Refnum - Title {}'.format(
    '(practice)' if obj.is_practice() else ''
    )
```

Each has their merits, just wondering

tymees:
> What's your opinion on condensed code like
> 
> ```python
> 'Refnum - Title {}'.format(
>     '(practice)' if obj.is_practice() else ''
>     )
> ```
> 
> Each has their merits, just wondering

It's not my style (I prefer no logic in formatting statements), but I wouldn't mind it either. In this case I probably wouldn't even have noted it :p

PR #263, opened 2021-09-27
Dependencies

miggol:
Somewhat convoluted way to do this: I'm practicing my git-fu


tymees:
Might it not be a better idea to run ``pip-compile -U``?

That way we update all dependencies. (From my experience, dependabot doesn't take into account python 3.6 compatibility and new required versions for the updated dependencies)

PR #267, opened 2021-11-15
Feature/nwo update

miggol:
Preassessments have been in need of an update for a while now. This brings textual clarifications for the start_pre page and menus. Also a flowchart for good measure.


miggol:
These are just text and html questions so I'm moving them straight to ACC

PR #272, opened 2021-11-23
Move to GitHub Actions for testing

tymees:
Previously, we used Travis CI to do automatic testing. As travis.org was discontinued, this no longer works.

We could move over to travis.com, but as GitHub now has build-in CI, it's easier to use that instead. This PR implements that move.


miggol (approved):
Nice. Now for me to figure out how to write good tests

PR #283, opened 2022-01-10
Hotfix pdf link

miggol:
Only bringing this into master as a temp fix. Develop deserves a less bodgy fix. I've been thinking of splitting the `documents_list` megafunction into something more reasonable anyway.


tymees:
>  I've been thinking of splitting the documents_list megafunction into something more reasonable anyway.

Again? :P

PR #285, opened 2022-01-13
Removed rigid dict

miggol:
Replaced it with container classes

### DIFF
--- a/reviews/templatetags/documents_list.py
+++ b/reviews/templatetags/documents_list.py
@@ -157,6 +157,153 @@ def give_name(doc):
                 return _("Extra documenten {}").format(n+1)
 
 
+
+class Container:
+
+    def __init__(self, header, **kwargs):
+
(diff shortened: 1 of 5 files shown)


miggol:
Oh, and also this fixes the PDF link bug. If that wasn't obvious

reviews/templatetags/documents_list.py
```diff
@@ -245,7 +244,7 @@ def documents_list_v2(review, user):
 
     for d in qs:
         items = []
-        files = [
+        potential_files = [
```
tymees, line 247: Why are they 'potential'?

tymees: nvm, i get it now

tymees (approved):
Looks good! I did spot some PEP8 problems (too much whitespace between classes/functions), but that's minor

Feel free to merge whenever you want

PR #304, opened 2022-03-03
Feature/textual updates

miggol:
* Add jspsych to registrations model
* Remove schoolleiding
* Fix a bad link on the homepage


tymees:
I'll try to go this review after the meetings today

> Question for @tymees: do fixtures get loaded on deploy?
No. The fixtures rarely change, so it wasn't worth the effort of engineering a way to load them reliably. Also, more complex fixtures aren't really safe to re-import, like when dealing with relations between models.

locale/en/LC_MESSAGES/django.po
```diff
@@ -5389,6 +5378,14 @@ msgstr "Task edited"
...
+#~ msgid ""
+#~ "De <a href=\"https://intranet.uu.nl/documenten-ethische-"
+#~ "toetsingscommissie-gw\" target=\"_blank\"> UU-webpagina van de FETC-GW</"
+#~ "a> voor nieuws en de agenda."
+#~ msgstr ""
+#~ "The <a href=\"https://fetc-gw.wp.hum.uu.nl/en/\" target=\"blank\">UU-"
+#~ "webpage</a> of the FEtC-H for news and upcoming dates."
+
 #~ msgid "Is er sprake van een controlegroep?"
 #~ msgstr "Is there a control group?"
```
tymees, line 5391: These can be cleaned up

tymees (approved):
Fine enough; 

However, in the future, it might be best to edit the fixture json manually for such a small change; there are a lot of lines that moved around but don't differ in contents. This makes reviewing a bit tedious

locale/en/LC_MESSAGES/django.po
miggol: Done

PR #306, opened 2022-03-04
Updated dependencies

tymees:
Makes dependabot happy

Tested to work locally with Python 3.9, did have to pin reportlab to a specific version because something inside docx2txt broke. Might want to investigate that later, but for now we should at least address the security problems in the other dependencies


tymees:
Okay, also added a fix for GitHub actions still using an old workflow config. Now it tests newer Python versions and doesn't stop all tasks if one fails

miggol (approved):

Issue #300, opened 2022-02-24
Assigning a new reviewer should unset Review.date_end


miggol:
Otherwise, the new reviewer will not be allowed to add comments. We could add this check to update_go() to keep the logic in the same place.

miggol:
Fixed in prod

PR #310, opened 2022-03-17
Fix/unlock after reassignment

miggol:
I've moved the assignment logic to review_utils so it might be hard to see what's changed. The gist is as follows:

`update_go` now unsets `date_end` if there are unmade decisions. This, combined with setting `review.COMMISSION` in `assign_reviewers()` means a review reverts back to being open when a reviewer is added. (#300)


reviews/utils/review_utils.py
```diff
@@ -431,3 +430,45 @@ def discontinue_review(review):
...
+
+
+def assign_reviewers(review, list_of_users, route):
+    """Assigns or reassigns a list of reviewers to a review"""
+
+    # Set stage to commission if users are assigned
+    if len(list_of_users) > 0:
+        review.stage = review.COMMISSION
+    else:
+        review.stage = review.ASSIGNMENT
```
tymees, line 442: Does this actually happen?

The assignment form has this validation rule:
```python
    def clean_reviewers(self):
        reviewers = self.cleaned_data['reviewers']

        if len(reviewers) == 0:
            raise ValidationError(
                _('Er moet tenminste één beoordelaar geselecteerd worden.'),
                code='no_reviewer')

        # See PR #188 if a secretary should be added to every review
        return self.cleaned_data['reviewers']
```

reviews/utils/review_utils.py
```diff
@@ -431,3 +430,45 @@ def discontinue_review(review):
...
+        review.date_should_end = timezone.now() + \
+            timezone.timedelta(
+                weeks=settings.SHORT_ROUTE_WEEKS
+            )
+    elif review.date_should_end is not None:
+        # We have no desired end date for long track reviews
+        review.date_should_end = None
+
+    # Create a new Decision for new reviewers
+    start_review_route(review, new_reviewers, route)
```
tymees, line 460: Like, not your fault, but this function name is really misguiding. ``create_decision`` would be a better idea

reviews/utils/review_utils.py
miggol: Nope, this should never happen under normal circumstances. But when moving it into the utils package I took the liberty of adding this to make the function more utility-like, should we ever want to bump an application back to the assignment phase manually.

tymees: That makes sense, good thinking

reviews/utils/review_utils.py
miggol: I know, I hate it too

PR #320, opened 2022-04-12
Upgrade to Django 3.2

miggol:
At time of creation, this PR is not ready for merge because `generate_pdf()` is still WIP. However, Django 3.2 is happily up and running on `gw-d-lab-fetc-tst`, though still with a couple of warnings relating to Django 4.0. 

`reportlab` requirement seemed to be unused so it was removed.

Let's use this PR to discuss anything relating to the upgrade from now on.


miggol:
I think I'm finally there! This now warrants a review and consideration for acceptance. Note that this update will require the editing of deploy scripts to reflect the new `settings.py`.

proposals/mixins.py
```diff
@@ -31,3 +35,94 @@ def get_context_data(self, **kwargs):
...
+        )
+
+        if pisa_status.err:
+            return HttpResponse('We had some errors <pre>' + html + '</pre>')
+        return dest
+
+    def render_to_response(self, context, **response_kwargs):
+
+        response = HttpResponse(content_type='application/pdf')
+        response['Content-Disposition'] = self.get_content_disposition()
```
tymees, line 126: Duplicate code found on lines 104/105; might be better to extract these into a helper method

tymees (approved):
Looks good; one small point that's in no way blocking.

I do really like the PDF mixin, a very Django like solution for this problem we should probably start reusing if needed :P

PR #328, opened 2022-05-03
Force QS evaluation during copy

### LINKED ISSUE #327: Django 3: copying seems to forget studies
They're simply missing. The 'define number of' page still knows how many you had, but the objects themselves are gone:

```
>>> p = Proposal.objects.last()
>>> p.is_revision
True
>>> p.study_set.all()
<QuerySet []>
```

tymees:
Fixes #327

This was an interesting bug to debug; when attaching a debugger and stepping through the code, the bug automagically dissappeared. Took me a while to figure it out; somewhere between Django 2.2 and 3.2 they changed the behaviour of QuerySet's lazy evaluation, in that it will insert the related PK at the last moment instead of upon definition. 

Thus, during copy it was actually querying the state of the copy proposal instead of the original, resulting in related objects not being copied. The debugger fixed this, as the debugger was forcing the evaluation of the QS directly after definition (to show in the variable list). As a result, it was actually a list of the original proposal's related objects when running with the debugger. 

This PR tries to do the same by wrapping all QS in a `list()`, ensuring the QS is evaluated immediatly. 

And no, this change in behaviour wasn't documented anywhere :-/

### DIFF
--- a/interventions/utils.py
+++ b/interventions/utils.py
@@ -21,7 +21,7 @@ def copy_intervention_to_study(study, intervention):
     """
     Copies the given Intervention to the given Study
     """
-    setting = intervention.setting.all()
+    setting = list(intervention.setting.all())
 
     i = intervention
     i.pk = None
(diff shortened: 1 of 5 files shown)


miggol:
Oof, this process sounds rough. Well found!

Issue #317, opened 2022-03-31
Compare proposal not working correctly with mismatching number of studies


tymees:
The compare proposal page doesn't work correctly (if at all) if the number of studies between versions differ.

The offending line seems to be:
``{% for study, p_study in proposal.study_set.all|zip:p_proposal.study_set.all %}``

This works great if the two iterables are of comparable length, but ``zip()`` stops if the shortest iterable is exhousted. This issue is also present a bit further down the page in the ``session`` comparision section.

One solution could be:
1. Use a different method to combine the two lists. I would suggest a zip method which fills in 'None' for when the shortest list is out of items
2. Change the template such that it can handle missing study/session objects. Right now it expects it to be there

Another might be to completely rewrite the template such that the page is split in two, with both sides displaying a full readout of a proposal. Thus, instead of zipping 2 proposals together, write one template to display 1 proposal and call it twice for both the current and old proposal. Then, use some JS to compare the two against eachother.
One problem with this solution might be aligning the different parts such that they are next to eachother in the page. (Otherwise you might get a study next to a documentlist).

The first option is way easier, but the second one might be better long term. I think it might be a good idea to create one template to display an entire proposal, such that we can reuse that. For example, the PDF template might also make use of that template (as that template really needs a visual overhaul). Another more uncertain usecase might be as general review page, as the proc.reg. wants a review page like that (and doing it this way means we could steal their comment code). 

In anycase, I think it might be a good idea to split the current diff template up into separate parts, as it's very long and a minefield of nested if's and for's.

miggol:
Solution 1 is the fix we need right now, and could implement with minimum effort. 

> Another might be to completely rewrite the template such that the page is split in two, with both sides displaying a full readout of a proposal. Thus, instead of zipping 2 proposals together, write one template to display 1 proposal and call it twice for both the current and old proposal. Then, use some JS to compare the two against eachother.

Could the JS to compare them to one another be adapted from the compare documents page? I don't know how well it would handle html formatting, but we could at least apply it to the textual answers.

> I think it might be a good idea to create one template to display an entire proposal, such that we can reuse that. For example, the PDF template might also make use of that template (as that template really needs a visual overhaul). Another more uncertain usecase might be as general review page, as the proc.reg. wants a review page like that (and doing it this way means we could steal their comment code).

[...]

PR #329, opened 2022-05-03
Refactor diff page

tymees:
Depends on PR #328 (Because testing this was impossible without that fix)

This PR does the following:
* Splits out the diff page template (even more) into smaller sub templates. This was done to improve readability 
* Adds levels of indentation where previously they were missing. Mostly seen around template tags, which were prevously not adding levels of indentation. This, again, was done to improve readability
* Fixes a bug in where comparing lists of objects (studies, sessions, tasks, etc) items were not displayed if the lengths of those lists were unequal. See also #317
* Added extra logic dealing with data missing from one or the other side of the comparison
* Changes behaviour if one research type was added/removed in the revision. Previously, it would only show a warning it was added/removed; now it also shows the contents alongside the warning.

I went a bit out of scope for the issue listed, but from what Desiree told me some committee members are using this comparison page as their main reviewing tool. Thus, I thought it would be best to make the page as complete as possible such that it can actually properly function in that role.


tymees:
I'm not 100% happy with this result, but I was mostly fighting annoying Django behaviour so I think this is good enough for now. At least we don't have a 650+ lines template anymore ><

miggol:
This is a great improvement

PR #336, opened 2022-05-23
Migrate to CDH Core 3.0

tymees:
Initial work to migrate to the upcoming Django shared core 3.0; turned out to be easy.

Draft as DSC is not quite at release candidate quality yet, and there's currently a weird bug in some DSC related translatables not being translated;


tymees:
The latest commits updates to the new DSC alpha 7

As part of this update, _all_ templates have been updated to fix breaking CSS changes in said update. (It was _not_ fun).

But bigger news is the initial work on some redesigns:
- Initial work on the frontpage redesign as seen in #364 (the quick continue stuff is not yet implemented, resulting in a weird empty spot)
- Initial work on redesiging the main proposals form
  - A new base template for those forms has been added which significantly reduces the amount of code in most form templates. Only templates which didn't neatly fitted into the template have custom versions. (Note: with some further work, most if not all of the proposal form templates can be handled by the new base template).
  - The old horizontal stepper is completely replaced with a vertical stepper (that also shows status). It's nowhere near complete, with missing mobile views and missing steps
   - All forms now use the new ``TemplatedForm`` from alpha 7. This change highlights the need to **seriously shorten** help texts. Right now it just look redicilous

tymees:
It can be previewed on the test server:
https://fetc.tst.hum.uu.nl/

miggol:
RIP

Issue #334, opened 2022-05-23
[Bug] Intervention form unnecessarily asks user to check "task-based research"


miggol:
![image](https://user-images.githubusercontent.com/7064137/169805325-5e20cce5-a230-4827-9c94-3e62902afc67.png)

Unclear if this behaviour is blocking or just annoying. Needs investigation.

tymees:
I checked when the mail came in, it's not blocking. It's just a *very* simple check if that field is 'yes', it doesn't look at anything else

miggol:
Thanks for checking!

miggol:
Fixed in prod

PR #338, opened 2022-05-24
Feature/refnum in subject

### LINKED ISSUE #331: Add reference number to all mail subjects
Most of them have it added already, but some don't. 

Bonus points for adding the ref.num. to the body in a logical manner as well

miggol:
Fixes #331 & #334

### DIFF
--- a/reviews/tests.py
+++ b/reviews/tests.py
@@ -44,9 +44,19 @@ class BaseReviewTestCase(TestCase):
         generate_pdf(self.proposal, 'proposals/proposal_pdf.html')
         self.study = Study.objects.create(proposal=self.proposal, order=1, compensation=Compensation.objects.get(pk=2))
 
+    def check_subject_lines(self, outbox):
+        """
+        Make sure every outgoing email contains a reference number and the
+        text FETC-GW
+        """
+        for message in outbox:
+            subject = message.subject
+            self.assertTrue('FETC-GW' in subject)
+            self.assertTrue(self.proposal.reference_number in subject)
+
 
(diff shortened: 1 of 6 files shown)


reviews/tests.py
```diff
@@ -44,9 +44,19 @@ def setUp(self):
...
         self.study = Study.objects.create(proposal=self.proposal, order=1, compensation=Compensation.objects.get(pk=2))
 
+    def check_subject_lines(self, outbox):
+        """
+        Make sure every outgoing email contains a reference number and the
+        text FETC-GW
+        """
+        for message in outbox:
+            subject = message.subject
+            self.assertTrue('FETC-GW' in subject)
```
tymees, line 54: Will this always return the Dutch version? The test passes, but maybe not in the future

reviews/utils/review_utils.py
```diff
@@ -153,8 +154,8 @@ def remind_reviewers():
 
     for decision in decisions:
         proposal = decision.review.proposal
-        subject = "FETC-GW: beoordelen aanvraag {ref_num} (herinnering)".format(
-            ref_num=proposal.reference_number
+        subject = "Herinnering: beoordeel aanvraag {}".format(
```
tymees, line 157: Missing FETC-GW prefix?

tymees (approved):
Although I made some comments, I don't think any of them are definite 'need changes' comments (just some suggestions basically). I'm happy with merging this as-is

reviews/utils/review_utils.py
miggol: Aesthetic choice in this case

PR #349, opened 2022-06-08
Feature/remove unfinished pdf action

miggol:
This removes the PDF link for draft proposals. Because this does not entirely solve the problem of PDFs looking a little too official, I've added a status line to the top of the proposal PDF template. This should tell anyone viewing the PDF if this proposal was actually accepted.

This also includes a new Proposal method that returns the status as a translated string. Useful for templates and the like. I've named it the same as the equivalent method in Vue.

### DIFF
--- a/locale/en/LC_MESSAGES/django.po
+++ b/locale/en/LC_MESSAGES/django.po
@@ -6,7 +6,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: \n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2022-06-08 15:53+0200\n"
+"POT-Creation-Date: 2022-07-12 12:06+0200\n"
 "PO-Revision-Date: 2020-10-07 11:22+0200\n"
 "Last-Translator: Anna Asbury <anna.asbury@annaasbury.com>\n"
 "Language-Team: \n"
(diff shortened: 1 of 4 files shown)


proposals/templates/proposals/vue_templates/proposal_list.html
```diff
@@ -64,24 +64,17 @@ <h4>
...
-            <a
-                v-else
-                :href="$url('proposals:pdf', [proposal.pk])"
-                target="_blank"
-            >
-                <img src="{{ img_pdf }}"
-                         title="{% trans 'Inzien' %}">
-            </a>
-
+</template>
```
tymees, line 77: So two things:
1. Please indent this properly, or:
2. Remove this, as you don't need this template tag here. You can directly apply this v-if to the <a> element. 

The <template> example above is for clarity, as there are 2 links with the same condition. <template> is a no-op Vue tag that allows you to group them in your template, without adding something in the output. (Which is needed here for CSS reasons)

locale/en/LC_MESSAGES/django.po
```diff
@@ -4311,6 +4310,37 @@ msgstr "Pending decisions committee members"
...
+msgstr ""
+
+#. Translators: The 'lookup_type' is one of 'date', 'year' or 'month'.
+#. Eg: "Title must be unique for pub_date year"
+#: src/uil-django-core/uil/rest_client/fields.py:35
+#, python-format
+msgid ""
+"%(field_label)s must be unique for %(date_field_label)s %(lookup_type)s."
+msgstr ""
+
```
tymees, line 4343: Uh what?

proposals/templates/proposals/vue_templates/proposal_list.html
miggol: Noted, thanks. Removed the template tag

PR #351, opened 2022-06-09
Feature/self assesment field

Meesch:
Adds a field to the Trajectory overview form page for researchers to self-asses their research. Also includes some backlog migrations.


tymees:
Looks good overall, but there's something missing. (My bad, should've told you).

The new field should also be added to the PDF of the proposal, otherwise the committee's won't (always) see it

Meesch:
> 

Hmm, ik heb het proposal_pdf.html template gevonden en daar heb ik het veld toegevoegd, maar ik kan nu niet zo 1-2-3 vinden wat ik nog meer moet aanpassen. De pdfs die ik genereer hebben echter nog geen nieuw veld erop...

Meesch:
Oh grapje hij doet het gewoon -.-

tymees (approved):

miggol:
In retrospect, this field should not have been added to the study model but to the actual proposal model. We only want one answer to this question per proposal. This means we might have to make an entirely new page.

PR #355, opened 2022-06-24
Feature/self assesment field

Meesch:
This PR re-implements the self_assesment field, this time for the whole proposal, on the general page.


tymees:
The field should also be added to the diff pages (proposals/templates/proposals/diff)

Meesch:
Ok boys, I think everything has been updated that needed updating. The only thing I am still unsure about is the conflict in locale/en/LC_MESSAGES/django.po... I sure don't understand what any of it means, and you probably do, so I await your input.

PR #356, opened 2022-06-24
add csrf error message page

Meesch:
Adds an error message page for CSRF errors:

Screenshot 2022-06-24 at 13 14 26

Let me know what you think!


miggol:
Thanks for picking this up! The majority of the time I would say that CSRF errors come from submitting a very stale form. So that users don't lose data they may have entered, we might want to recommend they hit the back button first and *then* reload. Most browsers will remember their changes, but the form will now have a fresh token.

Oh and please add a HTML title to the error template: `{% block header_title %}`

tymees:
> So that users don't lose data they may have entered, we might want to recommend they hit the back button first and _then_ reload.
> Most browsers will remember their changes, but the form will now have a fresh token.

Behaviour amongst browsers is somewhat inconsistent with hidden field values when you do that, it's even inconsistent within the same browser. (Depending on cache). Thus, you might not even get a new CSRF token that way, resulting in getting this error again. A fresh page is the only way to resolve this reliably. 

However, we could probably change out the instructions with a link instead. AFAIK you're still on the right url, so we could just create a hyperlink to the current page, and do a fresh request that way. (Which might be more user friendly); Yes, a user will lose changes, but considering this error only happens very rarely (not even yearly), I don't think it matters.

Meesch:
Ok feedback has been processed and I think this PR is ready to be merged! Let me know if you notice anything else!

PR #359, opened 2022-07-06
Feature/self assessment field michael

miggol:
I'm suggesting some changes to Mees' branch. I thought this was too code-heavy for a PR comment

Note:
* Commas at the end of multi-line lists, this helps a lot with diffs!
* "assesment" typo
* The idea behind soft validation is to not inhibit progress. I don't believe mark_soft_required is desired here, and prefer that only final submission is prohibited. This is why I manually add a required error. Though this is also a design decision and quite debatable.


miggol:
Multi-line list style was in fact advice once handed down to me by Ty, now changing hands once again.

Meesch (approved):
Nice, absolutely agree with making the requirement only for the final submission. I'll continue finishing the field in the og branch!

PR #366, opened 2022-07-19
Update master from develop

miggol:


miggol:
All seems well on acceptance, including PDF and compare pages, so I'm going ahead with the update.

PR #370, opened 2022-07-27
Feature/in revision list

### LINKED ISSUE #368: 'In revision' page for secretary
The secretary has requested a new overview page for proposals that:
- Have been send back for revision
- And have not received a new version that's been submitted

Intended for the secretary to know which proposals are being revised at the moment

miggol:
Apologies for the somewhat mind-bending queryset, this is the most readable I could get it.

Fixes #368

### DIFF
--- a/reviews/api/views.py
+++ b/reviews/api/views.py
@@ -430,6 +430,7 @@ class InRevisionApiView(BaseReviewApiView):
         # has a child with a revision review
         in_revision = candidates.exclude(
             Exists(revision_reviews.filter(
+                # OuterRef refers to the candidate being evaluated
                 proposal__parent__pk=OuterRef("proposal__pk")
                 )
             )
--- a/reviews/api/views.py
+++ b/reviews/api/views.py
@@ -414,7 +414,7 @@ class InRevisionApiView(BaseReviewApiView):
 
     def get_queryset(self):
         # 1. Find reviews of revisions:
-        # A proposal not having one of these means
+        # A revision not having a review means
         # that the review of its parent is "in revision"
         revision_reviews = Review.objects.filter(
             proposal__is_revision=True,    # Not a copy
@@ -422,7 +422,9 @@ class InRevisionApiView(BaseReviewApiView):
         )
         # 2. Get candidate reviews:
         # All reviews whose conclusion is "revision necessary"
+        # that are in the current committee
         candidates = Review.objects.filter(
+            proposal__reviewing_committee=self.committee,
             stage=Review.CLOSED,
             continuation=Review.REVISION,
         )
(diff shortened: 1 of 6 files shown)


reviews/api/views.py
```diff
@@ -405,6 +405,38 @@ def get_queryset(self):
...
+        revision_reviews = Review.objects.filter(
+            proposal__is_revision=True,    # Not a copy
+            stage__gte=Review.ASSIGNMENT,  # Not a supervisor review
+        )
+        # 2. Get candidate reviews:
+        # All reviews whose conclusion is "revision necessary"
+        candidates = Review.objects.filter(
+            stage=Review.CLOSED,
+            continuation=Review.REVISION,
+        )
```
tymees, line 428: For readability I'd suggest swapping these qs arround; first get the dataset you're going to filter (and return), then the dataset that will be used to filter.

I instinctively go for that order, and it threw me off for a second that it's reversed.

reviews/api/views.py
```diff
@@ -405,6 +405,38 @@ def get_queryset(self):
...
+        # 3. Finally, exclude candidates whose proposal
+        # has a child with a revision review
+        in_revision = candidates.exclude(
+            Exists(revision_reviews.filter(
+                proposal__parent__pk=OuterRef("proposal__pk")
+                )
+            )
+        )
+        return in_revision
```
tymees, line 438: Style thing, feel free to ignore;
I prefer just returning directly instead of assign var and immediatly returning said var.

reviews/api/views.py
```diff
@@ -405,6 +405,38 @@ def get_queryset(self):
...
+        # A proposal not having one of these means
+        # that the review of its parent is "in revision"
+        revision_reviews = Review.objects.filter(
+            proposal__is_revision=True,    # Not a copy
+            stage__gte=Review.ASSIGNMENT,  # Not a supervisor review
+        )
+        # 2. Get candidate reviews:
+        # All reviews whose conclusion is "revision necessary"
+        candidates = Review.objects.filter(
+            stage=Review.CLOSED,
```
tymees, line 426: I'm pretty sure you forgot to filter by commission?

tymees (changes requested):
Mind bending query indeed, but aside from one minor oversight it seems correct. 

Two stylistic comments can be ignored if you disagree, but that oversight (not filtering on committee) needs to be fixed. After that it's fine to merge imo

reviews/api/views.py
miggol: Yes I completely forgot

reviews/api/views.py
miggol: Yeah I started out that way, and ended up greatly preferring this. It's probably personal preference.

tymees (approved):

PR #374, opened 2022-08-03
Update master from develop

miggol:
Let's do these more often. For now this contains:

- In revision list
- Task model fix


tymees (approved):

miggol:
Merged this without thinking about it too much, but I won't deploy it until the start of next week for the weekend's sake.

PR #376, opened 2022-08-11
fix: prevent discontinueing of accepted proposals

### LINKED ISSUE #354: Prevent already closed applications being 'terminated'
I just noticed the option to definitively close an application is also available for applications that have been approved (and have thus been 'closed' already).

tymees:
Fixes #354

### DIFF
--- a/reviews/utils/review_actions.py
+++ b/reviews/utils/review_actions.py
@@ -152,13 +152,11 @@ class CloseReview(ReviewAction):
         return _('Deze aanvraag afsluiten')
 
 
-
-
 class DiscontinueReview(ReviewAction):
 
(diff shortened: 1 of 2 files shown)


miggol (approved):
Looks good! Good job thinking of making the view inaccessible. It's given me the idea to make an ActionMixin sometime in the future, which integrates view permissions with an action's `is_available()`.

tymees:
> It's given me the idea to make an ActionMixin sometime in the future, which integrates view permissions with an action's `is_available()`.

Yeah I thought about adding something like that as well, as I was essentially copying code (which is just annoying to maintain). However, I was too lazy to actually do it

Issue #352, opened 2022-06-13
Show warning when "intended start date" lies within two weeks of today


miggol:
In such cases, the FETC won't be able to provide official approval, and submitters should be warned before submission. We should show a warning on the submission page and on the first page where the field is located.

The warning on the submission page can easily be made dependant on user input. However, a warning on the same page as the field might be a hassle to make dynamically, and so a permanent warning would be fine there too.

tymees:
> The warning on the submission page can easily be made dependant on user input. However, a warning on the same page as the field might be a hassle to make dynamically, and so a permanent warning would be fine there too.

My first thought would be 'why not just add a validiation step in clean', but ofc that would block submission...

But I think making a bit more effort to make the warning on the same page dynamic is worth the investment. A static warning will be ignored unless we make it visually 'scary', but that would cause undue stress for those who the warning does not apply.

The WMO page already does something similar, we might be able to hook into that?

djhcapel:
Could this issue also get priority? It will save both the secretary and the applicant (a lot of) time if applicants are warned before they actually submit their application. It does happen regularly that applicants fill in a date that lies in the past or forget to update the 'intended start' date during revision.

miggol:
I've started work on this. Submission page now shows the appropriate warning.

> But I think making a bit more effort to make the warning on the same page dynamic is worth the investment. 

From my call with Desiree about this issue, which admittedly was a while back, I think we concluded this information should always be visible. Most submitters never read the FETC guidelines, and this tidbit is important enough to always emphasize. This is generally educational, but also important for submitters who first create their proposal with plenty of time, work on it for a few weeks, then finally submit after the deadline. They would have wanted to know about the two weeks from the start!

> A static warning will be ignored unless we make it visually 'scary', but that would cause undue stress for those who the warning does not apply.

This is a good point. I will look into adding a dynamic visual emphasis to the message if the deadline is close or has passed.

> The WMO page already does something similar, we might be able to hook into that?

That's a good tip, I will be basing whatever it becomes off of that!

PR #379, opened 2022-08-12
Feature/start date warning

miggol:
This fix for #352 seems like a good candidate to go straight to acceptation. It doesn't touch anything critical, but the wording could use a good look-over by @djhcapel.

Current state: Intended start date warning is always shown in the form. When input is actually within two weeks of today, the warning turns red. On the submission page the warning is only shown when relevant, with a link back to the first page.

### DIFF
--- a/proposals/templates/proposals/proposal_form.html
+++ b/proposals/templates/proposals/proposal_form.html
@@ -4,169 +4,170 @@
 {% load i18n %}
 
 {% block header_title %}
-    {% trans "Algemene informatie over de aanvraag" %} - {{ block.super }}
+{% trans "Algemene informatie over de aanvraag" %} - {{ block.super }}
 {% endblock %}
 
 {% block html_head %}
-    <script>
-        $(function () {
-            check_field_required('relation', 'needs_supervisor', 'supervisor', 'proposals');
-            depends_on_value('other_applicants', 'True', 'applicants');
-            depends_on_value('other_stakeholders', 'True', 'stakeholders');
-            check_field_required('funding', 'needs_details', 'funding_details', 'proposals');
-            check_field_required('funding', 'needs_name', 'funding_name', 'proposals');
+<script>
+$(function () {
+  check_field_required('relation', 'needs_supervisor', 'supervisor', 'proposals');
+  depends_on_value('other_applicants', 'True', 'applicants');
+  depends_on_value('other_stakeholders', 'True', 'stakeholders');
+  check_field_required('funding', 'needs_details', 'funding_details', 'proposals');
+  check_field_required('funding', 'needs_name', 'funding_name', 'proposals');
 
-            // Add datepicker for date_start, set locale to current language
-            $.datepicker.setDefaults($.datepicker.regional["{{ LANGUAGE_CODE }}"]);
+  // Add datepicker for date_start, set locale to current language
+  $.datepicker.setDefaults($.datepicker.regional["{{ LANGUAGE_CODE }}"]);
 
-            var date_format = '{{ LANGUAGE_CODE }}' === 'nl' ? 'dd-mm-yy' : 'mm/dd/yy';
-            $("#id_date_start").datepicker({
-                dateFormat: date_format,
-                minDate: 'now',
-            })
+  var date_format = '{{ LANGUAGE_CODE }}' === 'nl' ? 'dd-mm-yy' : 'mm/dd/yy';
+  $("#id_date_start").datepicker({
+    dateFormat: date_format,
+    minDate: 'now',
+  })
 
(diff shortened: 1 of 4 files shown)


proposals/templates/proposals/proposal_form.html
```diff
@@ -101,11 +101,72 @@ <h2>{% trans "Algemene informatie over de aanvraag" %}</h2>
...
+     .red {
+	 color: red;
+     }
+     #date_start_warning {
+	 margin-top: 0.5rem;
+	 margin-bottom: 0.5rem;
+	 line-height: 1.5;
+     }
+    </style>
+    <script>
```
tymees, line 122: Indentation in this script block could use some work. Everything seems to be on the same indentation level

proposals/templates/proposals/proposal_form.html
```diff
@@ -101,11 +101,72 @@ <h2>{% trans "Algemene informatie over de aanvraag" %}</h2>
...
+     }
+
+     /* Define warning update function */
+     function updateWarning()
+     {
+	 if (checkDateStart())
+	 {
+	     date_start_warning.removeClass("red");
+	 } else {
+	     date_start_warning.addClass("red");
```
tymees, line 157: Please use ``.text-danger`` instead. It uses a nicer shade of red, and will make the new BS theme work better

Also, you might want to use [toggleClass](https://api.jquery.com/toggleclass/)

proposals/templates/proposals/proposal_form.html
```diff
@@ -101,11 +101,72 @@ <h2>{% trans "Algemene informatie over de aanvraag" %}</h2>
...
         </div>
     </div>
+    <!-- Code for intended start date warning -->
+    <style>
+     .red {
+	 color: red;
+     }
+     #date_start_warning {
+	 margin-top: 0.5rem;
+	 margin-bottom: 0.5rem;
```
tymees, line 118: Why not use ``.mt-2`` and ``.mb-2``?

(I'm trying to use Bootstrap utilites more instead of custom classes if possible)

proposals/templates/proposals/proposal_form.html
```diff
@@ -101,11 +101,72 @@ <h2>{% trans "Algemene informatie over de aanvraag" %}</h2>
...
+	 line-height: 1.5;
```
tymees, line 119: Line-height should already be 1.5 right?

proposals/templates/proposals/proposal_form.html
tymees: Also, you're somewhat inconsistat with your semicolors... I don't mind it that much, but consistancy is nice

tymees (changes requested):
Logic seems fine (although time calculations do break my brean sometimes).

There are some code style and CSS issues that I'd like to see adressed first

proposals/templates/proposals/proposal_form.html
miggol: I just noticed it's mixed tabbed / spaced indentation. Apologies for that. Will convert it all to spaces in the future.

proposals/templates/proposals/proposal_form.html
miggol: No it is 1.8, inherited from top-level <p>s in uu-style.css

proposals/templates/proposals/proposal_form.html
miggol: No reason, fixed

proposals/templates/proposals/proposal_form.html
miggol: I added some more semicolons for good measure

proposals/templates/proposals/proposal_form.html
tymees: Really? Hmm

tymees:
Did you use a formatter tool? And is that tool set to use 2 spaces instead of 4?

Sorry if I'm nagging, but we use 4 spaces everywhere and the reformat now adds 2 spaced indentation everywhere...

miggol:
Yes the two spaces are intentional. IMO four spaces is too much for HTML.

tymees:
IMO I think two spaces are hard to read. So we disagree on that point.

I'm very willing to discuss switching indentation project wide, but I don't think it's a good idea to just switch one file in an indentation-unrelated PR. (Especially as my editor will just switch back to four spaces the next time I edit that file, causing more commit-gore).

That discussion aside, was the removal of indentation around Django tags (if/with/blocktrans/etc) intentional as well?

miggol (commented):
I added indentation to the Django tags but intentionally kept everything at two spaces per level

tymees (approved):

Issue #269, opened 2021-11-18
The success_url after making a decision should be the review details page


miggol:
As stated in title

tymees:
In all cases, or just when a committee member makes a decision?

miggol:
I'm thinking yes also for a secretary final decision, but not for a supervisor decision because that would result in a permission denied situation. Or am I forgetting something obvious?

There are a couple more success_urls we could change to be more consistent. Like ChangeChamberView and ReviewAssignView.

tymees:
> I'm thinking yes also for a secretary final decision, but not for a supervisor decision because that would result in a permission denied situation. Or am I forgetting something obvious?

Hm yes that is what I was suggesting. I should've said 'all cases except supervisor decisions'. The ReviewDetailView isn't accessible for supervisors.

> There are a couple more success_urls we could change to be more consistent. Like ChangeChamberView and ReviewAssignView.

Good point, let's redirect them to the detail view as well!

Meesch:
Addressed with PR #383

PR #383, opened 2022-08-16
update successurls except supervisor revision

Meesch:
Addresses issue #269. Redirects user to the details page of the review after making a decision, changing the reviewing chamber, or changing reviewers.

### DIFF
--- a/reviews/views.py
+++ b/reviews/views.py
@@ -242,7 +242,6 @@ class ReviewAssignView(GroupRequiredMixin, AutoReviewMixin, generic.UpdateView):
     group_required = settings.GROUP_SECRETARY
 
     def get_success_url(self):
-        committee = self.object.proposal.reviewing_committee.name
         return reverse('reviews:detail', args=[self.object.pk])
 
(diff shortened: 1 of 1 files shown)


miggol (approved):
Looks good!

reviews/views.py
```diff
@@ -243,7 +243,7 @@ class ReviewAssignView(GroupRequiredMixin, AutoReviewMixin, generic.UpdateView):
 
     def get_success_url(self):
         committee = self.object.proposal.reviewing_committee.name
```
tymees, line 245: Pretty sure this line can be removed

Same comment applies to the other modified ``get_success_url``

Meesch: You mean because the 'committee' is not referenced anywhere else? I was not sure what it is meant to do, so I left it alone...

tymees: Yes; Previously it was used as a variable for the URL (to make sure you went to the right version of the page).

Now it's not needed anymore, thus should be removed

Issue #350, opened 2022-06-09
Tag reviews as supervisor reviews properly


tymees:
Right now, we can distinguish committee reviews from supervisor reviews by it's `stage`. Supervisor reviews will always stay on that stage, even when they're concluded. 

That's counter intuitive, as you'd think they should reach a 'concluded' stage at some point. However, if one (accidentally) actually reaches that point the portal will treat it as a committee review. 

Thus, I think we should add a field to the Review model that explicitly states what kind of review it is. Maybe even increase the scope and try to sort out this whole review mess a bit, but it might be better to keep it small :P

Toughts @miggol @Meesch ?

tymees:
Ideally I'd like a new field, as the whole point is to _easily_ differentiate the two. If we'd add new stages and start using those, it's going to be even _worse_ to find the correct set. For example, it's quite common atm to do something like this:
````python
reviews = Review.objects.filter(stage__gte=Review.COMMISSION)
````
to get a list of all reviews that are being reviewed or have advanced even further during the process. This is easy, as the stage field is a numeric field with every stage given a number (with higher numbers being 'later' in the process)

If we'd add more stages specifically for those two variants, we're going to need to do something like this to accomplish the same:
````python
reviews = Review.objects.filter(Q(stage=Review.COMMISSION) | Q(stage=Review.CLOSING) | Q(stage=Review.CLOSED))
````
Not really that elegant and not to mention very error-prone...

Alternatively, having a new field would result in something like this:
````python
reviews = Review.objects.filter(stage__gte=Review.COMMISSION, is_commission_review=True)
````
This is way more descriptive in my opinion, and way more fault tolerant

tymees:
Oh also, this issue was sparked by a Review suddenly changing it's stage to ``Review.COMMISSION``, even though that should be impossible. So I don't think having different stages would really help in that regard :P

Meesch:
Fair enough. Let's create a new field!

miggol:
Somewhat related: if you're an application's supervisor and also a committee reviewer, you always get the supervisor decide view. Besides being a huge conflict of interest, this is of course a weird edge case. But reading this issue makes me realize why the decide view works that way at all. Indeed, there should be an unambiguous way to tell what kind of review you're dealing with.

tymees:
Just adding a comment, to indicate urgency:

[...]

PR #387, opened 2022-08-25
[RFC] New supervisor review type

miggol:
In response to #350 - Tag reviews as supervisor reviews properly

Reviews now have a type, creating two mutually exclusive categories of reviews. To me, this makes more sense than a boolean flag, and this allows for future extensibility. Now, supervisor reviews can also make use of the "closed" stage, which should allow us to model all supervisor review states using `review.stage` and `review.go`. So we can now more easily distinguish open and closed supervisor reviews as well. The `continuation` field is unused for supervisor reviews, for now.

This is just meant as a place to propose changes, obviously a lot more plumbing has to happen in the review creation and update process. But we might as well do this right and think ahead.


reviews/models.py
```diff
@@ -40,6 +40,19 @@ class Review(models.Model):
         (DISCONTINUED, _('Niet verder in behandeling genomen')),
     )
 
+    REVIEW_TYPES = [
+        ("supervisor", _("Beoordeling door eindverantwoordelijke")),
+        ("committee", _("Beoordeling door commissie")),
+    ]
```
tymees, line 46: I'd like to see the type values extracted as class constants (like the stages are). That way you can refer to them like this:
``Review.TYPE_SUPERVISOR`` (name is not great, just for illustration purposes).

reviews/migrations/0014_auto_20220824_1628.py
```diff
@@ -0,0 +1,35 @@
+# Generated by Django 3.2.14 on 2022-08-24 14:28
+
+from django.db import migrations
+
+
+def choose_review_type(apps, schema_editor):
```
tymees, line 6: This migration should work in theory, but I'm a bit afraid of the 'in practice' part. The DB is full of junk...

Not that you should change anything, I'm just voicing my concern and suggesting we should keep an eye on it when deploying it :P

tymees (commented):
Yes, this is definitly a better idea than using a boolean.

So as a course of action, I fully approve.

As a PR, not so much. :P Maybe convert the PR to a Draft PR for clarity?

reviews/migrations/0014_auto_20220824_1628.py
miggol: > This migration should work in theory

If I had the migrations the right way round, then maybe! The order has now been fixed, and `null` is allowed on `review_type` to span the intermediate deadzone.

As an aside, I have now learned that constants like `review.SUPERVISOR` are actually not available in historical models such as those you get in data migrations, unless they are contained in a Meta class. That's why there's a hardcoded 0 there now.

miggol:
Closing as deprecated

PR #395, opened 2022-09-01
chore: update to DSC 2.0.10

tymees:
DSC 2.0.10 fixes a small issue with help-text tooltips being placed on the wrong elements. 

Normally I'd not update the old unsupported version anymore, but this was affecting a required change that cannot wait till DSC 3


miggol:
Can't say I understand the bug, but whatever, this fixes it!

PR #423, opened 2022-10-25
Update acceptation from develop

miggol:
There are still blocking issues, so this is a draft for now. If we finish small textual things they can still go to acc and end up in prod before this merges.

Blocking:
- Nothing anymore!


miggol:
Merging this tomorrow, wednesday 07 feb

PR #424, opened 2022-10-25
Feature/pdf fields update

Meesch:
Adds fields to pdf from last week's PRs


tymees:
Maybe for a different PR, but I think these fields need to be added to the compare page as well?

miggol (changes requested):
The hierarchy model field is currently a `BooleanField` with `blank=True` which means its values are _True_, _False,_ or _None._ These values get outputted literally in templates:

image

These are kind of weird answers, and they're also untranslated. So we prefer to avoid those. Elsewhere in the models you will usually find `CharField`s with `choices=YES_NO_DOUBT` or the [`|yesno`](https://docs.djangoproject.com/en/3.2/ref/templates/builtins/#yesno) filter.

proposals/templates/proposals/proposal_diff.html
```diff
@@ -65,6 +65,29 @@ <h2 class="mt-3 mb-1">{% trans "Algemene informatie over de aanvraag" %}</h2>
...
                     {% endif %}
+                    {%  if proposal.relation.check_in_course or p_proposal.relation.check_in_course%}
+                    <tr>
+                        <th>{% get_verbose_field_name "proposals" "proposal" "student_program" %}</th>
+                        <td>{{ p_proposal.student_program }}</td>
+                        <td>{{ proposal.student_program }}</td>
+                    </tr>
+                    <tr>
+                        <th>{% get_verbose_field_name "proposals" "proposal" "student_context" %}</th>
+                        <td>{{ p_proposal.student_context }}</td>
```
miggol, line 76: Edge case: if unselected in the previous revision, `p_proposal.student_context` will display an untranslated `"None"`, it would be nicer if this just displayed an empty string.

This can be done with a filter, or maybe with `get_FIELDNAME_display()` because it's a choice (not sure about the latter).

https://docs.djangoproject.com/en/4.1/ref/models/instances/#django.db.models.Model.get_FOO_display

miggol (changes requested):
Some small things adjust still. The quirks of displaying fields correctly in Django come up a lot.

Meesch:
I think I addressed all your comments, but let me know if you see anything else. Thanks for letting me know about filters!

tymees:
IMO it looks good now. @miggol, agreed?

proposals/templates/proposals/proposal_diff.html
```diff
@@ -65,6 +65,31 @@ <h2 class="mt-3 mb-1">{% trans "Algemene informatie over de aanvraag" %}</h2>
                             <td>{{ proposal.supervisor.get_full_name }}</td>
                         </tr>
                     {% endif %}
+                    {%  if proposal.relation.check_in_course or p_proposal.relation.check_in_course%}
```
miggol, line 68: Lack of space before closing "%}" caught my eye

miggol (approved):
PR looks good!

PR #436, opened 2022-11-11
Serve user uploads with new UserMediaView

miggol:
This prohibits non-logged-in users from accessing them

### DIFF
--- a/fetc/urls.py
+++ b/fetc/urls.py
@@ -10,6 +10,8 @@ from django.contrib import admin
 from django.contrib.auth import views as auth_views
 from django.urls import path, include
 
+from main.views import UserMediaView
+
 handler404 = 'main.error_views.error_404'
 handler500 = 'main.error_views.error_500'
 handler403 = 'main.error_views.error_403'
@@ -19,6 +21,16 @@ handler400 = 'main.error_views.error_400'
 urlpatterns = [
     path('accounts/login/', auth_views.LoginView.as_view(), name='login'),
     path('accounts/logout/', auth_views.LogoutView.as_view(), name='logout'),
+
+    # Access user uploads
+    path('media/<str:filename>',
+         UserMediaView.as_view(),
+         name='user_upload'),
+    # And a duplicate path for user media
+    # Use this for testing on servers that rewrite /media/
+    path('view_media/<str:filename>',
+         UserMediaView.as_view(),
+         name='user_upload_2'),
 
     path('', include('main.urls')),
     path('proposals/', include('proposals.urls')),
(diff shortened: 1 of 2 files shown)


main/views.py
```diff
@@ -449,6 +452,27 @@ def delete(self, request, *args, **kwargs):
         return super(DeleteView, self).delete(request, *args, **kwargs)
 
 
+class UserMediaView(LoginRequiredMixin, generic.View):
+    """Respond with a given user uploaded file if the request
+    is logged in. Makes it so that user uploads are no longer
+    public."""
+
+    def get_response_file(self, filename):
+        filepath = os.path.join(settings.MEDIA_ROOT, filename)
```
tymees, line 461: As noted in the chat, this is dangerous. An attacker can craft a filename that will result in arbitrary file access on the server. (outside MEDIA_ROOT)

It's not a OMG WE ARE DOOMED scenario, we're still behind a LoginRequiredMixin and Apache seems to be filtering out the more simple malicious filenames, but I still really don't want to rely on those protections alone. Especially as the fix is one if statement ;)

Thus, there should be a check for making sure that the resulting filepath is still inside MEDIA_ROOT.

fetc/urls.py
```diff
@@ -20,6 +22,16 @@
...
 
+    # Access user uploads
+    path('media/<str:filename>',
+         UserMediaView.as_view(),
+         name='user_upload'),
+    # And a duplicate path for user media
+    # Use this for testing on servers that rewrite /media/
+    path('view_media/<str:filename>',
+         UserMediaView.as_view(),
+         name='user_upload_2'),
```
tymees, line 33: Is this still necessary?

It feels like a fix for uWSGI's static-map, but we are removing that anyways. In addition, we currently don't really control the urls presented to the user, sooooo...

tymees (changes requested):

main/views.py
tymees: For those interested, it's called a path traversal attack:
https://owasp.org/www-community/attacks/Path_Traversal

main/views.py
```diff
@@ -449,6 +452,27 @@ def delete(self, request, *args, **kwargs):
...
+        try:
```
Meesch, line 462:
```suggestion
if filename in os.listdir(settings.MEDIA_ROOT):  # check if file exists in media root
```

Would this be sufficient sanitization?

Meesch (commented):
Other than Ty's comment I have tested it locally and everything works as expected!

tymees (approved):

PR #439, opened 2022-11-21
Feature/remove passive consent

miggol:
Passive informed consent is no longer used in the regulations. Precisely how research based on Algemeen Belang should replace passive IC is yet to be decided. But in the meantime, all reference to passive IC should be removed from the portal to prevent confusion.

Passive IC has been removed from:
* Study forms and templates
* ProposalSubmit view and template
* Auto review reasons
* Templates where necessary

The way we used to require IC documents is kind of complicated and had some friction with reality. Validation has been changed to the following:
* The school-related consent document fields now appear if and only if schools are among the settings
* The school-related consent document fields are always optional
* The regular consent documents fields are now always required (They used to disappear when passive IC was selected)

Specifically I'm looking for feedback from @djhcapel on the handling of consent documents before merging. It should be noted that currently in production we have made all study settings "not schools" to avoid the school-related IC documents being demanded. With the school IC documents now being an optional extra, we could choose to undo this.


miggol:
The `auto_review` tests in `reviews/tests.py` are truly an abomination. I've started rewriting them in a different branch. For now `passive_consent` has just been cut out.

Meesch:
I would hold off on develop the `auto_review` tests for now, since I am removing several of them for the new regulation update!

miggol:
I'm expecting @djhcapel to write a small infobox that appears when a school is selected. Then this can get merged

djhcapel:
In the practice portal, I noticed that different sets are requested for different age groups (situation: recruiting and test at schools):
16+: IL and DoC participants ánd school management
6-11 years ánd 12-15 years: IL and DoC participants only.

**Documents for participants:**
The sets for the 16+ age group are okay.
The sets for the 6-11 and 12-15 groups are incomplete, see below: the documents for the management of the institution are missing. 

Further, in case the data collection concerns underage participants (I do not know if the text below can be displayed only when these age groups are ticked???), it should be stated somewhere that:

- If children are 11 years old or younger: only parents/guardians need to sign the declaration of consent.
- If children are between 12 and 15 years of age, both parents/guardians and the child needs to sign the declaration of consent. You may consider composing separate information letters for children on the one hand and parents/guardians on the other hand: the extra letter can be uploaded as an 'additional form'. (Note: two documents then have to be uploaded; in case you only have one additional form, please upload an empty document as the second form.)
- Children of 16 years and older may sign the declaration of consent independently of their parents/guardians.

**Documents for management:**
In all situations in which someone should consent to the research being conducted at their premises, a second set of documents should be uploaded, i.e. an IL and a DoC for the management of that institution. 

This not only concerns schools, but also 'peuterspeelzaal/dagopvang/buitenschoolse opvang', ' logopediepraktijk' and 'zorginstelling'.  

In those cases, the test for the second set of forms should not read: 'Toestemmingsverklaring van de schoolleider/hoofd van het departement' (which was incorrect anyway, cause it also concerns the IL), but:

Documenten voor de leiding/het management van de instelling:

Upload hier de informatiebrief voor de leiding/het management van de instelling(in .pdf of .doc(x)-formaat):
Upload hier de toestemmingsverklaring voor de leiding/het management van de instelling(in .pdf of .doc(x)-formaat):

djhcapel:
> I'm expecting @djhcapel to write a small infobox that appears when a school is selected. Then this can get merged

I realized that the three bullet points above (If children..... parent/guardians) should not only appear in case of schools, but in all situations in which underage participants are involved.

studies/models.py
```diff
@@ -234,25 +234,7 @@ class Study(models.Model):
...
-target="_blank">de FETC-GW-website</a>.')),
-        null=True,
-        blank=True,
-    )
-    passive_consent_details = models.TextField(
-        _('Licht je antwoord toe. Wij willen je wijzen op het reglement, \
-sectie 3.1 \'d\' en \'e\'. Passive consent is slechts in enkele gevallen \
-toegestaan en draagt niet de voorkeur van de commissie.'),
-        blank=True)
-
```
tymees, line 255: This will break backwards compatibility!

(Thus making #446 more important, but we don't have #446 thus we cannot do this)

proposals/templates/proposals/proposal_pdf.html
```diff
@@ -415,18 +415,10 @@ <h2>{% trans "Informed consent formulieren" %}</h2>
                 <th>{% get_verbose_field_name "studies" "documents" "briefing" %}</th><td><a href="{{ BASE_URL }}{{ study_documents.briefing.url }}">{% trans "Download" %}</a></td>
             </tr>
             {% endif %}
-            <tr>
-                <th>{% get_verbose_field_name "studies" "study" "passive_consent" %}</th><td>{{ study.passive_consent|yesno:_("ja,nee") }}</td>
-            </tr>
-            {% if study.passive_consent %}
-                <tr>
-                    <th>{% get_verbose_field_name "studies" "study" "passive_consent_details" %}</th><td>{{ study.passive_consent_details }}</td>
-                </tr>
```
tymees, line 424: Ideally we keep this in the PDF for older studies.

Moving the passive consent field below the if-statement would make it disappear if it's not used in the study, thus at least keeping the fields in the PDF's that do use passive consent.

proposals/forms.py
```diff
@@ -616,35 +616,11 @@ def clean(self):
...
-
-                if study.needs_additional_external_forms():
-                    if not documents.director_consent_declaration:
-                        self.add_error('comments', _(
-                            'Toestemmingsverklaring voor traject {} nog niet toegevoegd.').format(
-                            study.order))
-                    if not documents.director_consent_information:
-                        self.add_error('comments', _(
-                            'Informatiebrief voor traject {} nog niet toegevoegd.').format(
-                            study.order))
```
tymees, line 650: I think we still need this code? Desiree's last comment seems to suggest there forms are still required in some cases?

tymees (changes requested):
The PR as-is will alter older studies in the archive, and as we don't have a good, consistent, way to handle this atm we should not do this

proposals/forms.py
miggol: Sadly that comment goes a bit beyond the scope of this PR. Once algemeen belang has been sorted out we can always make them required again.

tymees: Fair enough, let's do that in a different PR then if necessary

studies/models.py
miggol: Very good point. I've restored this and removed the migration.

proposals/templates/proposals/proposal_pdf.html
miggol: This is now fixed with a little extra logic!

Passive consent will show up in PDF's and diffs it has any value set other than `None`. So that will include older proposals with a decisive No answer. The school forms will now always show up in the PDF if present, regardless of the newer logic for `Study.needs_additional_external_forms()`. I think that catches everything.

proposals/templates/proposals/study_consent.html
```diff
@@ -5,138 +5,150 @@
...
-                return false;
-            });
-
-            // Show the first add link
-            $("a.new").first().show();
-        });
-    </script>
+  <script>
+    // This function hides school-related extra forms
+    function hide_school_forms(prefix) {
```
Meesch, line 14: So, weirdly this does not work at all on my machine... I have tried this with Intervention, Observation, and Session studies, selected a school setting, nothing I did made this section appear on the page. I have tried some rudimentary debugging and the problem as far as I could trace it lead me to the `research_settings_contains_schools(self):` method in studies/models.py (line 386). It seems to be that there are no `settings_contains_schools` attributes for intervention or observation objects, nor is there a `setting__is_school` attribute in the session_set objects (also that attribute name contains a double underscore I don't know if that's a typo). It's 17:30 here and I am fried so that is how far I'm going to take this for today ;-)

Meesch (changes requested):

proposals/templates/proposals/study_consent.html
tymees: Double underscores in ORM filters can be seen as a ``.``, thus it would be equivalent to ``setting.is_school`` in python. 
Basically, you cannot use periods in parameter names (or names in general) but it is needed for nested structures, so Django implemented this instead. 
It's a bit more involved than that, but this explanation suffices for 99% of cases. 

The ``setting`` attribute does exists on the ``SettingModel`` model (and thus by extension, the intervention/observation models), and links to a model that has the ``is_school`` attribute. 


As for why it does not work? That's just plain weird honestly; the actual logic for that decision barely changed in this PR, and in fact were made less strict. So that does require some more debugging I think?

Just as an aside, looking at this template it might be a good idea to refactor (most of) the show/hide logic such that it happens inside the Django Form instead of the template. Just an idea...

miggol: @Meesch Could you double check in your local Django admin that some of your Settings have the `is_school` attribute set? It might be called something like "Needs external permission" as a verbose name.

What we did to prod to hide all these forms is uncheck `is_school` from all settings. That "solution" potentially still prevents the forms from ever showing up.

Meesch: yeeeees that's it! "Needs external permission" has been turned off for all the settings, including school. When I turned it on, it works! I see that is also how it is saved on prod. Should we include that in the fixture then?

Meesch (approved):
Other than that bug that you solved for me, no notes! Good job!

PR #444, opened 2023-01-04
Fix/djang admin reviews list

miggol:
I don't know exactly when this broke, the Django changelogs mention this function signature more often than I care to investigate further.


Meesch:
Very nice, tested here and it works for me

Meesch (approved):

PR #463, opened 2023-02-17
Fix check_field_required bug

Meesch:
The Map variable would only allow a single process per input_name to be executed, leading to a bug where fields that are checked several times per change (such as 'relation', which is checked for 4 different variables and 'funding which is checked for 2 variables) would only execute the final check and abort the preceding ones, leading to fields not disappearing when they should (see screenshots below).

Screenshot 2023-02-17 at 16 10 10

Screenshot 2023-02-17 at 16 10 30

Thus, a unique process name needed to be assigned for each input_name + required_input combination, so that you can have several processes per input change, but rapidly checking and un-checking of a variable is still aborted.

This took much longer to figure out than I care to admit.

### DIFF
--- a/main/templates/base/form_styling.html
+++ b/main/templates/base/form_styling.html
@@ -96,11 +96,13 @@ function check_field_required(input_name, field_name, required_input, app_name, 
     let input = $('input[name=' + input_name + ']');
     let r_input = $('#id_' + required_input);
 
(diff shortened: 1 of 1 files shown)


miggol (approved):
Not to worry, I also spent far too long reviewing this. Well found.

Issue #325, opened 2022-04-26
Allow editing of DMP file field by secretary


miggol:
It's currently missing

miggol:
My suggestion:

* On the review info details page, the secretary can change the files associated with a certain proposal.
* Note that this is an updateview for Documents objects
* Note that each trajectory gets its own Documents object
* Note that the DMP is in fact not a part of the Documents object, but a Proposal field
** This means we can't edit the DMP on the current edit documents page
* Create a new updateview, only accessible for the secretary, which can change a proposal's DMP file field
* Add a link (for secretary users only) to the updateview for DMP on the Review detail page
* Note that the sidebar is rendered through a template tag

Issue #312, opened 2022-03-18
Add submitter and supervisor to review_detail_sidebar


miggol:
I often miss this info while scanning reviews, just a personal nice-to-have

PR #465, opened 2023-02-25
Add DMP editing and extra details to review_detail_sidebar

Meesch:
Addresses #312 and #325. Creates an UpdateView for the Data Management Plan, only accesible by the secretary. Additionally, this branch shows the applicant's name and supervisor in the review_detail_sidebar.


reviews/templates/reviews/review_detail_sidebar.html
```diff
@@ -36,6 +36,14 @@ <h3>
                 </a>
             </li>
         {% endif %}
+        <li>
+            {% trans "Aanvrager(s)" %}: {{ review.proposal.created_by}}
```
miggol, line 40: This defaults to the username, which on the server would be Solis-ID. Not very useful as many of those are just numbers. Preferably use `User.get_full_name`, optionally with email behind it (in brackets) if that's not too long.

reviews/templates/reviews/documents_list.html
```diff
@@ -26,11 +26,19 @@ <h5>
         </a>
         {% if item.comparable %}
         {% simple_compare_link item.object item.field %}
-        {% endif %}
+        {% endif %} 
     </li>
     {% endfor %}
 </ul>
 
+{% if container.dmp_edit_link %}
```
miggol, line 34: I don't think this is the correct place for this. A proposal only has one data management plan, but if a proposal has four studies the DMP edit link now shows up four times.

proposals/views/proposal_views.py
```diff
@@ -357,6 +358,19 @@ def get_next_url(self):
     def get_back_url(self):
         """Return to the consent form overview of the last Study"""
         return reverse('proposals:consent', args=(self.object.pk,))
+    
+class ProposalUpdateDataManagement(braces.GroupRequiredMixin, generic.UpdateView):
+    """
+    Allows the secretary to change the Data Management Plan on the Proposal level
+    """
+    model = Proposal
+    template_name = 'studies/study_update_attachments.html'
```
miggol, line 367: Although I appreciate the efficiency of reusing the same template, it's almost more trouble than it's worth because there's context missing now:

image

proposals/views/proposal_views.py
```diff
@@ -357,6 +358,19 @@ def get_next_url(self):
...
+    form_class = ProposalUpdateDataManagementForm
+    group_required = settings.GROUP_SECRETARY
+
+    def get_success_url(self):
+        """Continue to the URL specified in the 'next' POST parameter"""
+        return self.request.POST.get('next', '/')
```
miggol, line 373: So instead of just returning a URL to `ReviewDetailView` here, we retrieve a `next=` parameter POSTed from a hidden input in the template which generates the URL from the `HTTP_REFERER` header in the Django request object. All to end up pointing at the same `ReviewDetailView`.

9/10 I'll allow it.

locale/en/LC_MESSAGES/django.po
```diff
@@ -9,7 +9,7 @@ msgid ""
 msgstr ""
 "Project-Id-Version: \n"
 "Report-Msgid-Bugs-To: \n"
-"POT-Creation-Date: 2023-02-07 16:47+0100\n"
+"POT-Creation-Date: 2023-02-25 01:32+0100\n"
```
miggol, line 12: Still contains some fuzzy translations

miggol (changes requested):
Some small issues to work out still

reviews/templates/reviews/review_detail_sidebar.html
Meesch: Check: het ziet er nu zo uit, met het emailadres op een newline in een nested list. 

Screenshot 2023-03-07 at 16 23 58

proposals/views/proposal_views.py
Meesch: Ah classic case of: stole it from elsewhere meaning to update it, but it happened to already just work :-p I updated it for that 10/10 rating

Meesch:
@miggol I updated the branch according to your excellent feedback. Sorry for all the sloppiness, I had some serious Friday-brain.

reviews/templates/reviews/review_detail_sidebar.html
miggol: Beautiful!

proposals/views/proposal_views.py
miggol: Yeah I see it comes from the UpdateAttachments page. Honestly it was just kind of funny but I'm glad you made it right.

reviews/models.py
```diff
@@ -143,6 +143,14 @@ def get_route_display(self):
 
     def accountable_user(self):
         return self.proposal.accountable_user()
+    
+    def get_applicant_names_emails(self):
+        names = [user.get_full_name() for user in self.proposal.applicants.all()]
+        emails = [user.email for user in self.proposal.applicants.all()]
+        name_email_string = []
+        for i in range(len(names)):
+            name_email_string.append(",\n".join([names[i], emails[i]]))
```
miggol, line 152: I'm not a fan of hitting the database twice to ask for `proposal.applicants.all()`. Also, this leaves room for bugs if the ordering or length of either request is different, however unlikely. I would suggest:

```python3
names_and_emails = [
    (user.get_full_name(), user.email)
    for user in proposal.applicants.all()
]
name_email_string = []
for pair in names_and_emails:
    name_email_string.append(
        ",\n".join(pair)
    )
return name_email_string
```

Like that you only hit the database once, and the name and email are guaranteed to be from the same user.

I did enjoy that `name_email_string` is, in fact, a list, so I left that in.

miggol (approved):
Please look at my final two comments, but otherwise I approve!

reviews/models.py
Meesch: Ah you are right of coure, I implemented your solution. I also updated the name to `name_email_strings`, just to make it a tiny bit easier to understand.

Issue #408, opened 2022-10-03
Direct users stuck before login to some kind of support


miggol:
Currently, the support and secretary email addresses are only visible once logged in. That's not very helpful for people who have trouble logging in, or are confused about for whom this portal is (e.g. lost FSW students). Same with researchers that need to be added to the HUM-LDAP.

We could add support contact info to the footer or FAQ pages. I don't know if there are spam / abuse reasons not to do so. We could also add the portal support email address to the FETC WP site.

tymees:
Adding it to the footer is a nice long-term solution. However, this will only be possible with the move to DSC3.

I'm not so sure about the FAQ pages, as I don't think people use it atm. However, it doesn't hurt.

There is some spam concerns, but seeing as a lot of our emails are public and we don't really see that much spam therel I think the spamfilter works well enough. Just to be sure tho, we might want to do some JS tricks to make it a bit harder.

For the time being, I'd say just adding it to the login screen and non-logged-in index might be a good stop-gap

Issue #399, opened 2022-09-08
Change site header text


tymees:
From "FETC-GW Portal" to "Portal FETC – Geesteswetenschappen"
("Portal FEtC – Humanities" for English)

tymees:
BUMP!

Does anyone have time to do this one somewhat soon-ish?

Issue #397, opened 2022-09-06
Self assessment length limits


tymees:
So someone decided to write a whole essay for their self assessment, which is obviously not going to fit inside 500 characters.

While I have some things to say about researchers really needing to learn to write succinctly, I will admit 500 characters is too few.

We could just remove the limit altogether, but that carries a small risk we'd have to take into account. The proposal PDF generation really does not like table cells that span over more than 2 A4 pages. (Read: it crashes).  This was discovered when someone added their 1000 word entire reasearch proposal into the comments field at the end. 

I solved it by simply not rendering that field in a table, just at the root of the document. 

Another way would be to set a different limit, but I can not really think of a limit that's high enough for users but small enough to avoid the problem described above.

miggol:
> I solved it by simply not rendering that field in a table, just at the root of the document.

Does this mean that the PDF issue is avoidable? 2000 characters seems fair to me, if we can get the PDF to render.

tymees:
Yes it should be (just look at how the comments at the end of the PDF are done, that should work). I brought it up because we need to do that as well with a increase of the limit. Although, in hindsight I could've been more clear and succinct (ironically enough) about it :P

miggol:
Okay good to know! Also, it might be worth testing if the bug still exists, because I think we moved from an unmaintaned easy_pdf to directly calling a way more recent version of xhtm2pdf since. You never know!

tymees:
From e-mail: we should limit this field to 300 words or so

Meesch:
Makkelijk te fiksen als we onze word-number validator gebruiken (wordt al gebruikt voor de description op de eerste pagina)

djhcapel:
Wat is het maximum dat mogelijk is zonder pdf- of andere problemen? We willen bijv. bij voortoetsing niet meer de hele projectaanvraag hoeven door te nemen, maar we willen af kunnen gaan op de ethische reflectie van de onderzoeker.

tymees:
Onbeperkt in principe, we kunnen namelijk om alle problemen (voor zover dat ze er (nog) zijn) wel heen werken.

Echter, ik zou afraden om geen maximum te hebben; je wilt immers niet 10 kantjes moeten beoordelen ;)

PR #473, opened 2023-04-12
Low hanging fruit

tymees:
This PR addresses #408 #397 and #399 (and a small warning during messages creation).

Regarding #408:

After a discussion with Michael I decided to make ``HomeView`` a login protected page and move all non-authenticated stuff from that view into the login template. This simplifies the ``HomeView`` template, and reduces a click for users. 

Support emails are now also added to the login template, with a basic anti-spambot protection.

Regarding #397: Basically, added the right validator, removed the character length constraint and added the wordcount script to self-assessment. (With a small tweak such that the wordcount now displays _below_ the textarea for a better visual layout).

Regarding #399: pretty self-explanatory 

Regarding the warning during messages creation: there was an empty translated string in the proposal diff template. That's bad. Changing it out for a plain empty string got rid of the warning with no change in functionality


main/templates/registration/login.html
```diff
@@ -1,11 +1,70 @@
...
+            <p>
+                <strong>{% trans 'Secretaris FETC-GW' %}</strong><br/>
+                <span class="email" data-email="fetc-gw">
+                    <i>{% trans "Om hier een e-mail adres te zien moet Javascript aan staan" %}</i>
+                </span>
+            </p>
+            <p>
+                <strong>{% trans 'Technische ondersteuning' %}</strong><br/>
+                <span class="email" data-email="portalsupport.gw">
+                    <i>{% trans "Om hier een e-mail adres te zien moet Javascript aan staan" %}</i>
```
miggol, line 62: So thoughtful

miggol (approved):
Very nice, attentive as always, Ty. Couldn't find anything missing.

It's not within the scope of the issues but it would be nice to have the system messages display on the login page as well.

main/templates/registration/login.html
tymees: I wish I could take the credit; the idea is from USCKI's website lol

tymees:
> It's not within the scope of the issues but it would be nice to have the system messages display on the login page as well.

Yeah, I was thinking the same thing when doing that commit. However, as the View in question is provided by Django itself it was not as simple as copying some stuff. It's possible by using a custom version of said View, but that's not in the 'low-hanging fruit' category lol

I'll create an issue for it

PR #482, opened 2023-05-15
Nullify the minor revision attributes on copy

### LINKED ISSUE #480: Proposal.has_minor_revision and minor_revision_description are inherited by copies and revisions
Example case of this bug: (reference number redacted). This should be a quick fix in `copy_proposal()`.

The result is that comments under the "Revisie" header in the sidebar appear, even though the review is still open and its proposal has not yet ever been decided on. It might even be for a completely unrelated proposal than the original comments were made for, with the copy only serving as a base.

image

miggol:
Fixes #480

Writing the issue took more time than the fix.


Meesch:
"Sometimes, writing the issue is solving the issue"
  -- Confusius
image

Issue #485, opened 2023-05-23
Revisie/amendement


djhcapel:
Under Details, it now says 'revisie/amendement' for any revision or amendment. It would be really helpful if this would be either 'revision' or 'amendment'. Now I need to dive deeper into the application to find out what it is.

tymees:
Some technical stuff for the person implementing this:

Technically amendments and revisions aren't tracked separately. (It's the same boolean). Thus, actually figuring out what kind it is is not that straightforward. ``Proposal.type()`` already has the correct check, but that method returns more than just revision/amendment. So I'd recommend refactoring that method by extracting the checks for revision/amendment into their own methods. That way we can use those new methods in the view to solve this issue, and we don't have duplicate code.

PR #493, opened 2023-06-06
Feature/amend revise info

EdoStorm96:
#485 solved.
The details of the review page initially displayed 'amendment/revision', whereas now the display the right label for the specific proposal. There was already a check for this in the proposals.type method, which has now been made into its own method of proposal, namely, proposals.amendment_or_revision. This method gets called to solve this issue.


locale/en/LC_MESSAGES/django.po
```diff
@@ -5653,3 +5647,13 @@ msgstr "Task edited"
...
+
+#, python-format
+#~ msgid ""
+#~ "\n"
+#~ "                        Revisie/amendement van<br/> %(parent)s\n"
+#~ "                    "
+#~ msgstr ""
+#~ "\n"
+#~ "Revision or amendment of<br/> %(parent)s\n"
+#~ "   "
```
tymees, line 5659: Small thing; this PR removes a msgid, so makemessages moves it here commented out.

It does no harm, but it's better to just remove it from the file completely. Otherwise we'd get a large(r) file with unused stuff at the end.

docs/installation/steps.rst
```diff
@@ -33,6 +33,7 @@ The following packages needs to be installed:
 	+ libsasl2-dev
 	+ libssl-dev
 	+ gettext
+	+ libpoppler-private-dev
```
tymees, line 36: This should be libpoppler-cpp-dev

tymees (approved):
Approved on the condition the right dependency is added (as discussed irl) ;)

PR #494, opened 2023-06-07
Increase nr. of docs that can be added. Changed the value of 1 variable.

EdoStorm96:
Very easy and minor fix, when Ty pointed out the correct variable to change.


tymees (approved):
Simple enough ;)

Issue #499, opened 2023-06-15
New cover image


tymees:
While I'd love some cooler temperatures, I do think something less winter-y would be nice lol

Issue #472, opened 2023-03-28
Update names of the different groups of CDH


Meesch:
Change Digital Humanities Lab and Utrecht Data School to Research Software Lab and Data School, respectively.

EdoStorm96:
I don't think these are mentioned within the Ethics portal ...

Issue #454, opened 2023-02-01
UIL OTS > ILS


djhcapel:
*General information about the application:*
- To which research institute are you affiliated? one of the options should refer to Institute for Language Sciences (ILS) (replacing UiL OTS)
- How is this study funded? direct government funding: replace UiL OTS by ILS
*Participants*
- UiL OTS database > ILS database (adult and/or infant participants)
*Intervention study*
- Specify where the data collection will take place: replace UiL OTS Labs by ILS Labs (including the Babylab)
*Observational study*
- Specify where the data collection will take place: replace UiL OTS Labs by ILS Labs (including the Babylab)
*Task-based research and interviews*
- Specify where the data collection will take place: replace UiL OTS Labs by ILS Labs (including the Babylab)

miggol:
UiL-OTS still appears in two places:

* The fixtures for institutions and such

PR #503, opened 2023-06-20
Feature/small issues

EdoStorm96:
This pull request fixes multiple small issues. Namely #499 and #454. So I've changed the cover image and change all occurences of UIL OTS to ILS. UIL OTS was featured in a fixture, which I've changed, but I've also changed this on the server through the admin site, following Michael's advice. Changing the fixture and running loaddata caused no errors on my local version of the code.

I also looked at issue #472, changing the names of UDS and DH Lab, but I've found no occurences of them, so this issue can also be closed IMHO.


miggol (changes requested):
Small issue: The photo credit at the bottom of the frontpage no longer corresponds with the actual picture. Or at least so I think. It would be a huge coincidence if in fact it did.

Not that much of an issue: Considering I don't know the credit for the summer image anymore, maybe it would be an idea to grab a new one off of [the new beeldbank](https://intranet.uu.nl/nieuws/mededelingen/huisstijlplatform-gelanceerd-direct-avg-proof-fotos-downloaden)? 

But fixing the credit and keeping the old summer image is also fine to me.

tymees:
We don't have a credit for that image; it's pulled from sites.uu.nl

EdoStorm96:
Ok, I took a new cover image from the beeldbank, and credited the appropriate author. What do you think, @miggol ?

miggol (commented):
Looks good to me! I like the picture, would work well for the fall as well.

miggol (approved):
Looks good to me! I like the picture, it would work well for later in the year as well.

PR #510, opened 2023-07-05
Feature/translated consent docs

EdoStorm96:
Adresses issue 401. Ask whether the consent forms will need to be translated eventually. Added new attributes to the proposal model and added a new page in the process, before uploading the consent forms. The implementation has been discussed with Desiree.


miggol (changes requested):
Good job thinking of the pdf and diff pages as well when adding the attributes.

proposals/utils/proposal_utils.py
```diff
@@ -97,12 +97,14 @@ def available_urls(proposal):
...
-        consent_url = AvailableURL(title=_('Documenten'))
+        consent_docs_url = AvailableURL(title=_('Uploaden'), url=reverse('proposals:consent', args=(proposal.pk,)))
+        consent_url = AvailableURL(title=_('Formulieren'), children=[consent_docs_url])
+
         data_management_url = AvailableURL(title=_('Datamanagement'))
         submit_url = AvailableURL(title=_('Versturen'))
 
         if proposal.last_study() and proposal.last_study().is_completed():
-            consent_url.url = reverse('proposals:consent', args=(proposal.pk,))
+            consent_url.url = reverse('proposals:translated', args=(proposal.pk,))
```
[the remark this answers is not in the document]
EdoStorm96: I did leave line 107 be, because the consent_docs_url is a child of the consent docs_url, so these lead to purposely different urls. I did implement all you other suggestions though.

proposals/utils/proposal_utils.py
```diff
@@ -97,15 +97,21 @@ def available_urls(proposal):
         urls.append(studies_url)
 
 
-        consent_url = AvailableURL(title=_('Documenten'))
+        consent_docs_url = AvailableURL(title=_('Uploaden'), url=reverse('proposals:consent', args=(proposal.pk, )))
+        translated_docs_url = AvailableURL(title=_('Vertaling'), url=reverse('proposals:translated', args=(proposal.pk, )))
```
miggol, line 101: Some of these lines are very long. Wrapping makes them hard to read and they hide the structure of the function call. Compare:

    translated_docs_url = AvailableURL(title=_('Vertaling'), url=reverse('proposals:translated', args=(proposal.pk, )))
    consent_url = AvailableURL(title=_('Formulieren'), children=[translated_docs_url, consent_docs_url])

With the following:

    translated_docs_url = AvailableURL(
        title=_('Vertaling'),
        url=reverse(
            'proposals:translated',
            args=(proposal.pk,),
        )
    )
    consent_url = AvailableURL(
        title=_('Formulieren'),
        children=[
            translated_docs_url,
            consent_docs_url
        ]
    )

I much prefer it when similar code has a similar "shape"!

I know much of the Ethics doesn't follow these guidelines very well, but we try to write all new code in line with PEP 8. If you have a small font size or wide monitor, consider turning on an IDE warning when lines get too long.

proposals/forms.py
```diff
@@ -630,3 +630,27 @@ def clean(self):
...
+    
+    def clean(self):
+        cleaned_data = super(TranslatedConsentForms, self).clean()
+
+        if cleaned_data['translated_forms'] is None:
+            self.add_error('translated_forms', _('Dit veld is verplicht om '
+                                                'verder te gaan.'))
+            
+        elif cleaned_data['translated_forms'] == True and not cleaned_data['translated_forms_languages']:
+            self.add_error('translated_forms_languages', _('Vul in in welke talen de formulieren '
```
miggol, line 655: These lines are also too long, and not my preference in indentation. You're jumping from an indent of 12 to 48 lines! I would much prefer

    self.add_error(
        'translated_forms_languages',
        _('Vul in in welke talen de formulieren worden vertaald.'),
    )

This also saves you from breaking up the string.

proposals/forms.py
```diff
@@ -630,3 +630,27 @@ def clean(self):
                     self.add_error('comments', _(
                         'Informatiebrief voor traject {} nog niet toegevoegd.').format(
                         study.order))
+    
+
+class TranslatedConsentForms(forms.ModelForm):
```
miggol, line 635: So there's a bit of a weird interaction/bug with Django and how we build this multiple page form. If you're on a page like TranslatedConsentForms and there's an error in your input, you actually can't navigate away from it with the forward or next buttons. That's because those buttons try to save the input along the way, but that causes a Form error and Django says no. 

For this reason, SoftValidationMixin exists. We use it on proposal forms to solve this issue. It allows validation to continue and output its errors, but doesn't hard-prevent saving or form submission.

TranslatedConsentForms should use sfot validation as well, and it should mention both of its fields in `TranslatedConsentForms._soft_validation_fields`. But try to recreate the bug first so you get what I'm talking about

miggol (changes requested):
It's become time to introduce you to `/proposals/utils/validate_proposal.py`.

When you're on the final page of the proposal form (ProposalSubmit), the page checks if there are any errors on the previous pages and then mentions them to you. Depending on those errors it allows you to submit or not.

Because TranslatedConsentForms isn't yet included in `validate_proposal.py`, errors from it don't show up or prevent submission.

Take a look and see if you can add TranslatedConsentForms to the validation yourself, but please ask for help at any point because proposal validation is a bit weird.

proposals/utils/validate_proposal.py
```diff
@@ -217,6 +217,13 @@ def _build_forms(proposal: Proposal) -> OrderedDict:
                     session,
                 )
 
+    forms['translated'] = (
+        TranslatedConsentForms,
+        reverse('proposals:translated', args=[proposal.pk]),
+        _('Vertaling Consent Formulieren'),
```
miggol, line 223: I think you forgot to `makemessages` after adding this string because it isn't in `django.po`

miggol (approved):
Great! Assuming the single missed string will be fixed soon enough I'm approving this PR.

Please do not merge into develop until we've merged migrations together.

miggol:
Migrations seem to be fine, so this can be merged! (once translations fixed)

PR #513, opened 2023-07-10
Add some 'working-with-translations' notes

tymees:
First off, @miggol nice additions! No comments, aside from that I might actually start doing that instead of my own method ;)

I did think it might be nice to add some more 'working with translations' notes to the file also, so here are my suggestions.


EdoStorm96 (approved):
Useful info!

PR #521, opened 2023-07-20
Feature/embargo

EdoStorm96:
This PR introduces an embargo option for researchers to keep their applications out of the users only archive until a specified date. It also reorganizes the archive naming scheme in the code, adds a querysetManager for different Queryset needs, and reintroduced functionality for the secretary to manually change the archive status of a proposal.

Also contains a fix for issue 505


tymees (changes requested):

tymees (changes requested):
You should run `makemessages` once more, as you changed that help text.

Otherwise, looks good!

PR #528, opened 2023-08-23
Fix/applicant names diff 

EdoStorm96:
This PR fixes three small issues.

518: Solis-id's were still being displayed for other applicants on the diff page.
504: A small bug on the create pre-approved page, where not all required JS conditions had been added.
292: Now other_applicants receive an email when a proposal is submitted


reviews/utils/review_utils.py
```diff
@@ -61,6 +62,14 @@ def start_supervisor_phase(proposal):
     msg_html = render_to_string('mail/concept_creator.html', params)
     send_mail(subject, msg_plain, settings.EMAIL_FROM, [proposal.created_by.email], html_message=msg_html)
 
+    if proposal.other_applicants:
+        params['creator'] = proposal.created_by.get_full_name()
+        msg_plain = render_to_string('mail/concept_other_applicants.txt', params)
+        msg_html = render_to_string('mail/concept_other_applicants.html', params)
+        other_applicants_emails = [applicant.email for applicant in proposal.applicants.all()]
+        other_applicants_emails.remove(proposal.created_by.email)
```
tymees, line 70: You probably also want to remove the supervisor, if it's in there.

It's very common that the supervisor is also included as an applicant.

proposals/templates/proposals/proposal_diff.html
```diff
@@ -98,8 +98,22 @@ <h2 class="mt-3 mb-1">{% trans "Algemene informatie over de aanvraag" %}</h2>
                     {% if proposal.other_applicants or p_proposal.other_applicants %}
                         <tr>
                             <th>{% get_verbose_field_name "proposals" "proposal" "applicants" %}</th>
-                            <td>{{ p_proposal.applicants.all|unordered_list }}</td>
-                            <td>{{ proposal.applicants.all|unordered_list }}</td>
+                            <td>
+                                <ul style="padding:0">
```
tymees, line 102: It's nicer to use the `p-0` (padding 0) class :)

reviews/templates/mail/concept_other_applicants.html
```diff
@@ -0,0 +1,23 @@
+{% extends "mail/base.html" %}
+
+{% block content_nl %}
+Beste collega,<br />
+<br />
+Uw collega {{ creator }} heeft een concept-aanmelding ingediend bij de ethische commissie voor een onderzoek, waar u aan meewerkt.<br />
```
tymees, line 6: (This comment applies to all emails)

The UU writing styleguide prefers informal language over formal, in Dutch, for internal communication. 
(Ofwel, gebruik 'je' ipv 'u').
I see that not all mails were rewritten in the app, so imo this can be fixed separately as a part of a generic 'fix language' issue.  

For context, the passage about it in the writing styleguide:
> In interne communicatie spreken we medewerkers en studenten aan met ‘je’, ook op schrift. In externe communicatie hangt de aanspreekvorm af van de persoon en de context.

reviews/templates/mail/concept_other_applicants.html
```diff
@@ -0,0 +1,23 @@
...
+Uw concept-aanmelding wordt door de eindverantwoordelijk onderzoeker {{ supervisor }} van deze aanvraag gecontroleerd.
```
tymees, line 7: I would say 'De aanmelding wordt eerst door de eindverantwoordelijk onderzoeker {{ supervisor }} van deze aanvraag gecontroleerd.' 

Mostly because I think 'your' is a tad odd here. Technically it's true, they are named as an applicant, but this email exists because some people were complaining they weren't informed someone (wrongly) named them as an applicant.

reviews/templates/mail/concept_other_applicants.html
```diff
@@ -0,0 +1,23 @@
...
+Zodra deze de concept-aanmelding, evt. na kleine aanpassingen, heeft goedgekeurd en ter toetsing bij de FETC-GW heeft ingediend, krijgt uw collega {{ creator }} van ons bericht.<br />
+Indien de concept-aanmelding nog aanzienlijke wijzigingen uwerzijds behoeft zal de eindverantwoordelijk onderzoeker zelf contact met uw collega opnemen.<br />
```
tymees, line 9: Two points, but very subjective: 

1. I would add 'alleen' here: 'Zodra [..], krijgt alleen je collega [..]'. Just to emphasize this will be the only communication from the portal from here-on
2. The second line could be left out I think. It's factually correct, but not really that interesting for the recipient. Making shorter emails brings along the message across better imo.

reviews/templates/mail/submitted_longroute_other_applicants.html
```diff
@@ -0,0 +1,35 @@
+{% extends "mail/base.html" %}
+
+{% block content_nl %}
+{% if was_revised %}
+Beste collega,<br/>
+<br/>
+Uw collega {{ creator }} heeft een gereviseerde aanvraag ingediend voor een onderzoek, waar u aan meewerkt, bij de ethische commissie. De aanvraag is door de FETC-GW ontvangen.<br/>
```
tymees, line 7: Which application? Is what I would ask when reading this sentence. 

As the recipient might not  be aware they were added as an applicant, or if the recipient is added to a lot of applications, it's not immediately obvious what the email is talking about. 

It would be nice if the title of the proposal is given as well. 

(Same comment applies to the other mails as well)

reviews/templates/mail/submitted_longroute_other_applicants.html
```diff
@@ -0,0 +1,35 @@
...
+<br/>
+Dit houdt in dat de revisie nu door twee leden van de FETC-GW zal worden beoordeeld; een eenvoudige revisie wordt alleen door de secretaris beoordeeld. We hopen uw collega, {{ creator }}, binnenkort een beoordeling te kunnen sturen.<br/>
```
tymees, line 9: The supervisor email has a link to the PDF of the proposal. It might be nice to add it here as well

(and in the shortroute email as well)

tymees (changes requested):
Overal looks good; only the supervisor not being excluded is a blocker imo.

The email comments can be done later in a general refresh of our emails; our dear secretary loves revising texts, so she might also want to weigh in on those

EdoStorm96:
Ok, I fixed the issues you adressed, and went over some issues with the e-mail, which I deemed most relevant. I suggest we leave the more stylistic issues for another time indeed :)

tymees (approved):
One small oversight found, which does not impact anything, so LGTM

I've opened a new issue for the email language review

PR #530, opened 2023-08-28
SAML to prod

tymees:
No, this does not mean that SAML will immediately be enabled on the prod server.

It only means I can set it up and couple it to the IdP. The actual switching over to SAML will be done through a config change which will then replace the LDAP login button with the SAML login button.


miggol (changes requested):
It's a large PR so I haven't reviewed everything in detail, just ran through most of the changed pages locally. No problems found except for a missing migration.

Develop already has a `proposals/0046` migration which is more elaborate than what is necessary here,. The changes are only to `verbose_name` fields so the effects are not breaking.

I suggest the following:

1. `makemigrations` in this branch before merging
2. In develop, remove migration `proposals/0046` and pull the smaller 0046 migration from acceptation
3. In develop, run `makemigrations` again to add the additional changes back again as a new 0047 migration

But please reply to this review with alternate suggestions if there's a better way.

tymees:
I'd like to propose an alternative:

Do nothing :)

Like you said, the changes that prompt Django's need to create a migration are not actually breaking. (The missing migration won't actually do anything!)
So, it's 100% safe to deploy without said migration. 

The other way around: making the migration like you said is not 100% safe. Well, it is for production, but it will break the test server and, more importantly, anyone working from (a branch derived of) develop currently. While obviously those DB's are not holy, I don't think a missing no-op migration really warrants everyone to delete and recreate their DB. 

Yes, in a vacuum it's not ideal to have a missing migration, but honestly we've been running prod without this migration for months now. Why invest time in something that does not actually fix anything other than a 'correctness' issue (and that's coming from me), but will break dev environments?

miggol:
> Why invest time in something that does not actually fix anything other than a 'correctness' issue (and that's coming from me), but will break dev environments?

Haha. I can live with doing nothing for the sake of the development servers, but these missing migrations issues have snowballed a bit in the past. Anyone who starts a feature branch off of acc has to think to include the migration manually or we'll be facing an annoying merge with migration re-ordering down the line. Even though the migrations do nothing, Django thinks they are critical. So I do see it as more than just a correctness issue.

But the suggestion to do nothing is a valid one, so I'll change my review to approve this.

miggol (approved):

tymees:
Thanks, I'll merge this in a sec.

> So I do see it as more than just a correctness issue.

In general, I fully agree! In fact, I'd like to propose adding a new GH workflow that checks if a PR has missing migrations. This would hopefully allow us to catch missing migrations on fix/feature PR's before they even hit one of the main branches.

It's a fairly easy workflow to write, just running `manage.py makemigrations --check` (the `--check` will cause the command to return an error code on missing migrations, which will fail the workflow). I'll write a PR for that later today, if that's okay?

PR #533, opened 2023-09-04
Add CI check for missing migrations

tymees:
This PR adds a Actions workflow that tests if the code in _any_ PR is missing a migration. It also fixes the tests-workflow to also run on the acceptation branch, we forgot to add that branch to the workflow when we created it.

The missing migration workflow works by simply running ``manage.py makemigrations`` with the ``check`` argument, this causes it to return an error code if there is indeed a missing migration causing the workflow to fail.

I considered adding it to the existing tests-workflow; however, as that workflow runs for both PR's and pushes in a specific subset of branches, I decided to create a new one that only runs on PR's (with no branch restrictions) 

As a fun bonus: we actually are missing a migration in develop! I thought I had to write a fake change to test it, but apparently not.

### DIFF
--- a/.github/workflows/check-missing-migrations.yml
+++ b/.github/workflows/check-missing-migrations.yml
@@ -1,0 +1,24 @@
+name: Check for missing migrations
+
+on:
+  pull_request:
+jobs:
+  run-check:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Install OS dependencies
+        run: |
+          sudo apt-get update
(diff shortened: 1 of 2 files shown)


tymees:
I didn't create the missing migration in this PR as it's a bit out-of-scope. However, I'm fine with adding it anyway if y'all want it.

tymees:
Bump @miggol @EdoStorm96

miggol (approved):
I like it, non-blocking and a separate workflow keeps things more oranised either way. Makes it easier to copy to my Django projects as well!

Issue #357, opened 2022-06-28
Diff shows answers that have become irrelevant


tymees:
Consider the following template code:
```
{% if study.legally_incapable or p_study.legally_incapable %}
    <tr>
        <th>{% get_verbose_field_name "studies" "study" "legally_incapable_details" %}</th>
        <td>{{ p_study.legally_incapable_details }}</td>
        <td>{{ study.legally_incapable_details }}</td>
    </tr>
{% endif %}
```

Then, imagine a study that has legally incapable participants in version 1, but not in version 2. In this case, one would expect that something like this will be displayed:

| Do you have legally incapable participants? | yes | no |
|---|---|---|
| Explain  | Well because I want to  |   |

However, this doesn't happen. Instead, you'll get the same answer twice:

| Do you have legally incapable participants? | yes | no |
|---|---|---|
| Explain  | Well because I want to  | Well because I want to  |

This is because we don't actually clear the 'details' field when someone changes their anser to 'no'. (This is complicated even more with multiselects with conditional questions).

Well, easy you think! Just do something like this:
```
{% if study.legally_incapable or p_study.legally_incapable %}
    <tr>
        <th>{% get_verbose_field_name "studies" "study" "legally_incapable_details" %}</th>
        <td>
                {% if p_study.legally_incapable %}
                    {{ p_study.legally_incapable_details }}
                {% endif %}
        </td>
        <td>
                {% if study.legally_incapable %}
                    {{ study.legally_incapable_details }}
                {% endif %}
        </td>
    </tr>
{% endif %}
```

And that works fine. However, it has a few drawbacks:
1. You're repeating the same if statement twice. In this case, it's not a real problem because the if is fast. However, this patterns happens a lot, also in places where the if statement is actually somewhat expensive. 
2. Fixing this everywhere is _a lot_ of work and is very error prone. (Exemplified by me accidentally typing study.has_trians vs study.has_traits)
3. It makes the code less readable. In other places, this is just a nightmare readility wise:

```
        {% if study.has_traits or p_study.has_traits %}
            <tr>
                <th>{% get_verbose_field_name "studies" "study" "traits" %}</th>
                <td>{{ p_study.traits.all|unordered_list }}</td>
                <td>{{ study.traits.all|unordered_list }}</td>
            </tr>
            {% if study.traits.all|needs_details or p_study.traits.all|needs_details %}
                <tr>
[...]

PR #537, opened 2023-09-21
Fix/generate diff pdf data

EdoStorm96:
Well here it is. The pdf_diff_logic page is the main event. This is where I mainly added documentation, so have a look at that. Both the diff and pdf are currently working.

There are some minor bugs. 
- Mainly the JS on the diff page probably needs some changes
- I did not implement pagebreaks for the PDF just yet. Do we want that?
- There's no fix for comparing different version observation/intervention objects as of now.


proposals/utils/pdf_diff_logic.py
```diff
@@ -0,0 +1,1132 @@
...
+        'approval_institution',
+        'approval_document',
+        'registrations',
+        'registrations_details'        
+    ]
+
+    def get_row_fields(self, obj):
+        rows = copy(self.row_fields)
+
+        if obj.version == 1:
```
tymees, line 607: It might make more sense to make a `v1` and a `v2` specific class (with a common base class or mixin for the stuff they share) and do the version check in `create_context_pdf`.

This will allow you to remove the 'remove these fields in case of version x' code and just not have those fields in `row_fields`.
In the future we could then make a new class for a possible `v3`, instead of adding an extra `if` here (making the class less clear).

proposals/utils/pdf_diff_logic.py
```diff
@@ -0,0 +1,1132 @@
...
+            
+        context['embargo'] = EmbargoSection(model)
+    
+    return context
+
+def create_context_diff(context, models):
+    '''A function to create the context for the diff page.'''
+
+    p_proposal = models[0]
+    proposal = models[1]
```
tymees, line 1047: Why use a tuple to pass the two proposals? I think having them as separate arguments to the functions is a bit clearer

proposals/views/proposal_views.py
```diff
@@ -445,7 +446,7 @@ def get_next_url(self):
 
     def get_back_url(self):
         """Return to the data management view"""
-        return reverse('proposals:data_management', args=(self.object.pk,))
+        return reverse('proposals:data_management', aPdfrgs=(self.object.pk,))
```
tymees, line 449: I don't think you intended to make this change :)

tymees (changes requested):
This is a lot more readable (and maintainable) than the templates of doom we had before! Nicely done :)

However, the class inheritance structure is a bit too intertwined in non-obvious ways. I know you're new to the whole class-inheritance thing, and as it is it is already really good. I've added some comments to help you separate them a bit, to make it that bit clearer. 

Also, I've added a lot of readability comments for the templates, but I have one more I could not fit in a comment as it's related to a file that wasn't changed. The tables on the diff page could use some padding inside the cells, so adding the following to `main/static/main/style.css` might be an idea:
```css
.proposals-diff th, .proposals-diff td {
    padding: 10px;
}
```

proposals/utils/pdf_diff_logic.py
```diff
@@ -0,0 +1,1132 @@
...
+                {
+                    'study_title': super().get_study_title(self.object.study)
+                }
+            )
+        return super().render(context)
+    
+class ObservationSection(InterventionSection):
+    '''This class receives an observation object
+    It inherits from the InterventionSection, as it can repurpose the same overridden
+    render method.'''
```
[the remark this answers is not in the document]
EdoStorm96: Ok, this makes a lot of sense. I will check out how I can reorganise it more sensibly.

proposals/utils/pdf_diff_logic.py
```diff
@@ -0,0 +1,1132 @@
...
+NOTE: As of now, the diff has no workaround for when observations or interventions are of a
+different version, in the different proposals.'''
+
+class PDFSection:
+    '''This is the main class, from which most classes are constructed.
+    It can take one object or two objects in a tuple, and produce html tables accordingly.
+    The most important methods are render and make_rows. Make rows makes use of get_row_field,
+    which is mostly overwritten for specific sections.
+    If this class is used in the diff, object should be the parent object and object_2 the 
+    revision/amendment.'''
```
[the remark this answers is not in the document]
miggol: Tacking on to this, as I have an alternative solution that might be interesting.

What we're seeing here in `PDFSection` and also in the `ObservationSection` is _type keying_, which tends to happen when a class or function fulfills two different functions resulting in a kind of split personality. In this case the class keeps checking if it has an extra object to determine what template it needs, what rows it desires, what it wants for dinner, etc. It is usually best to split such classes up, as Ty says.

In my opinion you have a perfectly good `PDFSection` for the PDF case. And simply put, a diff is just two of those. Why not have a standalone `DiffSectionV2` class, that takes two PDF sections as arguments. Specifically I mean that `DiffSectionV2` takes two instantiated section objects[1] and not two proposal or study (or whatever) objects. By standalone I mean that it would not inherit from `PDFSection`, or anything necessarily.

I think this design would be more resilient than passing two objects to a `PDFSection` spinoff class and then making it do everything twice. Let the two sections do their own thing as if they're just making a PDF and then combine them later. You already have most of this code hidden under diff-checking if statements in `PDFSection` waiting to be cut and pasted.

I must say that I find the multiple inheritance solution (e.g. `StudySectionDiff`) quite creative and cool. But sadly that's how you get MRO problems.

[1] Or maybe None to indicate a missing section. There are probably multiple ways to sort out the missing section problem here, do what feels right.

proposals/utils/pdf_diff_logic.py
```diff
@@ -0,0 +1,1132 @@
...
+
+        if extra_documents_objects:
+            context['extra_documents'] = extra_documents_objects
+
+        if model.dmp_file:
+            context['dmp_file'] = DMPFileSection(model)
+            
+        context['embargo'] = EmbargoSection(model)
+    
+    return context
```
miggol, line 1041: In my suggestion to remove even more logic from the templates, this function would just return a list of sections. The logic to check if sections exist or are required is already here anyway.

miggol (commented):
I don't have much to add on top of Ty's comprehensive review. I also ran into the missing METC issue. Otherwise my main points would be:

* Bring back page breaks, possibly as their own section
* Remove more logic from the templates
* Awesome work so far
* Great comments and docstrings. They help a lot

proposals/utils/pdf_diff_logic.py
EdoStorm96: I started working on this, but did not really find it that much more elegant ... If you really want, I will implement this, but creating a 2 extra baseclasses, 2 more version classes, plus more logic in the create context functions, to me, makes it more complicated.

EdoStorm96:
So I've overhauled the whole thing, as we discussed. I've, of course, incorporated the more minor fixes mentioned above. Thanks Ty! But I've also now made it so that a DiffSection does indeed just take two sections of a proposal and p_proposal, as suggested by Michael. This made the whole thing a lot more streamlined and removed a lot of basically useless classes. I suddenly saw the light ;) I also created a new way for dealing with the former `study_title` generation (I've now renamed them `sub_title`)

I think all in all this has improved a lot, but there are some new methods which are a bit iffy ... I've again written some pretty extensive documentation, so have a look at that. With some brief testing, this seems to work well.

I've also incorporated Michael's hotfix of this week, so that certain long fields go in a `<p>` element. This is implemented fairly sloppily though ... I am looking forward to  new suggestions and recommendations from you both, as I'm sure there are more improvements :)

proposals/utils/pdf_diff_logic.py
tymees: Yeah fair; at this point it's probably a bit unnecessary. It was a bit of a 'that might be neat in the future' suggestion, but it's fair to say we can defer this to if and when that future arrives

proposals/templates/proposals/proposal_diff.html
```diff
@@ -14,9 +14,13 @@
 {% block html_head %}
     <script src="{% static 'main/js/htmldiff.js' %}"></script>
     <script>
+        // The new implementation causes some issues with the js, 
+        // - The loading icon does not go away sometimes.
+        // - The colors appear for the first instance of a missing section warning,
+        //   which is kindoff ugly.
```
tymees, line 20: These comments can now be removed I think?

proposals/templates/proposals/table_and_text_with_header.html
```diff
@@ -0,0 +1,20 @@
+{% load i18n %}
+
+{% if page_break %}
+    <h2 style="page-break-before: always"> {% trans section_title %} </h2>
+{% elif section_title %}
+    <h2> {% trans section_title %} </h2>
+{% endif %}
```
tymees, line 7: Correct me if I'm wrong, but isn't this implying a page break must have a section title?
Would it not be better to do something like this:
```django
{% if page_break %}
    <div style="page-break-before: always"></div>
{% endif %}
{% if section_title %}
    <h2> {% trans section_title %} </h2>
{% endif %}
```

proposals/templatetags/proposal_filters.py
```diff
@@ -33,7 +33,6 @@ def necessity_required(study):
     age_groups = study.age_groups.values_list('id', flat=True)
     return check_necessity_required(study.proposal, age_groups, study.legally_incapable)
 
-
 @register.simple_tag
```
tymees, line 36: Why are we removing this new-line?

There should be two between functions

proposals/utils/pdf_diff_logic.py
```diff
@@ -0,0 +1,1134 @@
...
+            output = _('Niet aangeleverd')
+
+        return output
+    
+    def yes_no_doubt(self, value):
+        from main.models import YES_NO_DOUBT
+        d = dict(YES_NO_DOUBT)
+        return d[value]
+    
+class StudyTitleClass:
```
tymees, line 275: I think this should be called `SubTitle` or something; It's only used in `get_sub_title`, so it would be nice if the name matches up

tymees (changes requested):
Yes, this is much more logically structured!

It's really coming together now. Functionally and structurally you're pretty much there. 
The comments I added are (almost exclusively) about small cleanup and enhancements.

I do have 1 more general comment:

My IDE gives me 256 style-related warnings...
Mostly, you really try to get as much as possible on a single line, which is often against best practice. But there are some missing new-lines between classes/functions here and there as well.

So, if you are ever up for some light reading, I recommend you check out 'PEP8'. (The general styleguide for Python).

In addition, while we haven't done it yet, we are planning to do auto-format with a package called 'black'. I would recommend you run it on the `pdf_diff_logic.py` file:
```bash
pip install black
black proposals/utils/pdf_diff_logic.py 
```
(Yes, sorry for hammering so much on style related issues. I have trouble reading sometimes...)

EdoStorm96:
Thanks for your further suggestions, Ty. I implemented them all and reformatted using black. I will look into PEP 8 :) I think the code is looking good!

tymees (approved):
Yes, looks great!

miggol (changes requested):
Awesome! It all seems to be fixed. Just one last thing...

image

Do you think you could fix #357 while you're at it? :)

I kind of forgot this was the original issue that first started this whole journey.

tymees (changes requested):
So sorry, but I have an additional request...

Last week, Michael and I decided that we should ditch tables altogether. It caused problems we will never fully resolve. Thus, [I made a very quick and dirty solution](https://github.com/DH-IT-Portal-Development/ethics/pull/555) that changes it to a simpler 'question followed by answer' layout.

Could you change your PR to do the same? You should have a more easy time of it, as you only have to change 1 small template ;) 

As a starting point, I'd suggest copying the CSS changes I made, and rewrite your template to use div's instead of tables. 
Some notes:
- The `.row` class is not needed, I forgot to remove it lol. 
- You probably want to rename `.table`, `.th` and `.td` to something more descriptive. I was lazy, so I didn't. (`.section`, .`question` and `.answer` come to mind)
- The font related changes are not really related to the layout, they fix an issue with non-roman letters. It would be nice to copy that fix over :)
- Also, feel free to play around with styling; I made something fast, not nice :)

proposals/utils/pdf_diff_logic.py
```diff
@@ -0,0 +1,1271 @@
...
+    return extra_documents
+
+
+def create_context_pdf(context, model):
+    """A function to create the context for the PDF, which gets called in the ProposalAsPdf view."""
+
+    sections = []
+
+    sections.append(GeneralSection(model))
+    sections.append(WMOSection(model.wmo))
```
tymees, line 1068: (Same holds true for the diff variant)

This fails on pre-approved studies, which do not get a WMO object.

proposals/templatetags/diff_tags.py
```diff
@@ -10,8 +10,10 @@ def zip_equalize_lists(a, b):
     A zip implementation which will not stop when reaching the end of the
     smallest list, but will append None's to the smaller list to fill the gap
     """
-    a = list(a)
-    b = list(b)
+
+    a = [None] if a is None else list(a)
+    b = [None] if b is None else list(b)
+
```
tymees, line 16: This goes horribly wrong if `a == None` and `b == []`. You'll get a return of `[[None, None]]`, which of course should be `[]`.

As I think you tried to solve the issue of getting a plain `None` instead of an iterable, I think this should be changed to
```python

    a = [] if a is None else list(a)
    b = [] if b is None else list(b)
```

proposals/utils/pdf_diff_logic.py
```diff
@@ -0,0 +1,1271 @@
...
+            rows.remove("traits_details")
+        elif not needs_details(obj.traits.all()):
+            rows.remove("traits_details")
+        if not necessity_required(obj):
+            rows.remove("necessity")
+            rows.remove("necessity_reason")
+        if not needs_details(obj.recruitment.all()):
+            rows.remove("recruitment_details")
+        if not obj.compensation or obj.compensation.needs_details:
+            rows.remove("compensation_details")
```
tymees, line 591: I'm getting this field even tough I shouldn't. It seems that you forgot to invert the second part:

`if not obj.compensation or not obj.compensation.needs_details`

tymees (changes requested):
I found 3 minor bugs during this testing round; should be simple fixes.

However, I'd also ask @miggol to take a second look, as this is a pretty important thing to get right

miggol (changes requested):
I mainly focused on the most recent changes and didn't go through and try to break everything this time. 

The irrelevant fields issue needs some more work, and I would appreciate it if `StudySection` and `DMPFileSection` both got the page break mixin.

Otherwise no more issues besides what Ty mentioned!

miggol (changes requested):
Crazy how much code the new `make_diff_rows()` approach lets you remove!

The missing wmo issue pointed out by Ty earlier still needs more fixing though, and I would still appreciate some more page breaks.

I didn't find any new issues though. Looks almost done!

proposals/utils/pdf_diff_logic.py
```diff
@@ -1065,7 +992,9 @@ def create_context_pdf(context, model):
     sections = []
 
     sections.append(GeneralSection(model))
-    sections.append(WMOSection(model.wmo))
+
+    if model.wmo:
```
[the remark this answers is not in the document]
EdoStorm96: I am sorry. That was a bit sloppy of me.

EdoStorm96:
Ok, So I found that I hadn't really supported pre_approved proposal at all ... These are now properly incorporated. Plus I found and fixed a bug with generating link (Not starting the `href` with `http://` ... Should be good now. I am scared to merge though ...

tymees:
I've pushed two fixes; one missing italics font, and another to use https over http. Didn't want to burden you with fixing something I already had fixed locally.

I'm fine with this PR as is now; I've tested it with an utterly broken proposal, and it didn't crash. (It also didn't generate proper output, but it did it improperly in the proper way for how the proposal was broken :P)

As for merging: yeah ask Michael, he also wrote the canonical PDF so he probably knows best how to merge it

PR #541, opened 2023-09-26
Update master from acceptation

miggol:
Draft for now, but plenty of features that would be nice to have in prod such as:

* Translated consent docs
* Review overview table
* Embargo question
* Canonical PDFs

If I understand the situation correctly this PR will not enable SAML by default, that's a deployment change. However, the hidden SAML login URL for SSO is already available on prod and can be used to log in by anyone (also other faculties than GW) if they know the URL.


miggol:
Merge and deployment is planned for the beginning of next week

PR #544, opened 2023-10-03
Apply shrink style to tables too large for one page

miggol:
Fixes LayoutError in xhtml2pdf when a table is larger than a page. The `{-pdf-keep-in-frame-mode: shrink;}` CSS-style is applied to tables prevent this from happening in the future. Additionally the most likely offender, the self-assessment field, has been broken out of the table and given its own free-text field.

Sorry about messing with the indentation, recommended diff viewing with `git diff -w` or the appropriate checkbox on Github.

### DIFF
--- a/proposals/templates/proposals/proposal_pdf.html
+++ b/proposals/templates/proposals/proposal_pdf.html
@@ -7,525 +7,527 @@
 {% load uil_filters %}
 
 {% block extra_style %}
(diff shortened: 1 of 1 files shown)


tymees (approved):
I hate it, but as we need something that works yesterday, it's good enough

Let's think of a lasting fix during the meetintlg

PR #551, opened 2023-10-11
Suggestion/privileged faculty flag

miggol:
Suggestion to allow certain faculties based on an "is_privileged" flag on the model rather than hard-coding the Humanities faculty.


miggol:
Closing as obsolete

Issue #540, opened 2023-09-26
New application (that has been approved by another ethics committee) 


djhcapel:
Something goes wrong on this page. If you answer 'yes' to the first question (Do you have formal approval from an ethics committee, other than this FEtC-H committee?), the following questions (that should not appear in the first place) disappear; they remain, however, if you tick 'no'. 

- What is your course of study? |  
as a part of my thesis         as part of an internship         other
- Students (who do person-oriented research as part of their program) do, in principle, not need to submit a proposal with the ethical committee. Discuss with your supervisor whether you actually need to submit a proposal. If you do not not, you can terminate your proposal now. If you do, please explain what the reason is:

This first question, Do you have formal approval from an ethics committee, other than this FEtC-H committee?, seems to be a strange question anyway, or is it meant a control question?

djhcapel:
I just noticed that even if you click 'yes' at the first question, still student-related questions are being asked. This should not be the case.

EdoStorm96:
So basically, these fields should only appear when you've said that you are a student:

Wat is je studierichting?

In welke context doe je dit onderzoek?

Studenten (die mensgebonden onderzoek uitvoeren binnen hun studieprogramma) hoeven in principe geen aanvraag in te dienen bij de FETC-GW. Bespreek met je begeleider of je daadwerkelijk een aanvraag moet indienen. Als dat niet hoeft kun je nu je aanvraag afbreken. Als dat wel moet, geef dan hier aan wat de reden is:

But now they appear at all times. Should be an easy fix.

PR #552, opened 2023-10-12
Fix/improve workload overview

EdoStorm96:
I've made improvements to the workload overview of committee members, according to Desiree's wishes.

The queryset for get all open decisions had to be updated, because it was letting in the wrong things.
Also, to get all reviewers of a chamber, the code now finds all decisions in a chamber first and then looks at all associated reviewers. The old way also showed Michael for instance ...

The period for the second table can now also be customized, using two datepickers. This was a request of Desiree.

I've also incorporated a fix for issue #540, because it was a minor fix.


reviews/views.py
```diff
@@ -73,101 +79,136 @@ class DecisionOpenView(BaseDecisionListView):
...
+        initial = super().get_initial()
+
+        if "start_date" in self.request.GET and "end_date" in self.request.GET:
+            initial["start_date"] = self.request.GET.get("start_date")
+            initial["end_date"] = self.request.GET.get("end_date")
+        else:
+            initial["start_date"] = self.one_year_ago.strftime("%Y-%m-%d")
+            initial["end_date"] = self.today.strftime("%Y-%m-%d")
+
+        return initial
```
tymees, line 111: The way you're using the form here doesn't really make sense. It's basically a way to only display the form, which would probably be better done with just plain HTML at that point. 

It works... But you're leaving a lot functionality on the table, most importantly validation.

I would suggest actually using the form-processing; This is done by making the form use 'post', and then overriding the 'post()' method of this View to retrieve the validated and cleaned data. For example:
```python

    def __init__(self, **kwargs):
        super().__init__(**kwargs)
        self.today = date.today()
        self.start_date = self.today - timedelta(days=365)
        self.end_date = self.today

    def post(self, request, *args, **kwargs):
        form = self.get_form()
        if form.is_valid():
            self.start_date = form.cleaned_data['start_date']
            self.end_date = form.cleaned_data['end_date']
        else:
            return self.form_invalid(form)

        return self.get(request, *args, **kwargs)

    def get_initial(self):
        initial = super().get_initial()

        # You don't actually have to cast these to strings btw, DateField can handle date(time)s
        initial["start_date"] = self.start_date
        initial["end_date"] = self.end_date

        return initial
```

Then, use `self.start_date` and `self.end_date` in your querysets

tymees (changes requested):
Logic looks good, the new queries do seem to make a lot more sense now.

The way you use the form could be better, luckily that's a fairly simple fix. 

Lastly, it seems like you forgot to provide translations?

EdoStorm96:
Thanks, Ty. That does look a lot smoother. I still struggle with comprehending forms ;p

One thing I ran into, is that on my system, if I don't add the string formatting of the initial dates, when I load the page in Dutch, the initial value of the dates defaults to (literally) "mm / dd / yyyy" instead of "10 / 23 / 2023" for example. It works fine when I load the page in English without the `strftime` but in Dutch it causes this bug. When I then submit the form, the dates will show correctly in the Dutch version as well. Do you think this is just a bug of my system? For now I've kept the string casting in the `get_initial` method.

tymees (approved):
It's a bug in DSC

I can fix it there quite easily, but it will be a while before it can be used here. So, for now, your workaround is fine I guess?

I would like the comment above it to indicate it's a workaround for a bug. It's an approve for me, granted you do change it.

PR #555, opened 2023-10-17
'Emergency' pdf fixes

tymees:
This is a very dirty rewrite of the PDF templates to remove tables altogether. It also adds a font that can actually display non-latin characters. 

It's done with a very simple search/replace, thus the actual layout is kinda bonkers for what it looks like. #537 should probably implement a better non-table based layout. 

I'm totally not happy with this PR, but IMO, it's better to have a working PDF now than wait weeks till the proper rewrite is finished 

The PDF now looks like this:

image
(Name blurred because to protect my privacy)


miggol (approved):
I took the liberty of pushing some CSS and pagebreak tweaks directly to this branch. Without tables it was a lot harder to tell when sections started and ended.

Great that you could get this out of the door so quickly. We'll leave the rest of the PDF templates for the big PDF update.

tymees:
Good idea! Thanks for the tweaks;

I'm going to run this change past Desiree first before merging, as it's kinda a major change

tymees:
Desiree gave the thumbs up, merging for a monday deploy

PR #561, opened 2023-10-26
remove supervisors from overview

EdoStorm96:
Implemented tiny change to remove supervisors from the second workload overview.


tymees (approved):
Let's try to get this deployed into prod today :)

PR #566, opened 2023-11-01
Fix/supervisor clarity

EdoStorm96:
Fix for issue #531. Users can no longer select themselves as supervisor and the wording has changed in a few instances from 'researcher with final responsibility' to 'promotor/supervisor' as this could add more clarity.


tymees (changes requested):
Functionally it's approved, but there is one textual issue I'd like to see fixed before actual approval :)

PR #567, opened 2023-11-01
style: Indent templates with djhtml

miggol:
Run djhtml on all top-level directories tracked by git, except docs. No options were given to djhtml, resulting in the following defaults: four space indents and all CSS and JS are included.

Please give this a once-over to try and spot any weirdness and (parts of) files that should be excluded.

I'm leaving this as a draft until translations are fixed. Most template blocktranslations that did not use `trimmed` have to be adjusted.

### DIFF
--- a/main/templates/base/menu.html
+++ b/main/templates/base/menu.html
@@ -7,150 +7,150 @@
             <a class="pure-menu-link" href="{% url 'main:home' %}">{% trans "Startpagina" %}</a>
         </li>
         {% if user.is_authenticated %}
-        <li class="pure-menu-item pure-menu-has-children pure-menu-allow-hover">
-            <a href="#" class="pure-menu-link">{% trans "Mijn aanvragen" %}</a>
-            <ul class="pure-menu-children">
-                <li class="pure-menu-item">
-                    <a class="pure-menu-link" href="{% url 'proposals:start' %}">{% trans "Nieuwe aanvraag starten" %}</a>
+            <li class="pure-menu-item pure-menu-has-children pure-menu-allow-hover">
+                <a href="#" class="pure-menu-link">{% trans "Mijn aanvragen" %}</a>
+                <ul class="pure-menu-children">
+                    <li class="pure-menu-item">
+                        <a class="pure-menu-link" href="{% url 'proposals:start' %}">{% trans "Nieuwe aanvraag starten" %}</a>
+                    </li>
+                    <li class="pure-menu-item">
+                        <a class="pure-menu-link" href="{% url 'proposals:copy' %}">{% trans "Nieuwe studie starten op basis van een kopie van een oude studie" %}</a>
+                    </li>
+                    <li class="pure-menu-item">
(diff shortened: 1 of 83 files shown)


main/templates/base/menu.html
```diff
@@ -7,150 +7,150 @@
...
-                </li>
-                <li class="pure-menu-item">
-                    <a class="pure-menu-link" href="{% url 'feedback:create' %}">{% trans "Feedback" %}</a>
-                </li>
-            </ul>
-        </li>
-        <li class="pure-menu-item">
-            <a class="pure-menu-link" href="{% url 'logout' %}"><em>{% trans "Uitloggen" %}</em></a>
-        </li>
+            <li class="pure-menu-item pure-menu-has-children pure-menu-allow-hover">
```
tymees, line 10: Not a comment about styling, but shouldn't this file be gone-zo for years now?

main/templates/base/sidebar.html
```diff
@@ -5,23 +5,23 @@
 <link rel="stylesheet" type="text/css" href="{% static 'main/progress.css' %}">
```
tymees, line 5: Another file that should be removed

proposals/templates/proposals/pdf/observation_v2.html
```diff
@@ -34,50 +34,50 @@
...
-            <th>{% get_verbose_field_name "observations" "observation" "has_advanced_consent_details" %}</th><td>{{ observation.has_advanced_consent_details }}</td>
+            <th>{% get_verbose_field_name "observations" "observation" "is_nonpublic_space_details" %}</th><td>{{ observation.is_nonpublic_space_details }}</td>
         </tr>
+        <tr>
+            <th>{% get_verbose_field_name "observations" "observation" "has_advanced_consent" %}</th><td>{{ observation.has_advanced_consent|yesno:_("ja,nee") }}</td>
+        </tr>
+        {% if not observation.has_advanced_consent %}
+            <tr>
+                <th>{% get_verbose_field_name "observations" "observation" "has_advanced_consent_details" %}</th><td>{{ observation.has_advanced_consent_details }}</td>
+            </tr>
```
tymees, line 62: Hm, if I were an HTML formatter, I'd think this line is way too long, and would at the very least want to move the two elements to separate lines.

Ideally, I'd even change it to something like this:
```html
<tr>
    <th>
        x
    </th>
    <td>
        y
    </td>
</tr>
```

I guess djhtml is primarily an indenter :P
That's fine, I'm just used to Prettier also doing stuff like that for me.

tymees (commented):
I think the format is fine;

As DJHtml is primarily an indenter, I don't really think that its output is that great, but it's better than nothing. The flip side being that I cannot find a single weird thing caused by DJHtml. 

At some point we also looked at djlint, which also does a lot of formatting as well. I cannot recall why we went with djhtml over djlint tho?

miggol:
Closing as obsolete

PR #568, opened 2023-11-01
style: reformat templates with djlint

tymees:
So, basically a competitor for #567 using DJLint instead. Run without any special config, except excluding docs. (This means also untracked files btw...)

To run:
```bash
$ pip install djlint
$ djlint . --reformat
```

I haven't looked at it before making the PR, making a very pure format. As with #567, non-trimmed blocktrans are broken


main/templates/base/navigation.html
```diff
@@ -11,61 +9,45 @@
...
-                        {% if item.url %}
-                            <a class="js-submit-redirect" href="{{ item.url }}">
-                        {% endif %}
-                        <div class="navigation-bubble">
-                            {% counter main_counter value %}
-                        </div>
-                        {% if item.url %}
-                            </a>
-                        {% endif %}
+                    <div class="navigation-content {% if not item.url %}inactive{% endif %} {% if main_counter == active %}active{% endif %} ">
```
miggol, line 12: The original may have been weird but this definitely isn't better

main/templates/base/site_title.html
```diff
@@ -1,2 +1,2 @@
 {% load i18n %}
-{% trans "FETC-GW" %}
\ No newline at end of file
+{% trans "FETC-GW" %}
```
miggol, line 2: Yay! Newlines at end of file

proposals/templates/proposals/proposal_copy.html
```diff
@@ -1,8 +1,6 @@
 {% extends "base/base.html" %}
-
```
miggol, line 2: I think we should set `max-blank-lines=3` or something reasonable. I still want to be able to add empty lines for readability

https://djlint.com/docs/configuration/#max-blank-lines

proposals/templates/proposals/study_consent.html
```diff
@@ -1,15 +1,12 @@
...
 {% load uil_filters %}
-
 {% block header_title %}
-  {% trans "Informed consent formulieren" %} - {{ block.super }}
+    {% trans "Informed consent formulieren" %} - {{ block.super }}
 {% endblock %}
-
 {% block html_head %}
-  <script>
+    <script>
```
miggol, line 9: Could you update the branch with `format_js=true` in the config? I'm curious: might as well see what djlint is capable of

main/templates/base/site_title.html
tymees: Poe's law is at full effect here....

proposals/templates/proposals/study_consent.html
tymees: Done, also with `format_css=true`

proposals/templates/proposals/proposal_copy.html
tymees: Done; I've also added some config to require a blank line after the extend and load 'blocks'

proposals/templates/proposals/study_consent.html
```diff
@@ -5,150 +5,170 @@
...
+                if not form.instance.study.needs_additional_external_forms %
+            }
+            hide_school_forms("{{form.prefix}}");
+            {
+                % endif %
+            } {
+                % endif %
+            } {
+                % endfor %
+            }
```
miggol, line 59: I think we all saw this coming to a certain degree. Doesn't make the end result any more bearable.

tymees: Ehhhhh yeah that's not great...

Simple fix would be to _not_ format JS... But honestly, maaaybe we should just not do this horrible JS generation. It's kinda an anti-pattern better solved with some data attribute parsing

miggol: Let's keep looking for other places that are borked. Once the new attachments system comes in this code will be gone anyway.

tymees:
Superseded by the actual attempt in #600

Issue #512, opened 2023-07-07
Date sent for revision on 'In revision' page?


djhcapel:
Could we have the date that an application was sent for revision on the _Aanvragen in revisie_ page? I now have to click on an application to see what date it was sent for revision.

Issue #483, opened 2023-05-16
DMP and application editing possibilities secretary


djhcapel:
I would like to be able to update a DMP that has been sent by email (in case of really minor changes we do not request applicants to submit an entirely new version).
I would also like to be able to update at least the intended start date in the application form: quite often, this date needs to be adapted at the last moment.

Can we have an 'edit documents' button for those documents?

djhcapel:
This has been solved for the DMP, but not yet for the intended start date in the application form.

PR #570, opened 2023-11-07
Revision date in list & edit intended start date

EdoStorm96:
Fixes for issues #512 and #483. Just some niceties for Desiree. I'd say these work fine :)


proposals/templates/proposals/proposal_update_date_start.html
```diff
@@ -0,0 +1,36 @@
...
+    <div class="uu-inner-container">
+        <div class="col-12">
+            <h2>
+                {% trans "Startdatum aanpassen" %}
+            </h2>
+            <p>
+                {% blocktrans trimmed with title=proposal.title ref_number=proposal.reference_number %}
+                Op deze pagina kan de startdatum worden aangepast van de aanvraag {{ title }}
+                (referentienummer <em>{{ ref_number }}</em>). <b>Let op!</b> Als de review al is afgerond,
+                wordt de nieuwe startdatum niet automatisch weergegeven in de PDF. Mocht u de PDF
```
tymees, line 20: 'U' should be 'je'

proposals/templates/proposals/proposal_update_date_start.html
```diff
@@ -0,0 +1,36 @@
...
+                Op deze pagina kan de startdatum worden aangepast van de aanvraag {{ title }}
+                (referentienummer <em>{{ ref_number }}</em>). <b>Let op!</b> Als de review al is afgerond,
+                wordt de nieuwe startdatum niet automatisch weergegeven in de PDF. Mocht u de PDF 
+                opnieuw willen genereren, neem hierover dan contact op met 
+                {% endblocktrans %}
+                <a href="mailto:portalsupport.gw@uu.nl">portalsupport.gw@uu.nl</a>.
+            </p>
+            <p>
+                {% trans "Vul de datum in als 'DD-MM-YYYY'." %}
+            </p>
```
tymees, line 27: This is incorrect.

It might work, it will probably work when the locale is set to Dutch, but it might not always.

YYYY-MM-DD will always work

Or... Just use the custom year widget?

reviews/templates/reviews/vue_templates/review_list.html
```diff
@@ -118,6 +118,15 @@ <h4>
                 {{ review.get_stage_display }}
                 {% endverbatim %}
             </div>
+            {# For some reason context.review.REVISION and other class variables of continuation are None?#}
+            {# Only context.review.GO returns 0 for me, as it should #}
+            {# Which is why I've hardcoded the 1 here ... #}
```
tymees, line 123: I can answer this one for you. This is a Vue template (and a first-generation one, which is a bit crap compared to what we're migrating to).

Thus, the context used here is not the same as what you'd expect. It's a manually created dict in `reviews/api/views.py`. 
`GO` is in the dict; `REVISION` isn't.  If you add it to the dict, you can reference it here.

reviews/templatetags/documents_list.py
```diff
@@ -222,6 +223,10 @@ def documents_list(review, user):
 
     pdf_container.items.append(proposal_pdf)
 
+    if is_secretary(user):
+        pdf_container.date_start_edit_link = reverse('proposals:update_date_start',
+                                                     args=[proposal.pk])
```
tymees, line 228: I don't really like adding this link to the document list code. You're not updating documents. 

This should be a review action instead, I think.

tymees (changes requested):
I didn't test it locally, but I don't see anything would warrant a local test assuming you did test your code ;)

2 textual changes I'd like to see. The other two comments are niceties, I'm fine with as they are now if you don't think the niceties are required

proposals/templates/proposals/proposal_update_date_start.html
EdoStorm96: I added the `DateField` widget from cdh core

reviews/templatetags/documents_list.py
EdoStorm96: I agree that does make more sense! fixed

PR #573, opened 2023-11-08
Update dependencies for develop

tymees:
This was prompted by needing to fix two problems:
- We were still using an alpha version of DSC 3.1, even though the release version 3.1 is already out.
- djangosaml2 was pinned to <1.7 because of a bug. I submitted a patch for it upstream, and 1.8 has been released with this patch included. So, now it's better to unpin it (especially as my patch also patches a security flaw). 

During testing with these new deps, I quickly discovered that `xhtml2pdf` now requires 'cairo' development files to be installed. Under Fedora I had to install `cairo-devel`, but Ubuntu doesn't have a package with a equivalent name. 

I _think_ it's `libcairo2-dev`, but I cannot test that as I currently have no Ubuntu-based OS I can (easily) test on. Can y'all test that? We need to update our deployment code to install the relevant library... 

(Note: I did test if the xhtml2pdf code still works with the new cairo backend. It does, but I did find 2 unrelated PDF bugs :P)


EdoStorm96 (commented):
As I understand it correctly you just wanted us to test installing the requirements, with `libcairo2-dev` installed on our system, right? That works fine. It came pre-installed on Ubuntu for me.
Also the portal runs fine, based on first impressions. I can generate PDFs and stuff.

tymees:
> As I understand it correctly you just wanted us to test installing the requirements, with libcairo2-dev installed on our system, right? 

Yes! (And it was also intended as a warning that a new syspackage was needed lol)

> That works fine. It came pre-installed on Ubuntu for me.
> Also the portal runs fine, based on first impressions. I can generate PDFs and stuff.

Okay; hopefully it's that package then. I guess we'll find out on the test server then.

miggol:
`libcairo2-dev` was also already installed on my machine, didn't need to add anything. The Github actions runner doesn't seem to mind, either.

The best way to test these dependencies would be building a Docker container, but I don't have the time to do that right now. I would suggest we just keep an eye out for anything missing during the next deployment.

Issue #572, opened 2023-11-08
Non-standard proposals using the wrong PDF template


tymees:
At least pre-approved PDF's still use the old templates on the latest develop, using horrible tables. I didn't test, but it's highly likely the same thing happens with pre-assesments (as the logic that controls the used template is the same).

I think during the manual merge of the new PDF code the template-select logic was overlooked. The canonical-PDF change moved the logic to outside the main PDF logic, to the proposal model. As that would not have caused conflicts, it's probable that nobody noticed it.

EdoStorm96:
I can confirm that pre-assessment proposals do indeed use the old template. I'm also thinking now, that pre-assessments proposals have really been accounted for in the new PDF workflow, so that should also happen probably. Woops ... Where is the PDF logic you are talking about here located?

tymees:
In the `Proposal` model, there's a method called `pdf_template_name`

EdoStorm96:
Ahh ok. I guess it should be a fairly easy fix then? I will have a look.

EdoStorm96:
I propose that we also get rid of these unused templates, like `proposal_pdf_empty.html`, `proposal_pdf_pre_assessment.html` and `proposal_pdf_pre_approved.html`. As they are no longer neccessary/ cause confusion. I guess I should have done that already in the new pdf PR ... but better late than never

tymees:
Yeah, it should be an easy fix. Those templates probably just need to go. Didn't do it myself because 1) I wasn't entirely sure what was the cause and 2) I was busy testing something else

PR #574, opened 2023-11-09
Incorporate non-standard proposal in pdf workflow

EdoStorm96:
Fix for issue #572. I got rid of now obsolete pdf templates and made sure the pre-assessment and pre-approval proposals are properly incorporated in the new PDF workflow.

Maybe the `pdf_template_name` function can be removed entirely? I guess it is not being used in this version. What do y'all think?

@tymees I did notice that the change to `https` in the `handle_field_file` function of the `RowValueClass` causes an error in my browser when clicking links in the pdf, with the error code: SSL_ERROR_RX_RECORD_TOO_LONG

However, Django tells me this is because the development server only support http:

`You're accessing the development server over HTTPS, but it only supports HTTP.`

So it will be fine on the production server?


tymees:
> @tymees I did notice that the change to `https` in the `handle_field_file` function of the `RowValueClass` causes an error in my browser when clicking links in the pdf, with the error code: SSL_ERROR_RX_RECORD_TOO_LONG
> 
> However, Django tells me this is because the development server only support http:
> 
> `You're accessing the development server over HTTPS, but it only supports HTTP.`
> 
> So it will be fine on the production server?

I'll look at this PR tomorrow in-depth. But for now:

Yeah, it's because local dev doesn't have https. Normally, you'd solve this by retrieving the used protocol (http vs https) from the request object; but PDF gen doesn't (always) have access to that.

It will be fine on the server, those are HTTPS only. You could maybe create an if on the DEBUG setting, but I don't have any strong opinions tbh

proposals/models.py
```diff
@@ -703,13 +703,7 @@ def generate_pdf(self, force_overwrite=False):
...
     @property
     def pdf_template_name(self):
-        """Determine the correct PDf template for this proposal."""
-        template_name = 'proposals/proposal_pdf.html'
-        if self.is_pre_approved:
-            template_name = 'proposals/proposal_pdf_pre_approved.html'
-        elif self.is_pre_assessment:
-            template_name = 'proposals/proposal_pdf_pre_assessment.html'
-        return template_name
+        return 'proposals/proposal_pdf.html'
```
tymees, line 706: I agree that this is redundant now; differences should be handled by the new PDF framework

tymees (approved):

proposals/models.py
EdoStorm96: I got rid of this now, and its reference in the `ProposalAsPDF` view. This works fine for me. I was wondering does this now require a migration?

tymees: No it doesn't. Migrations are only needed for ORM Fields (which are always class variables, and (almost) always named `xField`). This is a class method, those _should_ never trigger migrations.

If you are unsure, you can run `python manage.py makemigrations --dry-run`. This will only display the migrations Django _wants_ to create, without actually creating them. 

Alternatively, we now have a check for PR's that will tell you if you are missing a migration. The latest run passed successfully: https://github.com/DH-IT-Portal-Development/ethics/actions/runs/6847879373/job/18616976001?pr=574

Issue #554, opened 2023-10-17
Put applicant in cc for affiliated applicants mail


tymees:
#292 added a new mail sent to the others applicants upon proposal submission.

Apperantly people expect the main applicant to be put in CC; so we should do that.

The text of the mail should also mention the main applicant was added as cc:

> Uw collega {NAME} (cc) heeft een aanvraag ingediend

Issue #549, opened 2023-10-06
Remove E-prime from registration from fixture


tymees:
It's already removed from the live server, removing it from the fixture is just to make sure it will never re-appear

Requested by the lab, as E-prime is a big no-no

Issue #529, opened 2023-08-24
Review all automatic mails


tymees:
The emails in the app have an inconsistent use of formal/informal language (in Dutch) between the emails, as discovered during the review of #528

All emails should be checked and changed if necessary, as it should all be informal per the UU writing styleguide.

djhcapel:
Not sure if it makes sense to mention it here, but I happened to come across a 'short route' message, which contains at least one link that does not lead to the right information (between ** text that should be adapted):

"The FEtC-H has received your application. The automatic screening of the portal has given your **application** the status of short-route research. This means that **we aim to send you** an assessment within two working weeks.

Note: It is possible that your application may still be classified as a 'long-route research', in which case the asssessment procedure may take longer, see the [regulations](https://fetc-gw.wp.hum.uu.nl/en/regulations-fetc-h/) for information. **DELETE: If the latter case, see [section 2.8, p. 14](http://fetc-gw.wp.hum.uu.nl/regulations/) for the possible time paths.
Please be aware of these possibilities and the different time paths involved.**

djhcapel:
I just noticed the Dutch text is okay. The English version should include the text: "Houdt hier a.u.b. rekening mee."

Issue #514, opened 2023-07-10
Remove mo file from tree


tymees:
AFAIK we put the mo file in our gitignore in other projects.

It makes conflicts unnecessarily annoying, and we recompile said file on deploy anyway. 

Thus, I'd like to propose we nuke the file from the repo and add it to the gitignore. The only real downside is that we have to manually recompile our local mo file from time to time

PR #576, opened 2023-11-10
Low hanging fruit mass-PR

### LINKED ISSUE #430: Feedback functionality; still needed?
Is anyone ever using the feedback module?

Do people even know it exists?

tymees:
Fixes #430 #514 #549 #550 #554 and a bonus non-issue unused template removal

Basically, all low hanging fruit aside from #529 as Desiree is still OOO. (I wanted something simple to do for once lol)

As they were all 5-min fixes I didn't really feel like creating separate PR's for them, but I did solve all issues in their own commit. 

The only real big impact is the removal of feedback; the deploy script needs to be updated to not try to load the now missing app.

It would be nice if this could be merged at the same time as #573, as that needs deploy script changes as well. As for the content: they are pretty simple, but as always I'd like double check of my english


EdoStorm96 (approved):
All looks good to me. Nice to get rid of these minor issues!

miggol (approved):
Nice work. I appreciate the cleanup of the menu file as well, that one tripped me up once or twice.

Went through all text and translations and ran through a dev environment with no issues.

Somehow I'm a little surprised that deleting an app does not generate a migration. Guess those tables will just linger. Fine by me.

Issue #578, opened 2023-11-14
PDF - chamber information missing


djhcapel:
It would be helpful to know to which chamber a particular application belongs. Could that info be added to the header of the pdf? It now only gives the title, reference number and applicant's name. If only 'LC' or GC' could be added to the reference number, that would be sufficient.

Issue #575, opened 2023-11-10
Supervisor and reviewer: no solisid but full name


djhcapel:
I just noticed that in the 'details' of an application, the solisid is given for the supervisor; could we change this to the full name?

djhcapel:
I also noticed that solisids are used for reviewers on the 'Details' page. Could these also be replaced by names?

tymees:
> I also noticed that solisids are used for reviewers on the 'Details' page. Could these also be replaced by names?

This issue was closed as the original request was fixed (and GH auto-closed because the fix was accepted). 

I've created a new issue for this request: #581

Issue #491, opened 2023-06-02
Privacy officer consulted? yes/no


djhcapel:
On the Data Management page, applicants are required to tick a box "Ik heb kennis genomen van het bovenstaande en begrijp mijn verantwoordelijkheden ten opzichte van de AVG.". Now that more and more people first discuss their study with the privacy officers (and data manager), I would like to change this to the following question, and instead of a tick box, a radio button yes/no:

"Ik heb mijn aanvraag en de documenten voor deelnemers besproken met de privacy officer. O ja O nee"

It will then be easier for me to check whether any correspondence about such study has been sent to the fetc-gw mailbox.

miggol:
The question will no longer be required to proceed.

djhcapel:
Question: where would I be able to see the response to this question?
Could it be incorporated in the pdf?

PR #579, opened 2023-11-15
Feature/4 small issues

### LINKED ISSUE #350: Tag reviews as supervisor reviews properly

EdoStorm96:
This PR fixes #350, #575 and #578. 

- 350: This was the biggest one. I added an `is_commission_review` field to the `Review` model, so that there is a simple check to see if a review is a supervisor review, besides the stage. Some things to note: 
    - Default is `True`, only get changed when creating a supervisor review. 
    -  I wrote a data migration, which worked well for me, but have a look at that.
    - I wrote some tests for this, which was a first for me. I guess this is fine? But I am open to suggestions.
- 575: Just make sure the supervisor's name, instead of solisID is displayed on the review detail page.
- 578: Added Chamber information to the PDF header. Right now, just using the acronyms, as this was requested by Desiree. It now looks like this, at the top of the page:

    'FEtC-H - _Looking at the elderly_rev_ (reference number 23-019-01, submitted by Justin Bieber) - LC'

    The code in the template is a bit ugly ... I was kindoff struggling with keeping the chamber name translatable, but IMO this is fine.


EdoStorm96:
Ok, so I decided to also incorporate a fix for #491 here. I hope this PR is not getting too messy for you both ... 

491:
So this alters how the `avg_understood` field works. It is now nullable, the widget for it has changed to a radioselect button, and it can be either yes or no, but not be left blank. This field is now also displayed in the pdf & diff.

I did run into one issue here ... The `AVGUnderstoodValidator` function (the logic of which I updated in line with the field's new requirements) was seemingly not doing anything for me, so I tried removing it entirely. This seems to not be possible, when making migrations. So now it is just there, but not really doing anything? I created an error message in the form's `clean` method, which works fine, but could/should I remove the validator? Or is it best to keep it and get rid of my new error message? I did not really see the purpose of this validator ...

miggol (changes requested):
Right off the bat, I see that there's a new way to differentiate supervisor reviews (yay) but that we still don't differentiate between open and closed supervisor reviews. Could you add that functionality, such that concluded supervisor reviews are given the stage `review.CLOSED`? This logic lives in `Review.update_go()` currently.

I wanted to bring that up right away, because closing those reviews might bring up more required changes, and would also require changes to the data migration to close all old supervisor reviews.

I only had time to look at the #350 part of this PR so far. I will review this further at a later time.

proposals/models.py
```diff
@@ -400,9 +400,10 @@ class Proposal(models.Model):
...
     avg_understood = models.BooleanField(
-        _('Ik heb kennis genomen van het bovenstaande en begrijp mijn verantwoordelijkheden ten opzichte van de AVG.'),
-        default=False,
-        null=False,
+        _('Ik heb mijn aanvraag en de documenten voor deelnemers besproken met de privacy officer.'),
+        default=None,
+        null=True,
+        blank=True,
         validators=[AVGUnderstoodValidator],
     )
```
tymees, line 408: I'm not sure if changing the label/help text of an existing, different question, is the way to go here. You're effectively rewriting history a bit. 

It's less of a problem now that we have canonical PDF, as the record will not change (and thus, the proper history is preserved). 

However, there are still two issues;
From a developer view, the variable name is confusing now. 
From an applicants view, upon revision/copy/amendment we're going to fill in 'yes' to all existing proposals for this question. (As it was required to be true before). This is less than ideal, as ideally we want to force applicants to re-evaluate if they actually did/need to contact the privacy officer. 

I'd remove the whole field and add another; that should also fix your migration error ;)

tymees (changes requested):
Fully agree with Michael's notes about #350, the whole point was to be able to _close_ those reviews as well. (The issue was created because of a different issue partly caused by supervisors reviews never closing).

I did take a look at the other issues. The fix for  #575 is a automatic approve ;)
I do have comments for the others

tymees:
One last thing: you only added the `is_committee_review` variable. It would be nice if a app-wide pass was made to actually use it in QuerySets over the old annoying method

EdoStorm96:
Ok, so I implemented the changes you both requested, some notes:
- I updated the `Review.update_go()` method as Michael suggested to close supervisor reviews. I now use this method on all past reviews in the data migration, to update their stage when relevant. See review migration 0013.
- So I did end up replacing field `avg_understood` with `privacy_officer`, as Ty suggested. This does indeed make more sense.
- To get rid of the `AVGUnderstoodValidator` I removed it and all references to it from the repo. However, this caused a problem, because it got referenced in proposal migration 0034. As suggested in the Django docs, I just placed this function in this migration (as an empty function) and now all seems to work fine. (Despite this solution looking shady to me ...)

reviews/migrations/0013_is_commission_review_field.py
```diff
@@ -11,6 +11,7 @@ def update_is_commssion_review(apps, schema_editor):
         SUPERVISOR_STAGE = 0
         if review.stage == SUPERVISOR_STAGE:
             review.is_commission_review = False
+            review.update_go()
```
tymees, line 14: In theory this should work... However, the production DB isn't as clean as we would like. 

I'd feel better if the code was something like this:
```
if review.go:
    review.stage = CLOSED
```

It's a lot simpler, and should be safer.

reviews/templates/reviews/review_detail_sidebar.html
```diff
@@ -90,7 +90,7 @@ <h3>
                 </em>
             </li>
         {% endif %}
-        {% if review.stage == review.SUPERVISOR %}
+        {% if review.is_commission_review == False %}
```
tymees, line 93: It's cleaner to write `{% if not review.is_commission_review %}`

reviews/views.py
```diff
@@ -556,7 +556,7 @@ def form_valid(self, form):
 
         # Don't notify the secretary if this is a supervisor decision.
         # If it was a GO they the secretary will be notified anyway
-        if not review.stage == review.SUPERVISOR:
+        if review.is_commission_review == True:
```
tymees, line 559: `== True` is redundant

proposals/templates/proposals/proposal_pdf.html
```diff
@@ -84,10 +84,9 @@
 
 {% block page_header %}
   <div id="page-header">
-    {% blocktrans with title=proposal.title reference_number=proposal.reference_number submitter=proposal.created_by.get_full_name trimmed %}
-      FETC-GW - <em>{{ title }}</em> (referentienummer {{ reference_number }}, ingediend door {{ submitter }})
+    {% blocktrans with title=proposal.title reference_number=proposal.reference_number submitter=proposal.created_by.get_full_name reviewing_committee=proposal.reviewing_committee.name trimmed %}
+      FETC-GW - <em>{{ title }}</em> (referentienummer {{ reference_number }}-{{reviewing_committee}}, ingediend door {{ submitter }})
```
tymees, line 88: Hate to be that guy... But conventionally we place the committee _before_ the rest of the refnum.

Sorry, should've clarified that in my original comment

tymees (changes requested):
Some minor tweaks, and it should be good to go :)

EdoStorm96:
All right, so I've implemented Ty's minor tweaks, but, more importantly, I realized that it should be `is_committee_review` and not `is_commission_review`. This is a typo that occurred when discussing the original issue (#350) and was then adopted in this discussion and also by me ;p I re-ran the migrations on my branch, so that there is no evidence of this. I think it should be all good now!

tymees (approved):

Issue #581, opened 2023-11-20
Use fullname for reviewers on the 'Details' page


tymees:
I also noticed that solisids are used for reviewers on the 'Details' page. Could these also be replaced by names?

_Originally posted by @djhcapel in https://github.com/DH-IT-Portal-Development/ethics/issues/575#issuecomment-1818358531_

tymees:
Not sure how feasible this is, as AFIAK this is some widget stuff pulling in data under the hood, but I could be wrong

Issue #563, opened 2023-10-31
Warning assessments complete


djhcapel:
Dear colleagues, could the secretary get a warning once assessments have been received by all allocated members? E.g. by the number of applications that is ready after "Nog af te handelen aanvragen" in the menu? 

Nog af te handelen aanvragen [3]

Or by an automatically generated email?

tymees:
Send an automated mail when:
- all decisions have been made

Should just be an `update_go` mod

PR #582, opened 2023-11-20
Feature/email secretary after decisions

EdoStorm96:
Some fixes for a two small issues. 

the fix for #581 is pretty self explanatory.

For the issue #563, I updated the `update_go()` of `Review`, as Ty suggested. Please check if this is in the correct spot ... In my testing this works as expected.

Furthermore, I'd appreciate feedback on the text of the e-mail. It's pretty bare bones, but gets the job done, with an overview of all decisions.

PS. I accidentally blacked review_utils.py, but undid it, because it made the diff unreadable.

### DIFF
--- a/reviews/utils/review_utils.py
+++ b/reviews/utils/review_utils.py
@@ -326,7 +326,7 @@ def notify_secretary_all_decisions(review):
         + (reverse("reviews:close", args=[review.pk])),
     }
     msg_plain = render_to_string("mail/all_decisions_notify.txt", params)
-    send_mail(subject, msg_plain, settings.EMAIL_FROM, ["fetc-gw@uu.nl"])
+    send_mail(subject, msg_plain, settings.EMAIL_FROM, [settings.EMAIL_FROM])
 
     # Reset the current language
     activate(current_language)
--- a/reviews/utils/review_utils.py
+++ b/reviews/utils/review_utils.py
@@ -52,62 +52,39 @@ def start_supervisor_phase(proposal):
     decision = Decision.objects.create(review=review, reviewer=proposal.supervisor)
     reference = proposal.committee_prefixed_refnum()
 
-    subject = _("FETC-GW {}: bevestiging indienen concept-aanmelding".format(reference))
+    subject = _('FETC-GW {}: bevestiging indienen concept-aanmelding'.format(reference))
     params = {
-        "proposal": proposal,
-        "title": proposal.title,
-        "supervisor": proposal.supervisor.get_full_name(),
-        "secretary": get_secretary().get_full_name(),
-        "pdf_url": settings.BASE_URL + proposal.pdf.url,
+        'proposal': proposal,
+        'title': proposal.title,
+        'supervisor': proposal.supervisor.get_full_name(),
+        'secretary': get_secretary().get_full_name(),
+        'pdf_url': settings.BASE_URL + proposal.pdf.url,
     }
-    msg_plain = render_to_string("mail/concept_creator.txt", params)
-    msg_html = render_to_string("mail/concept_creator.html", params)
-    send_mail(
-        subject,
-        msg_plain,
-        settings.EMAIL_FROM,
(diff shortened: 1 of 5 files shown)


reviews/utils/review_utils.py
```diff
@@ -304,6 +304,21 @@ def notify_secretary(decision):
...
+    subject = _("FETC-GW {}: alle beoordelingen toegevoegd").format(
+        review.proposal.committee_prefixed_refnum(),
+    )
+    params = {
+        "secretary": secretary.get_full_name(),
+        "proposal": review.proposal,
+        "decisions": review.decision_set.all()
+    }
+    msg_plain = render_to_string("mail/all_decisions_notify.txt", params)
+    send_mail(subject, msg_plain, settings.EMAIL_FROM, [secretary.email])
```
tymees, line 321: As per #563:
> [..] FETC npm (not users!) [..]

Thus, the recipient should be the general fetc email ;)

reviews/utils/review_utils.py
```diff
@@ -304,6 +304,21 @@ def notify_secretary(decision):
...
     send_mail(subject, msg_plain, settings.EMAIL_FROM, [secretary.email])
 
+def notify_secretary_all_decisions(review):
+    """
+    Notifies a secretary all Decisions have been made for a certain review
+    """
+    secretary = get_secretary()
+    subject = _("FETC-GW {}: alle beoordelingen toegevoegd").format(
+        review.proposal.committee_prefixed_refnum(),
+    )
```
tymees, line 314: You're translating the string, but not adding any translations...

Honestly, as the email is not translated, I'd just drop the `_([..])` tag altogether

tymees (changes requested):
Nice quick turnaround lol :)

Two small errors, otherwise: looks great!

About the email: looks fine; during the meeting we all agreed the contents should not matter that much, the subject is more important.

reviews/utils/review_utils.py
```diff
@@ -304,6 +304,21 @@ def notify_secretary(decision):
...
+    secretary = get_secretary()
+    subject = "FETC-GW {}: alle beoordelingen toegevoegd".format(
+        review.proposal.committee_prefixed_refnum(),
+    )
+    params = {
+        "secretary": secretary.get_full_name(),
+        "proposal": review.proposal,
+        "decisions": review.decision_set.all()
+    }
+    msg_plain = render_to_string("mail/all_decisions_notify.txt", params)
```
miggol, line 320: Any translations existing in this input, e.g. `decision.get_go_display`, will render to the currently selected language.

If the user who triggered the `update_go()` happened to be set to English, all translations in here will also turn into English, which is a bit weird. 

See `proposal_utils.notify_local_staff()` for an example of how we manually activate a language for email sending, and then reset it afterwards.

miggol (changes requested):

reviews/utils/review_utils.py
```diff
@@ -304,6 +304,33 @@ def notify_secretary(decision):
...
+        "review": review,
+        "decisions": review.decision_set.all(),
+        "review_detail_page": settings.BASE_URL
+        + (reverse("reviews:detail", args=[review.pk])),
+        "close_review_page": settings.BASE_URL
+        + (reverse("reviews:close", args=[review.pk])),
+    }
+    msg_plain = render_to_string("mail/all_decisions_notify.txt", params)
+    send_mail(subject, msg_plain, settings.EMAIL_FROM, ["fetc-gw@uu.nl"])
+
```
tymees, line 330: I should've clarified; you should use `settings.EMAIL_FROM` as the recipient.

It's kinda weird, I know, but `EMAIL_FROM` is the NPM email. Hardcoding this is a bad idea, as names (and thus, emails) in the UU tend to change. It's better if we only have to change the one value when that happens

tymees (changes requested):
One tiny tweak left, and then it's GTG

You don't have to re-request review from me if you fix that tiny detail

PR #586, opened 2023-11-23
Fix/refactor choices

### LINKED ISSUE #580: Refactor constant + tuple based choices to models.xChoices
Django 3 added special [enumeration types](https://docs.djangoproject.com/en/4.2/ref/models/fields/#field-choices-enum-types) for `choices`. We're not using them, as the app is old enough to have known both Django 1.6 and Python 2 :P

However, for models with multiple sets of choices it's somewhat hard to figure out which ones belong together. Using these enumeration types will make things more clear. 

The refactor can be done without migrations, if done correctly. (As those enumeration types generate the same format as the old method in the end).
 
Not a priority, just a nice to have.

EdoStorm96:
Fixes #580. Refactoring all the choices in the models to use Django's special [enumeration types](https://docs.djangoproject.com/en/4.2/ref/models/fields/#field-choices-enum-types). Had to change the code in lots of places, but really nothing major has happened. The most important changes happened in the models. When performing `makemigrations --dry-run` no changes are detected. A few things to pay attention to:
- I removed `DESIGNS` from the `Studies` model entirely, as these seem to not be used anywhere.
- I had to slightly adjust the list comprehension on line 105 of reviews/forms.py. It works fine like this, i guess. Although I do wonder if there is a better solution with the new enumeration types ...
- Also note the changes in `get_continuation_display()`. Again, this works, but I wonder if there could be a better solution?
- Lastly, keep in mind the new syntax that this change requires, in the future, eg.:
    - No longer: `Review.SUPERVISOR`, but: `Review.Stages.SUPERVISOR`
    - No longer: `choices = STAGES`, but `choices = self.Stages.choices` in models or `Review.Stages.choices` in forms
    - No longer: `Import YES, NO, DOUBT, YES_NO_DOUBT`, but `import YesNoDoubt`


reviews/models.py
```diff
@@ -127,14 +113,11 @@ def get_continuation_display(self):
...
-        if self.stage <= Review.COMMISSION:
+        if self.stage <= Review.Stages.COMMISSION:
             return _("Onbekend")
 
         # Get the human readable string
-        continuation = dict(self.CONTINUATIONS).get(
-            self.continuation,
-            self.continuation
-        )
+        continuation = dict(self.Continuations.choices)[self.continuation]
```
tymees, line 120:
```suggestion
        continuation = self.Continuations(self.continuation).label
```

This is the proper way to get the label for a choice :)

proposals/utils/pdf_diff_logic.py
```diff
@@ -286,9 +286,9 @@ def handle_field_file(self, field_file):
         return output
 
     def yes_no_doubt(self, value):
-        from main.models import YES_NO_DOUBT
+        from main.models import YesNoDoubt
 
-        d = dict(YES_NO_DOUBT)
+        d = dict(YesNoDoubt.choices)
         return d[value]
```
tymees, line 292: No need for the dict cast:
```python
        from main.models import YesNoDoubt

        return YesNoDoubt(value).label
```

reviews/forms.py
```diff
@@ -102,7 +102,7 @@ def __init__(self, *args, **kwargs):
         allow_long_route_continuation = kwargs.pop('allow_long_route_continuation', False)
         super(ReviewCloseForm, self).__init__(*args, **kwargs)
         if not allow_long_route_continuation:
-            self.fields['continuation'].choices = [x for x in Review.CONTINUATIONS if x[0] != Review.LONG_ROUTE]
+            self.fields['continuation'].choices = [x for x in Review.Continuations.choices if x != Review.Continuations.LONG_ROUTE]
```
tymees, line 105: This does not work?:

```
>>> [x for x in Review.Continuations.choices if x != Review.Continuations.LONG_ROUTE]
[(0, 'Goedkeuring door FETC-GW'), (1, 'Revisie noodzakelijk'), (2, 'Afwijzing door FETC-GW'), (3, 'Open review met lange (4-weken) route'), (4, 'Laat opnieuw beoordelen door METC'), (5, 'Positief advies van FETC-GW, post-hoc'), (6, 'Negatief advies van FETC-GW, post-hoc'), (7, 'Niet verder in behandeling genomen')]
```
I think it should be this:

```suggestion
            self.fields['continuation'].choices = [x for x in Review.Continuations.choices if x[0] != Review.Continuations.LONG_ROUTE]
```


As for a better solution: there isn't any really that's clearly better. One alternative is a filter:
```python
filter(lambda x: x[0] != Review.Continuations.LONG_ROUTE, Review.Continuations.choices)
```
But that's not really that much better.

You could also create a method in the new class to do the list comprehension there, but for this one usage that does not make much sense

EdoStorm96: I'm sorry. I just assumed ... Too bad that does not work. I just used your first suggestion.

proposals/utils/pdf_diff_logic.py
miggol, line 1: Missed a couple of `model.wmo.NO_WMO` instances on lines 1017 and 1022

reviews/tests.py
miggol, line 1: Found some unchanged doubts and no's on lines 18 and 32.

These flagged up in `python manage.py test`. Usually you would see this in the PR on Github as well, but it seems there's some issue with the runners right now. If this persists/happens again you can always run the tests locally to check.

proposals/tests.py
```diff
@@ -137,13 +137,13 @@ def test_revision_reference_number(self):
...
     def test_status(self):
         proposal = self.p1
-        self.assertEqual(proposal.status, Proposal.DRAFT)
+        self.assertEqual(proposal.status, Proposal.Statuses.DRAFT)
 
-        wmo = Wmo.objects.create(proposal=proposal, metc=YES)
-        self.assertEqual(proposal.wmo.status, Wmo.WAITING)
+        wmo = Wmo.objects.create(proposal=proposal, metc=YesNoDoubt.YES)
+        self.assertEqual(proposal.wmo.status, Wmo.WMOStatuses.WAITING)
         wmo.metc = NO
```
miggol, line 144: Stubborn metc can't take `NO` for an answer.

proposals/tests.py
```diff
@@ -279,22 +279,22 @@ def setUp(self):
         self.wmo = Wmo.objects.create(proposal=self.p1, metc=NO)
```
miggol, line 279: Undefined `NO`

proposals/views/wmo_views.py
```diff
@@ -139,8 +139,8 @@ def check_wmo(request):
     """
     This call checks which WMO message should be generated.
     """
-    is_metc = request.POST.get('metc') == YES
-    is_medical = request.POST.get('medical') == YES
+    is_metc = request.POST.get('metc') == YesNoDoubt.YES
+    is_medical = request.POST.get('medical') == YesNoDoubt.YES
 
     doubt = request.POST.get('metc') == DOUBT or request.POST.get('medical') == DOUBT
```
miggol, line 145: Undefined `DOUBT`

miggol (changes requested):
Nice, I like the new choices. I reviewed as follows:

* Ran tests
* Tried to make a pdf
* Migrated from scratch and loaded all fixtures
* Finally, pointed out all the places my IDE told me there were undefined variables like `DOUBT` and `NO_WMO`

I found a few issues, mainly from the final step.

Issue #536, opened 2023-09-19
Add link to revise proposal to list page


tymees:
Very low priority issue; some people expect a link on the list pages, which makes some sense

Issue #259, opened 2021-09-17
ProposalCopy - copy.js grabs study title with refnum


miggol:
Refnum in title of parent study is a great improvement. But it is now included in the initial value for the new title. There's already a hack in `copy.js` which solves this for the '(practice)' and last modified annotations:

```js
$(function () {
     $('#id_parent').change(function () {
         // Get the displayed value
         let chosen = $( "#id_parent option:selected" ).text();
         // Split on the ( and trim to remove the user part
         chosen = chosen.split('(')[0].trim();
         // Set it as the new title
         $('#id_title').val(chosen);
     })
});
```

But this beautiful mess obviously fails if there's an opening parenthesis in the study title (I couldn't resist). So I suggest we just remove the title field from the copy form altogether. Proposal titles can be changed after the fact anyway nowadays and there's no longer a uniqueness constraint IIRC.

PR #587, opened 2023-11-28
Fix/3_small_issues

EdoStorm96:
I fixed some more minor issues. They are each in a separate commit. 

#259: So, as suggested by Michael, I opted to remove all references to the title field in the copy forms. Which allowed me to also remove copy.js. To me, it seems like these were just artifacts of an older version.

#536: Probably requires the most attention in this pr. I really wanted a button to revise a proposal per item in the list. I opted to use the pencil.png image for this. With some brute forcing, I managed to figure out the Vue template, so that it basically grabs the same proposals, as the `_get_parent_queryset()` in `RevisionProposalCopyForm`. Then, to these specific items it adds the pencil image in the list, which links to `proposals:copy_revision`. I think this implementation is pretty nice, as it shows which proposals are revisable. But I am curious to hear your thoughts.

Furthermore, I had to change the `ProposalSerializer`, to pass the correct attributes to the Vue template.

Lastly, please have a look at the logic in the Vue template: `proposals/templates/proposals/vue_templates/proposal_list.html` (lines 77-83). I guess this is fine, but I was struggling quite a lot with how to check whether a proposal has children. This works, but is maybe a bit confusing. I am open to other suggestions on how to implement this.

#584: Yup, that a minor textual change.


proposals/forms.py
```diff
@@ -275,11 +275,6 @@ class Meta:
         widgets = {
             'is_revision': forms.HiddenInput(),
         }
-        error_messages = {
-            'title': {
-                'unique': _('Er bestaat al een aanvraag met deze titel.'),
-            },
-        }
```
tymees, line 282: Dropping the unique check sure does solve the issue easily, but it's kinda drastic.

This is a big functional change that should be discussed with the secretary first.

Especially as there's a pretty easy fix (albeit less easy than this fix...) that does not involve doing this...

proposals/templates/proposals/vue_templates/proposal_list.html
```diff
@@ -73,15 +75,17 @@ <h4>
...
+                !proposal.status_review &&
+                proposal.children.length === 0"
+                :href="$url('proposals:copy_revision')"
             >
-                <img src="{{ img_hide }}"
-                     title="{% trans 'Verberg' %}">
+                <img src="{{ img_revise }}"
+                     title="{% trans 'Maak revisie' %}">
             </a>
         </template>
```
tymees, line 87: Why are you removing an action for this?

Secondly, you could also implement this by creating a property-method (`is_revisable`) on `Proposal`, and then passing that as a field in the serializer. Honestly, it's not a bad idea to have that property for other checks in Python as well. Just a suggestion, this works fine :)

tymees (commented):
Not passing judgement just yet, as I'd like to see the solution for #259 be discussed first.

proposals/templates/proposals/vue_templates/proposal_list.html
EdoStorm96: Removing the action was a little mistake. I did like your suggestion of creating an `is_revisable` property method, so that exists now.

proposals/forms.py
tymees: Nooo, it's me that's being an idiot... Sorry!

I forgot we changed that lol. Your solution is correct.

EdoStorm96:
I decided to also include a fix for #590 here. Nothing crazy ... Just another minor textual change. It did require a migrations though.

miggol (approved):
All good!

PR #589, opened 2023-11-30
Fix/refactor more choices

EdoStorm96:
So, it turns out I missed some things on PR #586 ... Mostly stuff in templates or cases where the values are gained from instances of models, rather than the models themselves. I am sorry for not having been more thorough on the first pass ... I thought my IDE would be smart enough to catch all references, but it wasn't. So now I checked everything with regex as well, as Ty initially suggested, and am confident that this caught all effected references.


tymees (approved):
I did a manual check using the very permissive regex `[a-z]+\.[A-Z]+[^a-z]`, and could not find anything else either.

Issue #583, opened 2023-11-21
New revisions - add an option to 'Commissieleden aanstellen'


djhcapel:
An option in case of new revisions is actually missing. Current options are:

 O korte (2-weken) route
 O lange (4-weken) route
 O direct naar revisie

Could the option 'O revisie' be added?

The present regulations do not mention an assessment deadline; do you need one? Then for the time being, used 2 weeks.

EdoStorm96:
Could you further elaborate how the new field would differ from the old field 'O direct naar revisie'?

EdoStorm96:
We can for now just change the text of the short route option, to 'Korte route of revisie' as revisions will usually also take about two weeks. We should also update the text of 'direct naar revisie' to 'Direct terug naar de aanvrager'

Issue #539, opened 2023-09-25
Access chairs, privacy officer and data manager


djhcapel:
If an application is sent for revision, we often advise the researcher to contact the privacy officer/data manager for advice. These advisory members of the FEtC-H, can, at that point, not be linked to the application anymore (status: revisie noodzakelijk). Is there a way by which they can access all applications without having the secretary role? That would be the easiest solution (from the secretary's point of view), since researchers may also seek advice from p.o. and/or d.m. on their own initiative.

djhcapel:
Actually, this kind of access also applies to both chairs of the FEtC-H.

It should be clear to any of these people which applications they need to assess.

miggol:
Okay, seems like a good idea. Access controls are quite tight as they stand.

Currently, both chairs already have access to both chambers. But they can only see applications that they have been assigned to.

In the future, the desired situation is as follows:
* Select users such as Chairs, PO, and DM should be able to view ALL applications of both chambers.

EdoStorm96:
Make a group that has permissions in combination with a room. Should be able to see everything, yet not edit things. 'kijkrechten'

PR #591, opened 2023-12-19
Feature/new group chairs etc

EdoStorm96:
This PR fixes two issues: a bigger one and a minor one. To start with the bigger one:

#539: So, I have added a new group, which is meant for chairs, the privacy officer and the data manager (`GROUP_CHAIR` in constants.py). This group can basically see everything that the secretary can see, as long as they are a also a member of the chamber in question. (see menus.py. for an idea ...) However, they cannot perform any actions. They can just look at all the lists and open up a detail view for a proposal. However, no review actions are available to this group. This is what we discussed with Desiree. For now, I've even granted this group access to the committee workload overview, as I thought it might be useful. What do you think about this? Otherwise, this seems to work well :)

#583: In the last meeting, we decided to just clarify this with a simple textual change instead of a new option. Most revisions end up taking about two weeks anyways, so for now, we just treat it as a short route review, behind the scenes. Desiree was satisfied with this solution


tymees (approved):
LGTM, one minor docstring that could fixing up but meh

djhcapel:
Actually, the workload is not meant for privacy officers and/or data manager, but I'm afraid this can't be solved, can it?

Issue #18, opened 2018-04-17
unordered_list syntax


tymees:
Most unordered_list calls don't actually contain ul tags in the html output, as the unordered_list command does not provide them (despite it's name). 

Adding it actually changed the layout in certain places, so that needs to be fixed too if implemented

PR #603, opened 2024-01-31
Feature/multiple sessions tasks

### LINKED ISSUE #373: Meerdere sessies en/of meerdere taken
In de portal staat nu: Voor elke sessie stellen we in de komende schermen steeds dezelfde vragen. 
en: Voor elke taak stellen we in de komende schermen steeds dezelfde vragen.

Dit is verschrikkelijk onhandig, omdat in de meeste gevallen die ik ben tegengekomen, de sessies en/of de taken gelijk zijn aan elkaar. Je krijgt dus onnodige herhaling van informatie die een aanvraag compleet onoverzichtelijk maakt (ik heb wel eens een aanvraag voor een onderzoek met 7 identieke taken gehad: een enorm lang aanvraagformulier werd dat). Kunnen we dit niet veranderen, zodat men kan aangeven dat er 5 sessies/taken zijn die allemaal hetzelfde zijn? Of, iets ingewikkelder, 6 sessies/taken waarvan er steeds 3 precies hetzelfde zijn? 

Bijvoorbeeld: Bij meer dan één: Zijn deze sessies/taken exact hetzelfde? ja: één keer laten invullen, nee, meerdere keren laten invullen. Hoe dat moet als er 3x2 identieke sessies/taken zijn weet ik niet; hebben jullie een idee?

EdoStorm96:
Fixes #373. For the solution to this problem, I went with Michael's proposed solution to the problem. Basically, both sessions and tasks are now created iteratively. So when done creating a session or a task, you can choose to make another or continue to the study overview (eventually). To make sure we get rid of overly long PDF's with repeating tasks and sessions, both session and tasks now have an attribute called repeats, which just signifies how often these will be performed. This means that behind the scenes, we just have one session object, which might get repeated three times, whereas we used to have 3 sessions for such a case. Implementing this of course had other implications:

- `sessions_`/`tasks_number` is now a model method, returning eg. `self.task_set.count()`
- I got rid of the `SessionStart` view, as that was just used for giving the initial `sessions_number` and is now longer needed. Its help text is now on the `SessionUpdate` view, which is (from the user POV) the first page you see when creating session.
- I changed up some of the names of the views, as I found them to be a bit confusing.
- `session.net_duration` is now a bit more spicy, as I needed to multiply `task.repeats` by `task.duration` and then sum that for all tasks.
- I added edit and delete buttons for everything in the `StudyEnd` view. Although this leads to a weird UI quirk for me of like an `_` appearing between the items?
- I had to edit `form_buttons.html` to accomodate for a "Save and create a new session/task >>" option.
- I had to make some changes to the pdf_diff_logic. I got rid of the `SessionOverviewSection`, as that was the equivalent of the `SessionStart` view and just showed the sessions number. To not lose its section title, I put some logic in the `SessionSection` to display its title, `if session.order == 1`.
- I removed most of the copying functionality, as that seemed obsolete, given you can now just say a session/task has repeats ...

Some last bugs/things to note/ideas:
- We probably need some new help text, explaining this process. What do you think would be a good place for that? I think the process is bit more intuitive, anyway ...
- When the empty `UpdateView` is rendered after `SessionCreate` or `TaskCreate` all field required warnings light up. I guess because the page gets loaded with a mostly empty object ... any ideas on how to overcome this?
- Maybe we can add a button to create a task or session to the `StudyEnd` page?

I'm pretty pleased with how this turned out. Based on my testing, this works well, but I might have missed some things. I'm curious to hear your suggestions! Good luck on the review and take your time! ;p


EdoStorm96:
I'm not sure what to do about the black PR checks ... When I attempt to run black on the files in question (some of which I have not changed in this branch) on my device, it leaves them unchanged.

tymees:
Haven't looked at the code yet, but regarding this comment:

> I'm not sure what to do about the black PR checks ... When I attempt to run black on the files in question (some of which I have not changed in this branch) on my device, it leaves them unchanged.

I ran into this problem as well yesterday @ DIAPP, it seems Black has been updated to include a new rule. (Resulting in the most annoyed commit message I've written lately).
Try updating black.

(Although, the changes in the migration file should've been picked up by your black version ;) )

tasks/models.py
```diff
@@ -9,20 +9,14 @@
 class Session(SettingModel):
     order = models.PositiveIntegerField()
 
-    # Fields with respect to Tasks
-    tasks_number = models.PositiveIntegerField(
-        _("Hoeveel taken worden er binnen deze sessie bij de deelnemer afgenomen?"),
-        null=True,
+    repeats = models.PositiveBigIntegerField(
+        _("Hoe vaak wordt deze sessie herhaald?"),
```
tymees, line 13: Suggestion:
"Hoe vaak zal de participant deze sessie doorgaan?"

But then using proper Dutch/English

In addition, a help-text with more details could be worth looking at.

tasks/models.py
```diff
@@ -139,6 +138,16 @@ class Task(models.Model):
         blank=True,
     )
 
+    repeats = models.PositiveBigIntegerField(
+        _("Hoe vaak wordt deze taak herhaald?"),
```
tymees, line 142: Suggestion:
"Hoe vaak moet de participant deze taak doen?"

But then, again, with proper Dutch/English

In addition, again, a help-text with more details could be worth looking at.

tymees:
For if Michael wants to take a look at this PR as well, Edo and I also did an in person look at this PR. During that conversation, I ended up pushing some changes:

- Removed the weird '_' Edo noticed
- By complete accident we realized the current develop _always submits_ the application on the submit page, even if one presses the back button. This is due to a change in `start_review` and some blocking conditions that were accidentally left out. I restored them
- I refactored the form buttons template. The previous design was very crowded (with all the different variations of back buttons). This had the added effect of making the new 'create new task' and 'create new session' buttons not stand out as a separate action. I removed all but one of the previous buttons (the 'prev step' button), as our stepper can also be used  for that. In addition, I added as much whitespace as possible between the prev and next buttons to visually indicate what they do.

studies/templates/studies/study_end.html
tymees, line 1: As discussed, it might be an idea to add some 'create new task/session' buttons on this page as well for when the user didn't notice they had that option during creation.

tymees (commented):
Just commenting because I didn't find any blockers, but also had a lot of small enhancement suggestions. (So both approve/request changes didn't feel applicable.)

The code looks good, and the flow is definitely better this way, so this is already pretty great. With some small additional tweaks it could be  awesome!

studies/templates/studies/study_end.html
```diff
@@ -131,11 +131,15 @@ <h4>{% trans "Takenonderzoek met de volgende opbouw:" %}</h4>
...
+                                        minuten, aantal herhalingen: {{ session.repeats }})
+                                        <a href="{% url 'tasks:session_update' session.pk %}">
+                                            <img src="{% static 'proposals/images/pencil.png' %}"
+                                                 title="{% trans 'Sessie verwijderen' %}">
+                                        </a>
+                                        {% if study.last_session.order > 1 %}
                                             <a href="{% url 'tasks:session_delete' session.pk %}">
                                                 <img src="{% static 'proposals/images/delete.png' %}"
-                                                     title="{% trans 'Sessie verwijderen' %}">
+                                                     title="{% trans 'Sessie aanpassen' %}">
```
miggol, line 142: Delete and edit tooltips are mixed up here.

proposals/utils/pdf_diff_logic.py
```diff
@@ -1037,7 +1032,6 @@ def create_context_pdf(context, model):
                     if study.has_observation:
                         sections.append(ObservationSection(study.observation))
```
miggol, line 1033: I think all the major sections like `ObservationSection` and `InterventionSection` should get the `PageBreakMixin` treatment, and probably also each task and session.

If the subtitles like "Sesion 1, task 1" could also get a decent font size bump that would also be a great improvement.

proposals/views/proposal_views.py
```diff
@@ -481,7 +481,8 @@ def form_valid(self, form):
         # Checks for practice proposals and starting the right
         # kind of review happen over there.
         proposal = form.instance
-        start_review(proposal)
+        if 'save_back' not in self.request.POST and 'js-redirect-submit' not in self.request.POST:
```
miggol, line 484: Ah yes, here's what I forgot.

miggol (changes requested):
Final small things:

* In `task_confirm_delete.html`, the cancel button is missing the "button" CSS class. This must have already been missing.

main/templates/base/form_buttons.html
```diff
@@ -3,35 +3,27 @@
...
-                <a class="button js-submit-redirect"
-                   href="{% url 'tasks:start' session.pk %}">{% trans "Terug naar begin sessie" %}</a>
+            {% if task %}
+                {% trans 'Maak een nieuwe taak aan >>' as create_task_text %}
+                <input class="button button-colored continue-button"
+                       type="submit"
+                       name="create_new_task"
+                       value="{{ next_text|default:create_task_text }}" />
             {% endif %}
         {% endif %}
```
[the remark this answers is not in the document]
tymees: > The create task/session buttons are currently scoped within the {% if not no_back %} if statement. That doesn't cause any problems right now but I don't think it's right.

Yeah, good spot. It indeed does not cause any problems right now (if ever), but it's is kinda weird conceptually...

> Maybe it's always been that way but it can be fixed with an else clause and empty div:

Nope, I introduced this bug in this PR. I've pushed a fix for it, which also moves the new buttons to the right scope. (As it made the fix easier and more readable as well).

miggol:
Firstly, great job on this large PR. This was looking like quite the mountain of an issue but you've scaled it rather well.

When it comes to the flow this is of course a huge improvement, but we're not all the way there yet. Like Ty, I also think the Session overview and Study end pages should have create buttons for sessions and even tasks within them. But I would go even further and say that the StudyEnd/SessionEnd page should become a kind of "home base" that the submitter keeps returning to.

Currently I have the following issues with the flow:

* It's hard to create a new session or task in general
* It's easy to get lost inside the loop. One moment you're finishing up your intervention, and the next you're creating a new session
* The new task/session actions are hidden deep within the loop, making it hard to figure out how far done you are as a submitter and which task/session you might want to create next
* Lots of messy logic scattered throughout various `get_next_url()` functions to determine where to go next
    * For example, we keep stepping through every task with the next button, which is jarring if all you wanted to do is edit one detail 
* Annoyances like not being able to delete a task or session if that would leave you with none

I think using the overview pages as home bases would solve most of this:

* **Finished creating a session => Session overview**
From where you can create your first task

* **Finished describing a task => Session overview**
From where you can add another task

* **Completely finished describing a session => Session overview**
From where you can add another session, or optionally continue

I think this approach would simplify things for us and the users both.

Right now I'm leaving up in the air if the home base should just be StudyEnd which exposes all functionality in one go, or if SessionEnd should function as an intermediate home base. I think that everything could work on one page if we give StudyEnd a redesign. But maybe an intermediate stepping stone (SessionEnd) could be more userfriendly. My main argument is looping around an overview page rather than going on an adventure everytime you hit Next Step.

EdoStorm96:
Ok, so here is a new version of this PR. It did indeed get a bit unwieldy again, as Ty warned me ... But I think it is looking nice!

So this is what I demo'd last week, with the changes requested during that demo. I've also incorporated all feedback from this PR. I am still waiting for an improved text for `session_start.html` from Desiree, but that can also be done later. Now, this text is mostly the same as it was, so this is fine for now, imo.

Most commits are pretty manageable, except for [f40ceca](https://github.com/DH-IT-Portal-Development/ethics/pull/603/commits/f40cecabf703705618747d66af22502fd0a57b8f) ... I am sorry about that, things were a bit to intertwined. This was a large overhaul of the flow ... However, one thing to pay attention to in this commit, is that I added a template tag for creating a hacky unordered html list, which I added in `fetc_filters.py`: `create_unorderded_html_list()`. This was based on code used in `pdf_diff_logic.py`. So now we can officially close #18 right? ;p

**Sidenote:** It was at this point in writing this comment I realised I should probably replace the function in `pdf_diff_logic.py`, with the new function, which I did. This works fine. Upon testing, I found a bug in `create_context_diff()`, due to an if-statement, which should have been an if-not ... I fixed this too. See the last commit. It's ironic that my previous monster PR has come back to haunt this one ...

Otherwise, I think the commit flow, is kindoff doable ... Good luck with the review. I am looking forward to hearing your thoughts :)

tymees:
I haven't looked at the updates, but I wanted to react to two things in your update comment:

>  I am still waiting for an improved text for session_start.html from Desiree, but that can also be done later. 

I agree, if the merging of this PR is going to be held till that point. I'd like to avoid non-ready code changes to develop atm, as we're trying to create a release.... (Also, there's something to be said for leaving this change for a follow-up release, as the currently staged changes are already extensive)

> Otherwise, I think the commit flow, is kindoff doable ... Good luck with the review.

Hahahahahahahahahahahahahahaha

miggol (changes requested):
I'm very happy with the flow as you've defined it now. Though I can't exactly remember what changed since last time. I was messing around with the design to visualize what I think could be better, have a look at my PR into this branch: #622 

What I think still needs work is the final overview page. It would be nice if interventions and observations followed the same design language. It's a bit out of scope for this PR perhaps, but I would rather get this right in one go.

I don't know why we use a RedirectActionView for creating new tasks and sessions. Is this related to the previous setup? Is it still necessary? I think it would be nicer if these were just CreateViews, only creating the object once you hit save. Now you get an empty task or session even if you misclick and hit back immediately. Perhaps @tymees knows why it's a RedirectActionView.

Back buttons do not always work or make sense. When creating a new task, the back button now goes to the general session info. I think it should go to the overview. It could also be a javascript browser back.

The back button on SessionEnd 404's.

Also, finally, I think it should "Taakonderzoek en **/of** interviews" and not just "Taakonderzoek en Interviews".

EdoStorm96:
> What I think still needs work is the final overview page. It would be nice if interventions and observations followed the same design language. It's a bit out of scope for this PR perhaps, but I would rather get this right in one go.

Mmmm, Yes I agree. I'll look into it.

> I don't know why we use a RedirectActionView for creating new tasks and sessions. Is this related to the previous setup? Is it still necessary? I think it would be nicer if these were just CreateViews, only creating the object once you hit save.

Now that I'm thinking about it, I guess this should be possible? I'll look into it as well.

> Back buttons do not always work or make sense. When creating a new task, the back button now goes to the general session info. I think it should go to the overview. It could also be changed to a javascript browser back.

Here I just found it a bit weird to make forward and back redirect to the same page ... but why not. It was my initial intuition as well.

> The back button on SessionEnd 404's.
woops. On it.

EdoStorm96:
@miggol I've implemented all your requests. Have a look at them when you have time :) Ty told me he would not have time to review this, this week. But this is ofc in no hurry.

tasks/templates/tasks/session_list.html
```diff
@@ -0,0 +1,44 @@
...
+                   class="text-decoration-none">
+                    <img src="{% static 'proposals/images/delete.png' %}"
+                         title="{% trans 'Sessie verwijderen' %}">
+                </a>
+            {% endif %}
+        </h3>
+        <p>
+            <em>
+                {% blocktrans with duration=session.tasks_duration repeats=session.repeats trimmed %}
+                    Totale brutoduur: {{ duration }}
```
miggol, line 26: Right now this will show "None" instead "Onbekend" when no duration is yet known. 

One way to solve this is to write a two-line template tag that takes a variable, and just replaces a None value with a translatable "Onbekend".

miggol (changes requested):
Found lots of stuff to complain about, feels like we're getting close though!

-----

I'm not satisfied with the back buttons yet.

Heading back from SessionEnd brings you to SessionStart, when I think it should really bring you back to the sessions overview.

Heading back from SessionUpdate brings you all the way back to the sessions overview, rather than just SessionEnd. Although I can see that making sense if you're only just creating a new session, it feels very wrong when you're just updating a session's details. Overall I think going back to SessionEnd from SessionUpdate is a less bad option.

(though if we used CreateViews here instead of RedirectActionViews we could just have different back buttons for either situation)

-----

There's also a strange bug with the menu bar, where some menu items remain unavailable:

image

When clicking them, you can't go there, but something in the redirect-submit code sets a `?next=%23#` URL parameter, after which redirect-submit is broken and both the forward and back buttons won't work.

It's kind of a hilarious bug, the user is suddenly absolutely trapped. I don't have time right now to figure out why it happens exactly, but let me know if you get stuck and we can look at it together. Might be a complicated/weird/interesting bug.

EdoStorm96:
@miggol I've incorporated your requests. Since we are already using createviews, your suggestion about the back buttons, was easy to implement.

The bug came down to the session_url and task_url methods in utils.py. There were some checks for completeness, which now, imo, are obsolete. Removing this fixed the bug.

Have a look at the commit history. :)

tasks/templates/tasks/session_overview.html
```diff
@@ -0,0 +1,29 @@
...
+    <div class="uu-inner-container">
+        <div class="col-12">
+            {% with nav_items=study.proposal.available_urls active=3 %}
+                {% include 'base/navigation.html' %}
+            {% endwith %}
+            <h2>{% trans "Overzicht van alle sessies en taken" %}</h2>
+            <p>
+                <em>{% trans "Vanuit dit overzicht is het mogelijk om sessies aan te maken, te bewerken of te verwijderen." %}</em>
+            </p>
+            <div class="session_list">{% include "tasks/session_list.html" %}</div>
```
miggol, line 20: When there aren't any sessions yet, this page is awkwardly empty. I think there should at least be some filler text indicating what to do next. Or a large div that just says "No sessions yet, time to create one!"

This empty space on first visit also kind of invites the SessionStart intro text to be put there. But I'm not 100% on that idea. Perhaps we can ask Desiree for her opinion.

If we put the intro text in an `{% empty %}`, it of course would no longer be visible once a session was created, which is a downside.

EdoStorm96:
So @miggol I've implemented your requests. The validation now works well. Check it out :)

miggol (changes requested):
I think the SessionOverview form is a good place for the extra validation and error processing.

### Flow related

I like the new blue infobox on SessionOverview. But it's so friendly that I expect that the next button would bring me along. That is to say, bring me along to the session creation page when I haven't created a session yet.

### Errors and validation

#### StudyEnd

The blue infobox is very friendly, but I think it could be a little less friendly once you arrive at StudyEnd. If you've still not created a session at that point, something's gone wrong and the infobox should probably resemble more of an error.

Right it just looks like an innocuous message you can just scroll past and hit "next".

#### No tasks

Like with sessions, we need an error when no tasks have been created. In the same way "SessionOverviewForm" is used to create the error for Sessions, SessionEndForm could be where the existence of at least one task can be checked, for the purpose of showing an error on ProposalSubmit.

Besides an error before proposal submission, SessionOverview could also use a "no tasks created" error box.

image

I'm thinking of where the tasks table is, so somewhere in tasks_list.html? As an `{% else %}` to `can_edit_tasks` perhaps?

### Reflections

As much as I like soft validation, I must admit that it doesn't exactly shine here. Not having made any sessions or tasks is such an obvious fault condition that I'm starting to doubt if we should let users continue from that state.

Perhaps the situation will be improved by more obvious errors. We'll see.

PR #604, opened 2024-02-01
fix: made sure fixtures are loaded in the correct order

tymees:
For complicated reasons I had to create yet another clone of the project (counter is on 4 now lol). I thought, let's use the new fancy load-fixtures command!

Only to discover that it had some issues with load-order. It tried to load registrationkinds before registrations, but registrationkinds depends on registrations. So, this PR addresses that issue:

1. It sorts the list of discovered fixtures, as `os.listdir` returns the list in an arbitrary ordering
2. Adds a number to the filenames in `tasks`, to ensure one is loaded before the other

The other fixtures loaded fine without the extra filename ordering, so I left them as-is


EdoStorm96 (approved):
Sounds good to me!

miggol (approved):

Issue #596, opened 2024-01-11
Tweaks to Committee Workload Overview


EdoStorm96:
Some requests from @djhcapel:

- Ensure that the committee workload overview is not visible to Chairs and PO's
> * Please adjust format date from month-day-year to day-month-year
> 
> * Please make the default period three months instead of one year

djhcapel:
- It would be appropriate if chairs could see the workload, as well as the secretary.
- Perhaps the text could then read: "Start datum periode (mm-dd-jj):" to help me remind?
- It would be really helpful if the default period could be set to three months, instead of one year.

EdoStorm96:
Ok, I will get to work on this.

PR #605, opened 2024-02-01
Fix/minor fixes

### LINKED ISSUE #592: Practice portal - Textual change (new regulations....)
Just noticed that at https://fetc.hum.uu.nl/proposals/start/practice/ it still says: 

"The regulations of the [General Chamber](https://fetc-gw.wp.hum.uu.nl/en/regulations-general-chamber-2/) or those of the [Linguistics Chamber](https://fetc-gw.wp.hum.uu.nl/en/regulations-linguistics-chamber/)."

This should read:
"The regulations of the [FEtC-H](https://fetc-gw.wp.hum.uu.nl/en/regulations-fetc-h/).

EdoStorm96:
This PR fixes #592, #596 and #597. The first two are just minor textual changes, so not much to see there.

#596 was a bit more work. There did not appear to be a datepicker widget, which was a significant improvement over what we had, so, despite trying some other things, we ended up sticking with the one we had ... Desiree said that she would be using it less, if the initial period would be set to 3 months.

Because of the requests she mad about access to this page, I ended up adding another group (`GROUP_PO = "Privacy officer"`) ... It is a bit annoying that this is only really used for regulating access to this specific page, but maybe we will have some other use for it in the future. I wanted to streamline fetc_filter.py to accomodate for the many possible combinations of groups ... I made a 'base' function `is_in_groups(current_user, groups)` which can be used to create filters for specific groups.

Otherwise, this is pretty straightforward.


tymees (approved):
Jup, looks good.

Nice refactor of the permission checks btw :)

PR #609, opened 2024-02-14
Give privileged groups archive access regardless of Humanities affiliation

miggol:
Currently it's kind of weird that a secretary or chamber member can't view the archive if they're not in the humanities faculty. This PR resolves this by allowing privileged groups in.

Notes:

* I initially changed `FacultyRequiredMixin` to use `UserPassesTestMixin` to be able to chain it with the new mixin. But once I had written the utility function for the menus, I realized this was ultimately unnecessary. Because I do believe `UserPassesTestMixin` offers more flexibility, I left this change in. But I'm happy to revert it, there's an argument to be made for both methods.


tymees (changes requested):
Aside from one small oversight, looks good. If you address my one comment it's approved, no re-review needed.

EdoStorm96 (approved):
Looks good! approve :)

PR #613, opened 2024-02-20
Redirect supervisor to the right update page

tymees:
During initial triage of the issue, I saw that actually fixing the issue was easier than to write it. So here we go.

This PR makes sure supervisors get the right update link on their 'decide page'. Previously, it would always redirect to the general one. 

(I also moved the link-logic to python to keep the template sane).


EdoStorm96 (approved):
Looks good. I approve, given the current state of the code. I am wondering though if it wouldn't be nicer to have a separate view for supervisor decisions and reviewer decisions. I was initially a bit confused looking at this template again. I suppose this would be a bit of work to do and its also beyond the scope of this issue.

Issue #315, opened 2022-03-22
Document cron command


tymees:
Let's put this somewhere in the docs, and not in a to-be-closed issue we'll never find again:
``0 7 * * * /<snip>/data/env/bin/python /<snip>/data/source/manage.py send_reminders >/dev/null 2>&1``

PR #614, opened 2024-02-28
text: initial mail-rewording pass

tymees:
I cannot wait till we have the new mailing code. Keeping multiple versions in sync is a pain

### DIFF
--- a/reviews/templates/mail/concept_creator.txt
+++ b/reviews/templates/mail/concept_creator.txt
@@ -3,10 +3,12 @@
 {% block content_nl %}
 Beste collega,
 
-Uw concept-aanmelding wordt door de eindverantwoordelijk onderzoeker {{ supervisor }} van deze aanvraag gecontroleerd.
-Zodra deze de concept-aanmelding, evt. na kleine aanpassingen, heeft goedgekeurd en ter toetsing bij de FETC-GW heeft ingediend, krijgt u van ons bericht.
-Indien de concept-aanmelding nog aanzienlijke wijzigingen uwerzijds behoeft zal de eindverantwoordelijk onderzoeker zelf contact met u opnemen.
-U kunt uw ingediende aanmelding hier in PDF vorm bekijken:
+Je conceptaanvraag wordt door de eindverantwoordelijk onderzoeker {{ supervisor }} gecontroleerd.
+Zodra {{ supervisor }} de conceptaanvraag, evt. na kleine aanpassingen, heeft goedgekeurd en ter toetsing bij de FETC-GW heeft ingediend, krijg je van ons bericht.
+
+Als er nog aanzienlijke wijzigingen in de conceptaanvraag moeten worden doorgevoerd zal {{ supervisor }} zelf contact met je opnemen. Je aanvraag wordt dan voorlopig afgekeurd in de portal, zodat je de wijzigingen kunt doorvoeren.
+
+U kunt uw ingediende aanmelding hier in PDF-vorm bekijken:
 {{ pdf_url }}
 {% endblock %}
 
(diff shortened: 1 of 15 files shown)


reviews/templates/mail/submitted_longroute_other_applicants.txt
```diff
@@ -4,18 +4,20 @@
 {% if was_revised %}
 Beste collega,
 
-Uw collega {{ creator }} (cc) heeft een gereviseerde aanvraag ingediend voor het onderzoek {{ title }}, waar u aan meewerkt, bij de ethische commissie. De aanvraag is door de FETC-GW ontvangen.
+Je collega {{ creator }} (cc) heeft een gereviseerde aanvraag ingediend bij de ethische commissie (FETC-GW) voor het onderzoek {{ title }}, waar u aan meewerkt. De aanvraag is door de FETC-GW ontvangen.
```
tymees, line 7: Forgot one

reviews/templates/mail/concept_creator.txt
```diff
@@ -15,7 +17,8 @@ Dear colleague,
 
 Your draft application will be checked by your supervisor {{ supervisor }}, the researcher who has the final responsibility of this study.
 You will receive a notification as soon as {{ supervisor }} has approved and formally submitted your draft application, possibly after making some minor revisions.
-In case {{ supervisor }} finds it necessary for you to make more substantial revisions in the application, he/she will notify you about this, and meanwhile formally disapprove the study in order for you to be able to revise and resubmit it.
+
+In case {{ supervisor }} finds it necessary for you to make more substantial revisions in the application, he/she will notify you about this, and meanwhile formally disapprove the application in order for you to be able to revise and resubmit it.
```
tymees, line 21: Paging @miggol 

Desiree wants your feedback on the 'in order for you to be able to revise and resubmit it.' part. Word thinks it's wrong, but doesn't really tell why...

miggol: I wouldn't call anything here truly incorrect but it is quite hard to read with all these verbs and subclauses.

I would just end with `... in order for you to revise and resubmit it.` which is also less tautological.

Up front you can similarly simplify the sentence by cutting out more verbs, resulting in something like this:

`In case {{ supervisor }} finds more substantial revisions to the application necessary, they will notify you about this, and meanwhile formally disapprove the application in order for you to revise and resubmit it.`

Which at least for me significantly reduces the amount of brainpower required to comprehend the sentence.

reviews/utils/review_utils.py
```diff
@@ -264,6 +268,53 @@ def remind_reviewers():
...
+    for decision in decisions:
+        proposal = decision.review.proposal
+        subject = "Herinnering: beoordeel aanvraag {}".format(
+            proposal.committee_prefixed_refnum(),
+        )
+        params = {
+            "supervisor": decision.reviewer.get_full_name(),
+            "creator": proposal.created_by.get_full_name(),
+            "proposal_url": settings.BASE_URL
+            + reverse("reviews:decide", args=(decision.pk,)),
```
tymees, line 294: So this var isn't used in the template, as Desiree's text didn't contain it. It's here because it was in the other reminder mail code I copied, so I left it in for the inevitable 'why isn't there a link' comments. 

Feel free to nudge me to add it to the e-mail, but I'm tired atm :P

EdoStorm96 (approved):
I could not find any faults. Respect for doing this haha. :sweat_smile: I suppose it would indeed be best to add a link to `reminder_supervisor` already, if you can find the energy.

How do cron jobs work though? Where do these reside for ethics?

tymees:
> How do cron jobs work though? Where do these reside for ethics?

Cron jobs are set on the server OS; Under the hood, it's a program that runs in the background, that just fires off 'shell commands' depending on the configured interval. We configure it to run a django management command ('send-reminders').

As described in #315, we run this:
``0 7 * * * /<snip>/data/env/bin/python /<snip>/data/source/manage.py send_reminders >/dev/null 2>&1``

This means 'on minute 0, hour 7, any day, any month and any year, execute the following command'.

PR #615, opened 2024-02-28
fix: fix broken tests due to invalid fixture names

tymees:
My bad; should've caught this when I made this change. 

Edit:
I also cleaned up the test-runner output with the other two commits:
 
The first removes the 'font-weight' spam. Turns out our PDF generator has a pretty strange CSS parser that does not allow `font-weight: 400` (as 400 is the default) in the font definition. Any other CSS parser thinks this is fine, if redundant, but well... 

The 'Not saving PDF' warning spam is disabled by disabling all log levels below WARNING (inclusive) for the proposal-submit tests only.

### DIFF
--- a/proposals/templates/proposals/proposal_pdf.html
+++ b/proposals/templates/proposals/proposal_pdf.html
@@ -11,7 +11,6 @@
         @font-face {
             font-family: "Open Sans";
             src: url('proposals/static/proposals/pdf_font/OpenSans-Regular.ttf');
-            font-weight: 400;
         }
 
         @font-face {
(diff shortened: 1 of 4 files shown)


EdoStorm96 (commented):
Wow, fixing the tests output is super blessed. Nice one!

I approve of the bugfix, however I'd prefer renaming the fixtures to `observation_registrations`, `task_registrations` and `task_registrationkinds` or something.

tymees:
That would also involve debugging the load-order again (as it's dependent on filenames for the load command, not tests).

That's a bit more work than just some simple renames. Would you be fine with a separate issue for that?

EdoStorm96:
Ahh, ok. Yeah, it's not such a big deal. I'd be fine with that :)

Issue #617, opened 2024-03-01
Regenerate PDF when secretary changes document


tymees:
Currently we only do manual regenerations. This was not know by the secretary, resulting in a ton of outdated PDFs. There's a call in topdesk to regenerate those that still can be regenerated (M240300564)

However, we should add a regenerate-step to the change documents page ASAP to avoid this for the near future.

Long term, we should change the way we link to documents from the PDF to not directly link to the files themselves, but to a 'document resolves' page that never changes (even if the underlying document does).

miggol:
> Long term, we should change the way we link to documents from the PDF to not directly link to the files themselves, but to a 'document resolves' page that never changes (even if the underlying document does).

Absolutely.

Though I thought that as long as a proposal is being reviewed the PDF is always regenerated? `use_canonical_pdf` should always be false when changes can still be made. Though I'm saying that without looking at the code.

tymees:
I didn't look too closely at the code, so that might be the _intended_ behavior. But I can 100% confirm that that is not _always_ the case. (I had to manually regenerate a couple of them).

So, if that's indeed the case we have to debug that as well.

tymees:
So, I think I know what happened.

The canonical-PDF-protection kicks in as soon as `status_review` is not none. Which is indeed what Michael described. 

So, I think this is a order-of-user-operations problem. The code expects the following:

1. All decisions are made
2. Secretary updates any documents
3. Secretary closes review <-- PDF is frozen
4. Secretary sends confirmation letter

However, I think Desiree _actually_ does the following:

1. All decisions are made
2. Secretary closes the review <-- PDF is frozen
3. Secretary updates any documents
4. Secretary sends confirmation letter

Which, tbh, from a workflow point-of-view isn't actually all that strange. So, if this indeed the case, I kinda think we don't need to do any further changes. My hotfix allows the secretary to update documents at any time, which solves the problem in all cases. It's somewhat dangerous tho, as the protection is there for a reason. 

So, alternatively we could move the protection to step 4, confirmation letter sending. We do have a field for that in the DB, `date_confirmed`, so we could modify the check to include a `date_confirmed is not None`.

tymees:
Eh, no, that alternative also does not work. There's no confirmation letter when `status_review == False`, but that is still a valid 'lock PDF' condition. So, `(status_review == True and date_confirmed is not None) or status_review == False`?

tymees:
[...]

PR #618, opened 2024-03-04
HOTFIX: allow secretary-induced changes to update the canonical PDF

tymees:
**This PR targets acceptation, in order to quickly deploy this fix**

This is a band-aid fix for #617, which only ensures the PDF is regenerated if the secretary updates one of the uploaded documents.

There is still the question why there were problems with this, but having taken a quick look at the code I think I know. I'll update that issue with my theory. (Which will still warrant this change btw).


EdoStorm96 (approved):
Yup, looks good :)

miggol (approved):
All good!

PR #619, opened 2024-03-04
Merge hotfix to master

tymees:


tymees:
Merged without review as permission was gained over Teams previously

PR #620, opened 2024-03-05
fix: prevent submission on back actions

tymees:
Cherry-pick from #603 

This PR re-introduces a check that prevents proposal submission by clicking the back button. This was accidentally removed during a refactor of the start review logic. 

As this problem is blocking, it must be merged before #593 is.


EdoStorm96 (approved):
all good!

miggol (approved):

PR #625, opened 2024-03-13
Feature/revise study end questions

### LINKED ISSUE #546: Extra question re knowledge security and safety researchers
Could we add an extra question under the question about ethical reflection?

Zijn er kwesties rondom kennisveiligheid en/of de veiligheid van de onderzoeker(s)? ja/nee, zo ja, licht toe.

EdoStorm96:
Fixes #546. Mostly textual changes, but also added some new fields on study_end. These changes were made according to Desiree's requests which she wrote up in a e-mail. A lot of text, but nothing too crazy functionality-wise.


studies/models.py
```diff
@@ -315,74 +352,41 @@ class Study(models.Model):
...
-                "Dit zou bijvoorbeeld het geval kunnen zijn \
-bij een 'onmenselijk' lange en uitputtende taak, een zeer confronterende \
-vragenlijst, of voortdurend vernietigende feedback, maar ook bij een ervaren \
-inbreuk op de privacy, of een ander ervaren gebrek aan respect. \
-Let op, het gaat bij deze vraag om de door de deelnemer ervaren belasting \
-tijdens het onderzoek, niet om de opgelopen psychische of fysieke schade \
-door het onderzoek."
-            )
+            "Bevat bovenstaand onderzoekstraject elementen die <em>tijdens\
+            </em> de deelname zodanig belastend zijn dat deze vragen, \
```
miggol, line 356: I don't know if this is a conscious change, or something that Black took liberties with. But this new whitespace is actually significant.

A line with a trailing `\` is continued in a very low-level manner, without the sanity checks Python usually provides. The interpreter literally sees the next line as a the same line and that leading indentation whitespace gets included in the string. That's why these msgstrs have loads of whitespace in django.po now. 

Somehow gettext doesn't mind (they're not fuzzy or anything) but this looks pretty awful:
image

The recommended way to continue strings is to just do it implicitly:

```python

    negativity = models.CharField(
        _(
            "Bevat bovenstaand onderzoekstraject elementen die <em>tijdens"
            "</em> de deelname zodanig belastend zijn dat deze vragen, "
            "weerstand, of zelfs verontwaardiging zouden kunnen oproepen, "
            "bijvoorbeeld bij collega-onderzoekers, bij de deelnemers zelf, "
            "of bij ouders of andere vertegenwoordigers?"
        ),
        max_length=1,
        choices=YesNoDoubt.choices,
        blank=True,
    )
```

Just keep adding strings without commas inbetween and python will implicitly join them. Or use triple quotes, `str.join()`, or literally anything other than backslashes, lol.

A lot of the strings here were already continued with backslashes from ages ago, which might not have been such bad practice back then. I wouldn't ask you to fix all of them just try to keep the strings you change free of unnecessary whitespace.

reviews/utils/review_utils.py
```diff
@@ -482,19 +482,36 @@ def auto_review(proposal: Proposal):
...
-als geheel zodanig belastend is dat deze ondanks de verkregen informed consent vragen zou kunnen oproepen."
+                    "De onderzoeker geeft aan dat er mogelijk kwesties zijn rondom \
+                    kennisveiligheid."
+                )
+            )
+
+        if study.researcher_risk in [YesNoDoubt.YES, YesNoDoubt.DOUBT]:
+            reasons.append(
+                _(
+                    "De onderzoeker geeft aan dat er mogelijk kwesties zijn \
```
miggol, line 496: If we end up keeping these fields in the study model, it might be nice to add the trajectory number in this string.

f"De onderzoeker geeft aan dat er mogelijk kwesties zijn rondom de veiligheid van de betrokken onderzoekers in traject {{study.order}}."


Right now, if a proposal has two trajectories and both have trajectory risk, this reason will just show up twice identically.

Also please fix the backslash whitespace issue here as well.

miggol (changes requested):
I have a couple things to be addressed.

EdoStorm96:
So, I pretty much implemented your requests, Michael. However, I've discussed with Ty that this PR is a bit redundant right now, with the big makeover coming up. I just wanted to finish this implementation, as documentation, but there is no real need to review it as we are most likely going to redesign this from scratch when changing to DSC 3.

PR #626, opened 2024-03-18
Fix/blocking issues on acc

EdoStorm96:
Some minor bugfixes on acc, before merging to main. These points are all in seperate commits.

- Firstly, I fixed the redirect, when clicking 'terug naar begin aanvraag', when making a pre-approved proposal. This previously redirected to the regular update page.

- I reworked the buttons on proposal_confirmation.html, so the back button would be on the left and changed the past tense text on the button to present tense. This seemed a bit more intuitive to me.

- I inverted the logic on the diff page by changing an if-statement to an if-not-statement, as that was currently incorrect.

- I also fixed the buttons on the task and session delete pages. As they were looking a bit ugly.

Otherwise, I didn't find any issues and think this version is ready for deployment.

### DIFF
--- a/proposals/utils/pdf_diff_logic.py
+++ b/proposals/utils/pdf_diff_logic.py
@@ -1103,7 +1103,7 @@ def create_context_diff(context, old_proposal, new_proposal):
             DiffSection(WMOSection(old_proposal.wmo), WMOSection(new_proposal.wmo))
         )
 
-        if new_proposal.is_pre_assessment:
+        if not new_proposal.is_pre_assessment:
             if (
                 new_proposal.wmo.status != new_proposal.wmo.WMOStatuses.NO_WMO
                 or old_proposal.wmo.status != old_proposal.wmo.WMOStatuses.NO_WMO
--- a/tasks/templates/tasks/session_confirm_delete.html
+++ b/tasks/templates/tasks/session_confirm_delete.html
@@ -18,8 +18,7 @@
                 <input class="button btn-danger"
                        type="submit"
                        value="{% trans 'Verwijderen' %}" />
-                <a class="button"
-                   href="javascript:history.go(-1)">{% trans "Annuleren" %}</a>
+                <a class="button" href="javascript:history.go(-1)">{% trans "Annuleren" %}</a>
             </form>
         </div>
     </div>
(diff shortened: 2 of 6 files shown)


proposals/utils/pdf_diff_logic.py
```diff
@@ -1103,7 +1103,7 @@ def create_context_diff(context, old_proposal, new_proposal):
             DiffSection(WMOSection(old_proposal.wmo), WMOSection(new_proposal.wmo))
         )
 
-        if new_proposal.is_pre_assessment:
+        if not new_proposal.is_pre_assessment:
```
miggol, line 1106: Man this nested if is really uninviting. Might be worth tidying up at some point.

Still surprised we didn't catch this though. Wouldn't this have meant that _all_ diffs wouldn't render any trajectories previously?

miggol (approved):
Nice, just one minor issue.

Good thing you tested preassessments and such. I didn't find any new issues on acc either.

tasks/templates/tasks/session_confirm_delete.html
```diff
@@ -15,11 +15,10 @@ <h2>{% trans "Sessie verwijderen" %}</h2>
                 <p>
                     Weet u zeker dat u deze sessie <strong>{{ session.order }}</strong> in de aanvraag <em>{{ session.study.proposal.title }}</em> wilt verwijderen?
                 </p>
-                <input class="pure-button pure-button-primary"
+                <input class="button btn-danger"
```
tymees, line 18: Wait, this works? Huh, how....

tymees (approved):
Conditional approval on that you change the point Michael raised. Otherwise, looks good.

(Yes, I'm doing reviews while on conference. I'm taking a break, with a nice view of the Eiffel Tower ;) )

tasks/templates/tasks/session_confirm_delete.html
EdoStorm96: ehm ... It turns red when hovering, which I could not explain, but kindoff liked ...

proposals/utils/pdf_diff_logic.py
EdoStorm96: Yes, the nesting is kinda crazy. It works best to look at the logic of the pdf, to make any sense of this. (Also black made this less readable imho ...)

I know! Bit crazy, but yeah, basically it never rendered anything.

PR #629, opened 2024-03-21
Hotfix/to conclude review api view qs bug

### LINKED ISSUE #628: "Applications waiting for conclusion" shows supervisor reviews
This is a result of us now closing supervisor reviews when a decision has been made. 

Previously we avoided supervisor reviews getting mixed in with the real deal by never closing them. Now we markg committee reviews explicitly with `Review.is_committee_review`, but this has not been reflected in `ToConcludeReviewApiView`.

We should go over the other API views to see if they don't have the same issue.

Review querysets checked:

- [x] MyDecisionsApiView
- [x] MyOpenDecisions
- [x] OpenDecisions
- [x] OpenSupervisorDecisions
- [x] InRevision
- [x] AllOpenReviews
- [x] AllReviews

EdoStorm96:
Fixes #628. The is_committee_review check was not properly implemented yet, which led to some stray supervisor reviews, appearing in certain API views. See the issue for more info. This should fix the problems.


reviews/api/views.py
tymees, line 1: While not an issue, `OpenSupervisorDecisionApiView`'s QS is still written with the old behaviour in mind. In this specific case, it works out. Still, probably a refactor candidate at some point.

Again, no fix is needed for this hotfix

tymees (changes requested):
Changes look good. One small comment, but that is not an actual issue AFAIK.

However, the Request Changes is related to the branch. Please go through the `acceptation` branch first. You do not have to do an acceptation deploy (alhtough, you may if you want), this is purely a git-thing. 

(We've had... Issues with backmerging master into acc and then into dev).

reviews/api/views.py
miggol: Oh yeah, good spot

Issue #608, opened 2024-02-12
Page Data management - link to workshop broken


djhcapel:
The workshop Handling Personal Data does not exist anymore. The link in the following text on the Data management page therefore does not work anymore: "the workshop [Handling personal data in research](https://www.uu.nl/en/node/81799)"

Please replace following text under "Privacy: AVG and GDPR":

old:
Recommended workshops
If you have not already done so, it is highly recommended you attend:

the workshop [Handling personal data in research](https://www.uu.nl/en/node/81799)

new: 
Recommended:
If you have not already done so, it is highly recommended you check:
- the e-learning module [Privacy basics for reserachers](https://www.uu.nl/en/research/research-data-management/workshops/privacy-basics-online-training)
- the [Data Privacy Handbook](https://utrechtuniversity.github.io/dataprivacyproject/handbook)

Issue #641, opened 2024-03-28
Practice portal - not only students 


djhcapel:
At the moment, a practice application can only be done as 'student'. You then also have to add a supervisor. It should be possible to select any other position (esp. annoying because now you have to add a supervisor....) as is the case in regular applications.

Issue #624, opened 2024-03-12
Date review round finalized and date 'sent for revision'


djhcapel:
On the details page, it now says in the left menu: (e.g.) Reviewronde beëindigd op 26 februari 2024, 13:23. 
As far as I understand, this is the date and time of submission of the comments of the last FETC-GW reviewer, not the date the application was sent for revision, so it's a bit confusing. For the secretary, it would be helpful to know when the status of the application was changed to 'in revision', i.e. the application was sent for revision.

Suggestion:

Laatste beoordeling ontvangen op [date, time] _(instead of 'Reviewronde beëindigd op')_
Afhandeling: Revisie noodzakelijk [date, time] _(thus: add date and time of status change)_

Would that be possible?

EdoStorm96:
Note to self: for the latter, we can use `proposal.date_reviewed`

PR #643, opened 2024-04-02
Fix/bugfix and textual changes

### LINKED ISSUE #634: "Details on the review process of application"
Could this phrase be deleted from the 'details' page? It is now confusing where the title of the application starts. If this phrase needs to be included on this page, perhaps better as a heading. However, I do not mind if it is deleted altogether.

EdoStorm96:
Fixes #634, #641 and #642. 

#642 and #643 are textual changes. #641 was a bug caused by an incorrect refactor, when changing the choices to model choices ... The labels for the two different options in `PracticeReasons` were switched around ... my bad ;p


EdoStorm96:
I have also just included a fix for #624. This just adds `proposal.date_reviewed` to the "Continuation:" section of review_detail_sidebar.html.

tymees (approved):
I only looked at the diff in GH, so no real full review, as the changes are minor.

LGTM :)

(Sidenote, you forgot to tag #624 in your description, this PR fixes that one as well I think).

EdoStorm96:
Decided to just push on more fix to this branch, as it was also a textual change. this fixes #608. Now I will leave this PR alone, haha. Could you have one more look, @tymees? :innocent:

tymees:
Ehhh, I'm not great at reviewing these textual changes. I'm kinda word-blind remember?

The only check I would do is seeing if the translations apply correctly, but I trust you to have tested that ;)

If you really want me to make a comment, there's some junk gathering at the end of the `.po` file (due to outdated translations being moved there automatically). It's a pet-peeve of mine, but not a problem ;)

PR #646, opened 2024-04-02
Update dependencies in develop

miggol:
The branch has "backmerge" in its name, but it turns out `acc` and `dev` are already in sync so no backmerge is necessary.


tymees:
As discussed in Teams, I added the sentry-sdk to the requirements. Ended up being a rather small change in dependencies, luckily. 

For now, this is the only change needed to integrate sentry, as all other stuff need to be done in `settings.py`. (And we don't ship server versions of it in the repo atm).

Issue #638, opened 2024-03-26
Members: editing of comments


djhcapel:
In the LC, some of us remembered being able to edit comments. This does not seem possible anymore. One of us inadvertently pressed 'Enter' halfway a sentence, and could not undo that anymore. Could an editing option be re-installed? Or incorporated? Perhaps with a tick that a comment is ready?

miggol:
Short term because this used to be possible, so if we can easily reinstate that feature then great.

However, if it turns out that this feature was removed because of some kind of complexity or bug, this can be moved to the "long term" milestone.

tymees:
The problem was introduced here:
https://github.com/DH-IT-Portal-Development/ethics/commit/c7e2fedc53fdec51c7a5b30be9b737094f2d7641

The first change was restoring old behaviour, it always locked the option after all comments where in (I checked against the v2 tag). The second change locks a decision after a go/no-go is set, which was not in the pre-review-actions code.

miggol:
Well found, Ty.

@djhcapel We'll restore the old behaviour for now: anyone can edit their feedback *unless* they were the final reviewer and the review enters the "To conclude" stage.

PR #648, opened 2024-04-02
Feature: allow decision editing

miggol:
This addresses #638 

Besides restoring the editing link, reviewers now get a warning if they are making the final decision that will cause the review to enter the `CLOSING` stage. For in that case, they won't be able to edit their response after the fact.

Note: this branch started in acceptation, but I think it's more suited to go through develop. Therefore, this has some extra commits to be backmerged.


reviews/templates/reviews/decision_form.html
```diff
@@ -24,6 +24,13 @@ <h2>{% trans "Aanvraag beoordelen" %}</h2>
                             De aanvraag is <a href="{{ pdf_url }}" target="_blank">hier</a> in te zien (downloadt als PDF).
                         {% endblocktrans %}
                     </p>
+                    {% if decision.is_final_decision %}
+                        <p>
+                            {% blocktrans trimmed %}
+                                <strong>Attentie:</strong> jij bent de laatste beoordelaar van deze aanvraag. Daarom kun je na het versturen je oordeel en commentaar niet meer bewerken. Controleer je antwoorden goed.
+                            {% endblocktrans %}
```
tymees, line 31: You might want to consider using an `alert` box instead. To draw more attention to it.

locale/en/LC_MESSAGES/django.po
```diff
@@ -3731,10 +3738,14 @@ msgid "Commissieleden aanstellen"
 msgstr "Appoint committee members"
 
 #: reviews/templates/reviews/review_assign_form.html:31
-#, python-format
+#, fuzzy, python-format
+#| msgid ""
+#| "<p> Kies hier de geschikte route en commissieleden voor de aanvraag "
+#| "<em>%(title)s</em>. klik <a href=\"%(workload_url)s\">hier</a> voor een "
+#| "overzicht van de werkverdeling van deze commissie. </p>"
```
tymees, line 3745: Fuzzy! Please check if you need to resolve this and remove the fuzzy bit if not

tymees (changes requested):
I've only reviewed your commits, and ignored the to-backport commits. 

Looks good, simple but effective. 

The only two commits are very minor, so when you've looked at those you can consider it approved.

reviews/templates/reviews/decision_form.html
miggol: An alert does look better, good idea

locale/en/LC_MESSAGES/django.po
```diff
@@ -3743,7 +3746,10 @@ msgid ""
 "<p> Kies hier de geschikte route en commissieleden voor de aanvraag "
```
EdoStorm96, line 3746: I think you need to remove `fuzzy, ` here (line 3744), so that this will compile correctly.

EdoStorm96 (changes requested):
Nicely done overall. I have one minor comment and one other suggestion. I think it would be nice if the text for the decide action changes, depending on whether you've already taken a decision. So in `DecideAction.description()` maybe you an add something like:
```
if decision.go == "":
    return _("geef jouw beslissing en/of commentaar door")
return _("Pas jouw beslissing en/of commentaar aan")
```

locale/en/LC_MESSAGES/django.po
miggol: Ah yes, I was getting used to `po-mode` in emacs which apparently doesn't do that automatically. Thanks

Issue #652, opened 2024-04-09
Datamanagement en Privacy - link to online training


djhcapel:
The data manager told me this link is useful: https://www.uu.nl/en/research/research-data-management/training-workshops/online-training-learn-to-write-your-dmp

On the Datamanagement en Privacy page, please adapt the following text (in bold new info, but should not appear in bold....):
NL:
"Als je dit nog niet gedaan hebt, wordt er sterk aangeraden om de volgende workshop **/ on-line training** te volgen:

- de workshop [Quick start to Research Data Management](https://www.uu.nl/en/node/75859)
**- de on-line training Write your DMP(https://www.uu.nl/en/research/research-data-management/training-workshops/online-training-learn-to-write-your-dmp)**

EN:
"If you have not already done so, it is highly recommended you attend:

- the workshop [Quick start to Research Data Management](https://www.uu.nl/en/node/75859)
**- the on-line training Write your DMP(https://www.uu.nl/en/research/research-data-management/training-workshops/online-training-learn-to-write-your-dmp)**

PR #653, opened 2024-04-09
feat: add help text to Documents.informed_consent

### LINKED ISSUE #650: Forms in case of public interest - textual change
Since for 'public interest', no consent forms are required, people now get confused by:

"Upload hier de toestemmingsverklaring (in .pdf of .doc(x)-formaat):"

Could we add: "(Bij algemeen belang, dien hier s.v.p. een leeg Word document in."?

EdoStorm96:
fixes #650. A small textual change Desiree would quite like to see published. Does require a migration, but nothing to crazy. I opted to place this as a help text, which I think is nicer design wise, but maybe less intuitive? Let me know what you think :)


EdoStorm96:
I've also added added a fix for #652 here. This is also just a small textual change.

proposals/templates/proposals/proposal_data_management.html
```diff
@@ -46,16 +46,22 @@ <h2>{% trans "Data Management Plan" %}</h2>
...
+                    {% blocktrans trimmed %}
                         de workshop <a href="https://www.uu.nl/en/node/75859" target="_blank">Quick
                     start to Research Data Management</a>
-                </li>
-            {% endblocktrans %}
+                {% endblocktrans %}
+            </li>
+            <li>
+                {% blocktrans trimmed %}
+                    De online training <a href="https://www.uu.nl/en/research/research-data-management/training-workshops/online-training-learn-to-write-your-dmp"
```
miggol, line 61: Inconsistent capitalization

studies/models.py
```diff
@@ -515,6 +515,7 @@ class Documents(models.Model):
 
     informed_consent = models.FileField(
         _("Upload hier de toestemmingsverklaring (in .pdf of .doc(x)-formaat)"),
+        help_text=_("Bij algemeen belang, dien hier s.v.p. een leeg Word document in."),
```
miggol, line 518: The English version is definitely not informative enough. I would argue the Dutch version could be more elaborate as well.

I would suggest: "When your lawful grounds for processing data is the Public Interest, for which no explicit consent is required, you may submit a blank document in this field."

And then something similar in Dutch.

miggol (changes requested):
Small changes requested

Issue #611, opened 2024-02-19
Some users are not flagged properly as humanities


tymees:
We've had two topdesk tickets (M240215147 and M240204594) indicating the portal does not think they work at GW. 

Not entirely sure why, in both cases their Solis ID is properly flagged as hum-affiliated. Maybe the IdP just didn't send that info?

In both cases the user in question did not have any faculty assigned, so maybe we should change the text of that specific case to be less definite?

tymees:
So the 'we don't know any faculty' text is definitly too harsh, and should be more honest about the situation.

Proposal new texts:

Faculty known: (unchanged)
> Deze portal is bedoeld voor medewerkers (en studenten) van de Faculteit Geesteswetenschappen. Volgens onze gegevens werk/studeer je bij de {X}. Controleer of je inderdaad een aanvraag wilt indienen bij de Facultaire Ethische Toetsingscommissie van Geesteswetenschappen.  Als je vermoedt dat dit incorrect is, neem dan contact op met portal support.

Faculty unknown:
> Wij konden niet verifieren bij welke faculteit je in dienst bent. Deze portal is bedoeld voor medewerkers (en studenten) van de Faculteit Geesteswetenschappen. Controleer of je inderdaad een aanvraag wilt indienen bij de Facultaire Ethische Toetsingscommissie van Geesteswetenschappen. Indien dat het geval is kan de portal gewoon gebruiken. Voor vragen kan je contact opnemen met  portal support.

@djhcapel Is this okay?

djhcapel:
I did not understand what 'dit' referred to in the Faculty known text? Suggestion (if my understanding is correct, but why would they need to contact you?):

Deze portal is bedoeld voor medewerkers (en studenten) van de Faculteit Geesteswetenschappen. Volgens onze gegevens werk/studeer je bij de {X}. Controleer of je inderdaad een aanvraag wilt indienen bij de Facultaire Ethische Toetsingscommissie van Geesteswetenschappen. Als je niet verbonden bent aan {X}, neem dan contact op met portal support.

tymees:
'Dit' indeed refers to 'we think you belong to x'. So your new wording would be clearer.

As to why they would need to contact us; well, it's mainly just to provide a 'complaints' contact. But, it might happen that someone is employed by multiple faculties, and we simply do not know atm what that would do. (No real-life testcase for it).

But my main concern is the 'faculty unknown' text. Is that text okay?

[...]

PR #659, opened 2024-04-18
Log errors during faculty processing

tymees:
**Note: this targets acceptation to hasten deployment**

I took a look at how prevalent #611 is... it's bad. Over 85% of new users this year do not have a faculty attached.

Before I start accusing ITS, I want to see if we're at fault. So, this PR adds a log call to the except we've been ignoring. This should show up in Sentry.

### DIFF
--- a/main/models.py
+++ b/main/models.py
@@ -163,6 +163,6 @@ class SamlUserProxy(User):
                     continue
 
                 faculty_obj.users.add(self)
-            except Exception as e:
-                logger.error(f"Error processing faculty for user: {e}", exc_info=True)
+            except:  # noQA
+                logger.error(f"Error processing faculty for user", exc_info=True)
                 continue
--- a/main/models.py
+++ b/main/models.py
@@ -7,6 +7,7 @@ from django.utils.translation import ugettext_lazy as _
 
 logger = logging.getLogger("ethics.main")
 
+
 class YesNoDoubt(models.TextChoices):
     YES = "Y", _("ja")
     NO = "N", _("nee")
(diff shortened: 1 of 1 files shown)


miggol (approved):
Looks good

EdoStorm96 (approved):
👍

main/models.py
```diff
@@ -159,6 +163,6 @@ def process_faculties(self, faculties):
                     continue
 
                 faculty_obj.users.add(self)
-            except:
-                # Just ignore any errors...
+            except Exception as e:
+                logger.error(f"Error processing faculty for user: {e}", exc_info=True)
```
miggol, line 167: I kind of half-remember it being best practice to not format log strings actively, but to use the formatting provided by `logger.error()` itself. Like that the logging handlers and formatters can work their magic.

But I can't find a reference to that anywhere so it might not even be the case anymore. Not blocking issue either way.

XanderVertegaal: Just to add my 5 cents here (I don't know why I'm getting email notifications from this PR): Pylint was mad at me last week for using f-strings instead of the old format: https://stackoverflow.com/questions/34619790/pylint-message-logging-format-interpolation. Maybe you are thinking about that?

tymees: Yeah y'all are right that this isn't the ideal/correct way to do this. (One point not even mentioned is me formatting an exception into a string, which is definitely the worse offense here actually). 

The method I used (f-strings) was me being lazy. The reason (doing it at all) was because I wasn't sure if Sentry would correctly capture the whole exception context. (That's what `exc_info` does btw).

I just did a [local test](https://centre-for-digital-humanities.sentry.io/issues/5248791949/?query=is%3Aunresolved&referrer=issue-stream&stream_index=0), and it does! As this makes the string-format unnecessary, I removed it

PR #660, opened 2024-04-22
Initial DSC & Django update

tymees:
This PR lays the foundations to built V4 upon. 

Note: this PR targets the `major/4` branch for now. This is a fork of develop made at the same time as the source branch, to have a clear diff. New changes in develop will need to be merged into there as well. 

It's a rather large PR; I've tried to make proper commits for individual changes, to help reviewing. I've also added some comments over adding them here, such that the comment is located near the relevant code.
Also, please ignore the changes in the docs folder. Those are caused by my bulk-refactors.

Highlights:

- Update to Django 4.2
  - I think I got this all sorted, but there might still be some problems lurking in the later form pages
- Updated to DSC 3
  - New app-specific base template (and moved all pages to it)
  - New form-specific base template (and moved the first form + WMO forms to it)
  - New home-page design. (Needs sign-off from desiree)
  - Initial automatic CSS migration pass (renaming classes to their new equivalents mostly)
  - Ported form-JS utils to new form (non-table-based) layout 

Still todo:
Well, a lot. A non-exhaustive list:

- Layout polishing pass 
- All other form pages need to be converted to the new form-base-template
   - In addition, some pages might need to be split up to be less intimidating
- All other pages need to be (re-)reviewed. 
- A proper stepper implementation
- New overview pages using UU-List. (Ideally we'd cut down on the amount of those pages in favour of filters)
- Everything else in https://github.com/DH-IT-Portal-Development/ethics/milestone/3
- And probably stuff I'm forgetting


main/static/main/style.css
tymees, line 1: I've moved the original contents to `style.css.bak`. I did this because I want us to manually review anything we need from there. A lot of it is unneeded now, and other stuff might be better implemented using existing classes from BS.

main/templates/base/fetc_base.html
tymees, line 1: This is the new base template for ethics, to be used by all pages either directly or indirectly. It mostly implements the header styling, leaving the main content styling up to the DSC provided template. 

Note: the current header isn't quite finished, it still needs a polishing pass. However, that would delay this PR, and it's functional enough for now

main/templates/base/fetc_form_base.html
tymees, line 1: This is the more controversial one of the two new base templates. This template is designed to be used by most 'form'-templates (with 'form templates' I only mean the ones part of the actual application. We have other forms in the app, like 'copy', that should just the normal base template. (Altough, we could create a more generic form-template that handles for `FormView`-derivatives, that handles the HTML for forms.)

Note: some of our application-form-pages are a bit special, and might need a custom template instead of this 'automagic' template.

main/views.py
```diff
@@ -57,8 +57,58 @@ def get_queryset(self):
         return self.model.objects.filter(not_after__gt=now, not_before__lt=now)
 
 
-class HomeView(LoginRequiredMixin, _SystemMessageView):
+class SeasonalCoverImageMixin:
```
tymees, line 60: Okay so this addition is 100% unnecessary. I just wanted to have a bit of fun during a boring meeting. 

It adds a seasonal-appropiate cover-image automatically. 

The winter and autumn covers are the ones we already had, the spring and summer ones are new. 

Note 1: I made the old 'summer' version the 'autumn' one. Mostly because it had people in winter-coats in it, so it wasn't really summer. 
Note 2: I found out that our attributions we had before were wrong! We listed the person who added the image to the image-bank, not the actual photographer themselves. The ones listed here now are correct.

main/views.py
```diff
@@ -540,9 +626,9 @@ class DeleteView(LoginRequiredMixin, UserAllowedMixin, generic.DeleteView):
     """Generic delete view including login required and user allowed mixin and
     alternative for success message"""
 
-    def delete(self, request, *args, **kwargs):
+    def form_valid(self, form):
         messages.success(self.request, self.success_message)
-        return super(DeleteView, self).delete(request, *args, **kwargs)
+        return super(DeleteView, self).form_valid(form)
```
tymees, line 631: Django 4 changed how deleteviews worked. For those wondering where this and other similar changes comes from.

EdoStorm96 (approved):
Wow, very good work :) I mainly looked at what you changed, not so much at what stayed the same.  And i did some, but non-exhaustive testing. It all looks good and consistent. The one thing I think you overlooked is to remove `{% load uil_filters %}` from a few templates. This caused an error for me at one point in testing.

I also quite like the philosophy of the new fetc_form_base template. I think the places you've used the new boxes, show the potential of this nicely.

All in all, I'm happy and excited to continue working on this!

reviews/templates/reviews/committee_members_workload.html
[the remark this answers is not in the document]
tymees: Yeah, this is one of the files only touched by my automatic refactoring runs; and I didn't do a check for naked submit buttons. (There are probably more?)

And not really your fault; the previous CSS handled that automatically for you, the new one doesn't.

tymees:
Thanks for the review!

I've added a commit that refactors all uses of `uil_filters`. 

There are probably more of these 'well you could've done that in bulk' problems still living in other places; I've only done those I was aware of, and those I found during my work on the pages I gave individual attention. 

Normally we go with the 'one approval is enough' strategy, but I really want @miggol's review before merging, as it's quite an impactfull PR.

miggol:
### Landing page

image

* The middle-aligned navigation bar up top looks weird. Like it's crying for help. I think aligning it to the right/flex-end would help. I expect this would look find for most other pages with more nav items.

main/templates/base/fetc_form_base.html
miggol: Awesome work, and great image.

I think the `pre-form`, `form-buttons`, and `post-form` should be easily splittable into a their main content and an aside, as with the actual form.

This makes a more sense stylistically, for example to align the buttons to the main form body. This also allows for an aside to the pre-form intro text, for example with some "handige links" like we do in procreg.

Of course this is already possible if you code it yourself or reuse the classes. However, it would be nice to have it supported by default everywhere in the form by having separate `pre-form-content` and `pre-form-aside` blocks you could use right away.

And then a parent `pre-form` block which gives you the full width if you replace it.

miggol (approved):
Alright, I've been browsing for a while and overall I'm very impressed. Most things are simply working.

There's some stuff still missing but I'm sure we'll get to it soon enough in the django-4 branch.

main/templates/base/fetc_form_base.html
tymees: That's a great suggestion! As it's quite easy to implement, I've done so:
image
Some notes:
- As you suggested, these new blocks are childs of the existing `pre/post-form` blocks, so you can still override those as before. 
- I called `pre-form-content` `pre-form-text` instead, as that mirrors the class `.uu-form` provides for this usecase
- I didn't split up `form-buttons` into an aside, as I don't see a usecase where that makes sense to actually use. I did change the HTML to use the same `text` classes as the new blocks, such that the buttons are aligned with the form as you suggested.
- With this buttons-change, we probably want make those form buttons smaller. Currently I think it's a bit claustrophobic

main/templates/main/landing.html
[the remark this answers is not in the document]
tymees: Yeah honestly I think the menu should just be empty if you're not logged in, especially if we add the footer from index.html here as well. 

It's middle-aligned because procreg is middle-aligned, and it's a bit annoying to fix that manually for this one case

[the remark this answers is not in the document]
tymees: Good idea!

However, it's (sadly) not as simple as moving some HTML around. `landing.html` has some extra JS to protect against spambots scanning `mailto:` links, which we'd have to re-introduce. I'd rather create a separate issue for this, to implement that correctly.

PR #666, opened 2024-05-06
Base stepper PR

miggol:
Not ready for in-depth review, but feel free to comment on the general direction.

Structure of the stepper object will be similar to the previous AvailableURL object, with children as an attribute. However, the stepper items themselves will be more complex, with knowledge of errors or missing fields on the page they reference.

I don't want to create an "everything object", but considering the overlap in functionality I expect the stepper to at some point absorb `validate_proposal.py`.


proposals/utils/stepper.py
```diff
@@ -0,0 +1,99 @@
...
+    def get_context_data(self, *args, **kwargs):
+        context = super().get_context_data(*args, **kwargs)
+        # Try to determine proposal
+        if hasattr(self, "proposal"):
+            proposal = self.proposal
+        else:
+            proposal = Proposal.objects.get(pk=self.kwargs.get("pk"))
+        # Initialize and insert stepper object
+        context["stepper"] = Stepper(proposal)
+        return context
```
tymees, line 73: I think you know this as well, but this won't work for any page that does not operate on Proposal objects.

Maybe we could extract some code from [`UserAllowedMixin` ](https://github.com/DH-IT-Portal-Development/ethics/blob/8ac03c15561b1eb8068499d0eeeee0b4845ecb23/main/views.py#L339) and add it to a `ProposalMixin` (which, annoyingly, already exists, but that should be renamed)

proposals/utils/stepper.py
```diff
@@ -0,0 +1,99 @@
...
+    if lst == []:
+        return lst
+    first_item = lst[0]
+    rest = lst[1:]
+    if type(first_item) is list:
+        return flatten(first_item) + flatten(rest)
+    return [first_item] + flatten(rest)
+
+
+class renderable:
```
tymees, line 16:
```suggestion
class Renderable:
```

Altough, these generic classes should probably live in `main`?

proposals/utils/stepper.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def get_context_data(self):
+        return {}
+
+    def render(self, context={}):
+        template = loader.get_template(self.template_name)
+        context.update(self.get_context_data())
```
tymees, line 23: Shouldn't this be the other way around? 

Like, any provided context should override the context from `get_context_data`?

proposals/utils/stepper.py
tymees: Also, as an aside, it _might_ be an idea to create the stepper object in `Proposal`, as we may also want to use the data from it as part of the logic of the app. Then, we add proposal to the context instead and call `proposal.stepper` in the template.

This is just me brainstorming, the current setup is fine with me too

miggol: > I think you know this as well, but this won't work for any page that does not operate on Proposal objects. 

Yeah I know, it's early days. A proposal object is all a stepper should ever need to generate itself. Determining the proposal from any of the sub-objects is quite trivial, but I'm not sure whose responsibility it is.

I could grant the stepper the power to initialize from any of the sub-objects ("DWIM"). ALternatively I could make a method in StepperContextMixin called `determine_proposal_from_object()` that can figure it out. The latter could prove useful to have lying around.

We could also have every proposal view be separately responsible for setting `self.proposal` as a rule. But that feels brittle.

miggol:
Closed to create a fresh PR. Thanks for the feedback!

PR #667, opened 2024-05-27
Feature/multiple sessions tasks dsc 4

EdoStorm96:
So this is an initial merge of the multiple sessions and tasks branch with the DSC 4 update. I've merge major/4 into my old dev branch. It works fine, although there is basically no CSS. Might be a fine starting point though, especially to incorporate into the design of the stepper...


studies/templates/studies/study_end.html
miggol, line 1: We need to figure out styling for all the tables in the proposal form. 

Styling for tables in the proposal form can differ from what we use in the review detail tables and such. But preferably we'd have consistent tables everywhere within the proposal form.

tasks/templates/tasks/task_update.html
miggol, line 1: I don't know why radio buttons are so broken in these templates, but not in others. The DMP and submit pages for example have working radios, but anything in the studies category seems to be broken. Even the intervention/observation forms which haven't been touched by this PR.

tasks/templates/tasks/session_overview.html
miggol, line 1: I think we should bring back a slightly darker background for these session displays. Make each session a separate object visually.

miggol (changes requested):
I consider the `django.po` issue blocking. All other comments can be postponed and made into an issue for later, so that we can merge this PR.

Other stuff we still need to do:
* Handle errors in the `studies` forms, and choose which ones should be blocking and which are soft.
* Rework all form controls and page layouts #668 
* Re-implement Javascripts where desired, such as `study_start` trajectory title fields.

miggol:
Nice, approved!

miggol (approved):
Nice, approved!

PR #676, opened 2024-06-12
Clearer faculty warning messages

tymees:
This PR changes the faculty-warning texts to be more honest and nuanced. While the hotfix I deployed today _should_ make this less important, I still think this changer is needed. 

The faculty-known message variant is, mostly, the same. It adds a line stating that the portal is usable despite the warning.

The faculty-unknown message is now honest that we don't have that info, and also assures the portal is still usable. 

New faculty known message:
image

New faculty unkown message:
image

(Ignore the colors, I have darkreader on)


miggol (approved):
I approve. It's a lot clearer like this, and the different messages make it easier for us to tell what's going on based on user feedback.

The only change I might suggest is to not say "employed" but "associated with" to be more inclusive to students. With something like "aangesloten bent bij" in Dutch. But that's just a suggestion.

EdoStorm96 (approved):

miggol (approved):

PR #679, opened 2024-07-09
Feature/break up page 1

EdoStorm96:
Well, here it is: the long awaited break up of page 1. It took me a while to wrap my head around it ... I've had a meeting with the secretary about the desired new architecture and we landed on this. I think it make quite a lot of sense like this, breaking up the giant start page, actually allowed for a lot of simplification. Notably in the forms and the views. For instance, we now just have one `UpdateView`, instead of different versions for different proposal types, which is quite nice.

I think this is mostly self explanatory. There are some big commits here, but that was kindoff unavoidable.

Regarding 5b3b043, This template was not displaying pre form messages correctly. This solution works well for me, but I am open to other suggestions.

I am pretty happy with how this turned out, so I am eager to hear your thoughts :)

### DIFF
--- a/main/templates/base/fetc_form_base.html
+++ b/main/templates/base/fetc_form_base.html
@@ -3,7 +3,9 @@
 {% block content %}
     <div class="uu-container">
         {# A block one can use to insert content before the stepper/form container #}
-        {% block pre-form-container-content %}{% endblock %}
+        <div class="col-12 text-center">
+            {% block pre-form-container-content %}{% endblock %}
+        </div>
 
         {# todo: responsive design #}
         <div class="col-3">
--- a/main/templates/base/fetc_form_base.html
+++ b/main/templates/base/fetc_form_base.html
@@ -4,7 +4,8 @@
     <div class="uu-container">
         {# A block one can use to insert content before the stepper/form container #}
         {% include "proposals/practice_or_supervisor_warning.html" %}
-            {% block pre-form-container-content %}{% endblock %}
+        {% block pre-form-container-content %}{% endblock %}
+
         {# todo: responsive design #}
         <div class="col-3">
             {% block stepper %}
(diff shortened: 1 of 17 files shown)


main/templates/base/fetc_form_base.html
```diff
@@ -3,8 +3,10 @@
 {% block content %}
     <div class="uu-container">
         {# A block one can use to insert content before the stepper/form container #}
-        {% block pre-form-container-content %}{% endblock %}
+        <div class="col-12 text-center">
+            {% block pre-form-container-content %}{% endblock %}
 
+        </div>
```
tymees, line 9: I'm a bit unsure about this one. It does solve the problem you probably encountered, but it does introduce some limitations. 

In my intended usecase you would actually add this `col-12` div as part of the content you add in this content block. 

That way, you can choose to use a different col-layout (like a `col-3` and a `col-9` to mimick the form-layout below). 

In any case, the `text-center` class doesn't make sense here. Why would you want to force all content inside to have centered text? That's something you should decide per-usage

proposals/templates/proposals/funding_form.html
```diff
@@ -0,0 +1,26 @@
...
+        $(function () {
+            check_field_required('funding', 'needs_details', 'funding_details', 'proposals');
+            check_field_required('funding', 'needs_name', 'funding_name', 'proposals');
+        });
+    </script>
+{% endblock %}
+
+{% block pre-form-container-content %}
+    {% include "./practice_or_supervisor_warning.html" %}
+{% endblock %}
```
tymees, line 22: Is there any page we would _not_ want to show this warning?

If not, we probably should move this to the form base template

tymees (changes requested):
Oh this is sooooo much better! It also integrates the pre-approved/pre-assesment questions a lot more elegantly into the flow. 

The code also looks a lot cleaner now too :P

I did find 2 bugs, both related to `SearchableSelectWidget`, but those ended up being bugs in DSC; I've created issues for them.

I made two other comments, they're relatively minor.

tymees (approved):

PR #685, opened 2024-08-16
Fix/small django4 fixes

miggol:
This fixes two issues:

1. A selfexplanatory issue with reverse URLs in `validate_proposal.py`
2. Study names aren't inserted into `StudyStartForm`'s data, causing a KeyError

With regards to the second issue, I admit that I haven't looked deep into the mechanics causing this. I'm a bit apprehensive for messing with the behaviour of `SoftValidationMixin`, which is a bit of a beast that we rely on in many places. This is just what I cooked up quickly so that I could test the rest of #684, and I'm open to better solutions.


EdoStorm96 (approved):
Good catches!

PR #687, opened 2024-08-22
Fix/improve forms UI

EdoStorm96:
Here is a mishmash of different small issues. Mostly to do with improving error displays, and general UI improvements for the forms, mostly around sessions and tasks ...

I also found out the non-field-errors were not caught by validation, so I added a small fix for this.

### DIFF
--- a/studies/forms.py
+++ b/studies/forms.py
@@ -239,7 +239,7 @@ class StudyDesignForm(TemplatedModelForm):
         # error msg to a built-in required error message.
         if not "study_types" in cleaned_data:
             error = forms.ValidationError(
-                _("Je dient minstens één van de opties te selecteren."), code="required"
+                _("Je dient minstens een van de opties te selecteren."), code="required"
             )
             self.errors["study_types"] = error
 
--- a/studies/forms.py
+++ b/studies/forms.py
@@ -225,11 +225,11 @@ class StudyDesignForm(TemplatedModelForm):
         - at least one of the fields has to be checked
         """
         cleaned_data = super(StudyDesignForm, self).clean()
-        #This solution is a bit funky, but by using add_error(), it appends our
-        #error msg to a built-in required error message.
+        # This solution is a bit funky, but by using add_error(), it appends our
+        # error msg to a built-in required error message.
         if not cleaned_data:
             msg = _("Je dient minstens één van de opties te selecteren.")
-            self.errors["study_types"]=[msg]
+            self.errors["study_types"] = [msg]
 
 
 class StudyConsentForm(ConditionalModelForm):
--- a/studies/forms.py
+++ b/studies/forms.py
@@ -217,7 +217,11 @@ class StudyDesignForm(TemplatedModelForm):
 
     class Meta:
         model = Study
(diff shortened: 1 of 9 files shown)


miggol (changes requested):
Sorry this took so long, I keep ending up diving super deep into the code in these reviews and then don't finish them.

Nice improvements in general.

tasks/templates/tasks/task_list.html
```diff
@@ -48,7 +48,7 @@
     </table>
 {% else %}
     <div class="alert alert-info" role="alert">
-        {% trans "Deze sessie bevat nog geen taken. Klik op het podloodje om taken aan te maken!" %}
+        {% trans "Deze sessie bevat nog geen taken. Tijd om er één aan te maken!" %}
```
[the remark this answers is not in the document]
EdoStorm96: I was thinking we create a seperate issue for translation ... Because a lot of stuff is not translated or not up to date ...

EdoStorm96:
Ok, So I've adressed your comments, and I also found some bugs, which I fixed along the way. 

I noticed that StudyEndForm was missing from the validation, so I've added that. (I see now that I put the wrong commit message regarding this ... oh well.) -> 9899ca7

I found a bug in this form. It was, unnecessarily receiving a study kwarg, this is now fixed.

And I tried to work on StudyDesignForm, and made some changes that will hopefully make this form easier to validate externally. However, as I showed, I did run into a really strang bug, where this form does not get properly initialized in the validator. Its clean method does not get called, and I cannot figure out why this is. I would really appreciate if you could have a look at this, because it is driving me crazy!

studies/forms.py
```diff
@@ -217,17 +217,36 @@ class StudyDesignForm(TemplatedModelForm):
...
         cleaned_data = super(StudyDesignForm, self).clean()
-        if not cleaned_data:
-            msg = _("Je dient minstens één van de opties te selecteren")
-            self.add_error("study_types", forms.ValidationError(msg, code="required"))
+
+        # This solution is a bit funky, but by using add_error(), it appends our
+        # error msg to a built-in required error message.
+        if not "study_types" in cleaned_data:
+            error = forms.ValidationError(
+                _("Je dient minstens één van de opties te selecteren."), code="required"
```
miggol, line 242: I recently learned (guess from who) that you're not supposed to write "één" in Dutch if there's no chance of ambiguity. Same with  "Tijd om er één aan te maken!", by the book it shouldn't have accents.

studies/forms.py
```diff
@@ -285,7 +304,6 @@ def __init__(self, *args, **kwargs):
         - Remove empty label from deception/negativity/stressful/risk field and reset the choices
         - mark_safe the labels of negativity/stressful/risk
         """
-        self.study = kwargs.pop("study", None)
```
miggol, line 288: This View is still getting a `study` kwarg argument from somewhere, causing it to error out.

miggol (approved):
Approving, assuming that the study kwarg comment gets fixed. The orthography issue is of course minor

PR #688, opened 2024-08-28
Feature/review and secretary pages redesign

EdoStorm96:
So, here I've restyled a lot of the remaining (form) pages. Mostly relating to the review process and secretary side. This is mostly self explanatory, however, I did make some minor design changes: 
- I've renamed to stepper block on fetc_form_base to sidebar, as it can be used to accomodate other things, like the review detail sidebar.
- I've simplified the buttons block, to make overriding buttons a lot easier.
- I was running into bugs with review_actions, so I've updated the logic. See b702725. To me, it seems that this is how it should work. Please pay close attention here.

I am unsure if there are pages leftover that still need a redesign. Let me know if you can think of any.


reviews/utils/review_actions.py
```diff
@@ -183,10 +183,9 @@ def is_available(self):
         if not settings.GROUP_SECRETARY in user_groups:
             return False
 
-        if self.review.stage not in [
-            Review.Continuations.REVISION,
-            Review.Continuations.NO_GO,
+        if self.review.stage in [
             Review.Stages.CLOSING,
+            Review.Stages.CLOSED,
```
miggol, line 188: The blame history of this logic is a mess, lol. How did continuations get in here? This should have been caught by someone.

Well, it's been caught now at least. Good find.

In my opinion, this action should be available to the secretary in the following stages:

* Assignment (obviously)
* Commission
* Closing
    * This stage is automatically reached when all opinions have been gathered. But it happens often enough that the secretary wants to add another reviewer before making the final call, so this action should still be available in that stage.

As it stands now, this action will be available in the supervisor stage, which I don't think is right. Could result in weird behaviour if the secretary accidentally adds or removes someone in that stage.

miggol (changes requested):
Good progress on the redesign.

> *    I've renamed to stepper block on fetc_form_base to sidebar, as it can be used to accomodate other things, like the review detail sidebar.
> *    I've simplified the buttons block, to make overriding buttons a lot easier.

These are excellent changes

main/templates/base/fetc_form_base.html
```diff
@@ -31,37 +31,39 @@
...
-                {% block post-form-pre-buttons %}{% endblock %}
+                {% block auto-form-render %}{{ form }}{% endblock %}
 
-                <div class="uu-form-text-row">
-                    <div class="uu-form-text">
-                        <div class="mt-4 mb-3 w-100 d-flex">
-                            {% block form-buttons %}
-                                {% include "base/form_buttons.html" %}
-                            {% endblock %}
+                {% block form-buttons %}
```
miggol, line 36: I think something went wrong when merging. This template now has two `{% block form-buttons %}` (see below, line 42) which causes an error.

miggol (changes requested):
Just one template issue but otherwise all seems well

PR #689, opened 2024-08-28
Base stepper PR

miggol:
Alright, this is ready for review. I don't consider *everything* done but I want to move on to other things and allow the rest of the team to continue development with a working stepper.

It's a bit of a beast with some rough edges. I encourage asking for docstrings, comments, and clarification during review.

Things that still need doing:
* Disabled mode for stepper items, with correct classes and no link
* Error collection, which should be an easy addition as all stepper items that point to a `ModelForm` view have the instantiated form available
* Stepper bubble colour depending on errors and completeness
* Collecting all errors in the Stepper class for presentation in `ProposalSubmit`
* The Documents page, which will come with the Attachments update
* Collapsing of non-active stepper subfolders
* Stepper flow for non-standard Proposal types.

Thankfully, these are all smaller issues that are very doable by others now that the framework is there.


proposals/utils/checkers.py
```diff
@@ -0,0 +1,1016 @@
...
+
+from tasks.views import task_views, session_views
+from tasks.models import Task, Session
+
+from .stepper_helpers import RegularProposalLayout, PlaceholderItem, StepperItem
+
+
+class BaseStepperComponent:
+
+    def __init__(self, stepper, parent=None):
```
EdoStorm96, line 21: Maybe just make this stuff part of the Checker and get rid of BaseStepperComponent?

proposals/utils/checkers.py
```diff
@@ -0,0 +1,1016 @@
...
+        self,
+    ):
+        return reverse(
+            "proposals:study_start",
+            args=[self.proposal.pk],
+        )
+
+
+class TrajectoriesChecker(
+    Checker,
```
EdoStorm96, line 418: Couldn´t you just use a ModelFormChecker? And get rid of TrajectoriesItem?

proposals/utils/checkers.py
```diff
@@ -0,0 +1,1016 @@
...
+
+    def check(
+        self,
+    ):
+        self.stepper.items.append(
+            self.make_stepper_item(),
+        )
+        return []
+
+    def make_stepper_item(
```
EdoStorm96, line 801: Can we bury this somewhere deep? xp

EdoStorm96 (changes requested):
Ok, Nice job! Whew, what a beast! ;p I am starting to kindoff get it. It seems like it will be quite powerful for different apps in the future! 

For my review, I had a look through everything you altered and did some minor testing. It looks like everything is working as intended, therefore, I did not really go into the minutia of how stuff works (Like `SessionCheck.make_stepper_item()` ...) but had a pretty good look and tried to understand the design as best I could. In general I do find it a bit complicated, but I think going over these points would improve the whole thing quite a bit!

- TaskMixin (or somewhere) still needs a StepperContextMixin. Otherwise TaskCreate and stuff cause an error. (minor)
- A clearer separation of classes that are meant to be inhereted and classes that are meant to be instantiated. Maybe these can live in different files? Because I find it a bit messy if they are all in checkers.py. Maybe just move stuff like "Base" checkers and stepper items to the helpers file. Maybe have comments throughout this file to seperate different types of Base classes ... And group different types together.
- Also Docstrings for anything intended as a baseclass would be very nice!
- Lastly, I have a bit of an issue with the Layout class. I find it a bit unclear when the Stepper does things and when the Layout does things and I feel like these could be one big Stepper class? Maybe it would be to much, but now they are also very intertwined, which is indeed a bit spaghetti-ish.

See my other comments for minor things I found.

miggol:
Alright thanks for the valuable feedback!

I'm going to go through the smaller stuff first, then consider the Layout class and some of the other class hierarchies, and then document everything after.

main/templates/base/stepper.html
```diff
@@ -3,21 +3,17 @@
...
             <li>
-                <a class="stepper-item" href="{{ url.url }}">
-                    <span class="stepper-bubble stepper-bubble-largest">{% counter counter value %}</span>
-                    <span>{{ url.title }}</span>
+                <a class="stepper-item {{ item.css_classes }}" href="{{ item.get_url }}">
+                    <span class="stepper-bubble {{ bubble_size.0 }}">{% counter counter value %}</span>
+                    <span>{{ item.title }}</span>
                 </a>
-                {% if url.children %}
+                {% counter depth create 1 %}
```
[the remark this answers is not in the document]
miggol: It prevents a bit of ugliness caused by the empty `<ul />`.

image

Note the inconsistent item spacing and "tail" at the end

studies/views/study_views.py
```diff
@@ -23,12 +23,17 @@
...
 from ..utils import check_has_adults, check_necessity_required, get_study_progress
+from ..mixins import StudyMixin
 
 
 #######################
 # CRUD actions on Study
 #######################
-class StudyUpdate(AllowErrorsOnBackbuttonMixin, UpdateView):
+class StudyUpdate(
+    StudyMixin,
```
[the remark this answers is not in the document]
miggol: Do you mean doing `context["proposal"] = self.get_proposal()` or moving the setting of the proposal context variable into `StudyMixin?`?

Or something else?

studies/views/study_views.py
```diff
@@ -23,12 +23,17 @@
 )
 from ..models import Study, Documents
 from ..utils import check_has_adults, check_necessity_required, get_study_progress
+from ..mixins import StudyMixin
```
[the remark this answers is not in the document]
miggol: Because `StudyMixin` inherits from `StepperContextMixin`, which in turn gets imported in `studies/mixins.py`

proposals/utils/checkers.py
miggol: Ah yes, that's better. Would be needed for checking errors in the future, too. Don't know why I chose not to

proposals/utils/checkers.py
miggol: Hmm... I like `BaseStepperComponent` to keep track of what is shared between items and checkers for now.

miggol:
Okay, I think I've addressed everything. Ready for another review!

* Moved base classes into stepper_helpers.py with new docstrings and comments
* Merged the Layout and Stepper classes. Layout came into existence when I didn't yet know what direction the classes would take, but in the end it is kind of superfluous
* `renderable` Mixin now lives in `main/utils.py`
* `StepperItems` are now also renderables
* The Stepper has a more functional and less stateful `build_stepper()` method

EdoStorm96 (changes requested):
Alright, some really good improvement! Nice ideas and good catches. I do have some more comments.

- I am having a strange problem where clicking around in the stepper, in a new proposal or older proposal, keeps bringing me to an old proposal ... A few pk's back. Somewhere along the route, the stepper is receiving the wrong proposal object, and redirecting to it. I think this has something to do with ProposalMixin. When I get to the WMOForm and further in the form, the problem doesn´t occur anymore for me.
- I am not a huge fan of proposal type hint and suggest we get rid of it. It seems like one more place to keep track of this. I would suggest the following:
	- During construction on major/4, the stepper would not appear until after you filled in the first form. I did not mind this design as it gives you kind off a zen start. Maybe we can bring this back? This would then, maybe also eliminate the need for the proposal_type_hint, as we could then just examine the proposal, and provide the correct layout based on that.

We can of course discuss this proposed solution :) See you tomorrow and enjoy the rest of your birthday.

proposals/utils/stepper.py
```diff
@@ -0,0 +1,157 @@
...
+                return item.get_url()
+
+    def build_stepper(
+        self,
+    ):
+        """
+        The meat and potatoes of the stepper. Returns a list of top-level
+        StepperItems to be rendered in the template.
+        """
+        layout = getattr(self, "layout", False)
```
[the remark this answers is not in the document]
miggol: I don't think so, because in almost all cases the Stepper can determine its type (and thereby layout) all by itself. The createviews are the only exception I can imagine. So if the layout was an argument it definitely should not be required.

Second, the concept of a "layout" is completely internal to the stepper implementation. I don't think every View that instantiates a stepper should need to know what a layout is and when to choose which one. The createviews know what *type of proposal* they're creating, because they already have to, and that should be good enough.

I think this is important because there may not exist a 1:1 mapping between proposal types and layouts. A practice proposal might use the same layout as a regular proposal, but require a different checker flow because it asks for a practice reason. Because a checker can do both of those things with just the proposal type, it makes sense to me to put that in a `ProposalTypeChecker` which then also determines the layout.

Adding to the above, although it isn't necessarily relevant to your comment, is that I think that the state of the stepper object should only be modified by Checkers. Once an object is complicated enough to warrant this kind of crazy depth-first search operation with Checkers, I want to respect that and not also change object state in various object methods.

Otherwise, I can imagine myself stepping through the checkers for ages trying to debug something, only to find that it was in fact that one thing I decided to put in a Stepper method instead.

miggol:
I was a little hasty pushing the last fixes.

* Imports were still in the wrong place
* Incorrect references
* Super weird stuff going on caused by #[4b8b130](https://github.com/DH-IT-Portal-Development/ethics/pull/689/commits/4b8b130e5e1e5097b3fd2e6c3739299a26fd8013)

But I think everything is fixed now. Thanks for finding the above.

> I am having a strange problem where clicking around in the stepper, in a new proposal or older proposal, keeps bringing me to an old proposal ... A few pk's back. Somewhere along the route, the stepper is receiving the wrong proposal object, and redirecting to it. I think this has something to do with ProposalMixin. When I get to the WMOForm and further in the form, the problem doesn´t occur anymore for me.

I *think* this was caused by #[4b8b130](https://github.com/DH-IT-Portal-Development/ethics/pull/689/commits/4b8b130e5e1e5097b3fd2e6c3739299a26fd8013), but please double check if it's still happening.

> * I am not a huge fan of proposal type hint and suggest we get rid of it. It seems like one more place to keep track of this. I would suggest the following:
>     *    During construction on major/4, the stepper would not appear until after you filled in the first form. I did not mind this design as it gives you kind off a zen start. Maybe we can bring this back? This would then, maybe also eliminate the need for the proposal_type_hint, as we could then just examine the proposal, and provide the correct layout based on that.

Although I posted a lengthy motivation above about why it currently is the way it is, I think removing the stepper from the createviews altogether is a totally cool idea that would save us some complexity. The reason I haven't done this is that it would be extra work and debugging *right now* (just before leaving for the UK), whereas the current implementation at least works.

And if it turns out there's something we hadn't thought about with the alternate proposal types that *does* require some kind of proposal type hint, well at least it's already there.

Feel free to not include the stepper in the createviews if that saves you some time, should you decide to work on the alternate proposal flows.

EdoStorm96 (approved):
Nice that you found the fix! It looks good for now, I´d say :) Nice work!

Issue #690, opened 2024-09-03
Change logic for DMPFileSection in pdf_diff_logic


EdoStorm96:
Currently the logic is:
```
if model.dmp_file:
    sections.append(DMPFileSection(model))
```
But this section should still appear if there is no DMP file. It should still show the checkbox.

PR #698, opened 2024-09-11
Feature/footer

EdoStorm96:
A little footer, based on what we had. I don´t know if we want any other text or features from this? It seems fine like this imo.

Also went ahead and removed WMOCheck. One more issue ;p


EdoStorm96:
I've also just added a fix for #690 to this PR. Very minor change.

miggol (approved):
Nice! I like that this now stretches the page nicely in the vertical direction. All seems well here.

Issue #691, opened 2024-09-03
StudyDesignForm does not get clean()'ed during validation.


EdoStorm96:
For some mysterious reason StudyDesignForm's `clean()` does not get called during instantiation in `get_form_errors()`. Other forms do not have this issue. This form's `full_clean()` does get called during validation for some reason.

It is very unlikely that a user tries to submit with errors on this form (They would have to try to have a study with no intervention, no observation and no sessions), so it is not the end of the world in practice ... But it is very annoying.

Maybe this goes away when we switch to a new method of validation, but maybe not ...

Issue #692, opened 2024-09-04
Add "disabled" class to Stepper items


miggol:
https://utrechtuniversity.github.io/UU-Bootstrap/components/stepper/

Currently disabled items just have an empty link. They should instead have no link at all (no `<a />` tags) and the correct CSS class.

PR #700, opened 2024-09-12
Feature/stepper refinement

### LINKED ISSUE #693: Collect errors from stepper for Proposal validation
It's (hopefully) time to retire `validate_proposal.py`!

The `stepper.items` list already contains all the modelform items that a proposal requires to be completed. Each has an `item.errors` property that will allow you to access the form errors.

Implementing this may require slightly more work than just glueing everything together. There could be quirks in `validate_proposal.py` and errors that get generated elsewhere than in modelform instantiation. Specifically I hope that #691 doesn't rear its head.

### LINKED ISSUE #694: More stepper styling
https://utrechtuniversity.github.io/UU-Bootstrap/components/stepper/

In the reference and procreg steppers, there are green and yellow stepper bubble classes that have not yet been implemented here in Ethics.  

This should not be hard, and can be implemented once in the base stepper item class. Though #693 should probably be completed first so the errors implementation is complete.

### LINKED ISSUE #699: Add stepper bubble for wmo application
Currently, the only wmo bubble directs to WMO update/create, but you need to use the next buttons to get to WmoApplication. I think we need to pay some attention to this view anyway, because I think we often kindoff glance over it.

EdoStorm96:
So I went a bit ham ... I think I've made some nice improvements to the stepper. It seems to work well, but I'm curious to hear your thoughts, @miggol.

I am now a lot more intimate with the new stepper, and also more appreciative of its utility. Kudos again :)

So, what I've done:
 - Firstly, I did get rid of the stepper for the first form page. This allowed me to simplify quite a bit, because the stepper is always working with a proposal of which the type is known. This is chill imo.
 - I've implemented the flows for the different proposal type (PreAss and PreAppr, as I refer to them.)
- I've messed a lot with the WMOChecker ...:
  - Firstly I noticed that the WmoApplicationForm (different from WMOUpdate and such) was not implemented in the stepper, so I've incorporated that. (fixes #699)
  - I've also done some work on the Wmo views and forms themselves, as I found some incorrect urls ...
  - I also figured I could make the WmoChecker into a ModelFormChecker, to also get rid of the WmoItem. This was quite the hassle ... I initially wanted to use an CreateOrUpdateChecker, but this did not quite work, so I took inspiration and made it into its own design. This seems to work fine now. However, I do think we need to look into the `clean()`'s of WMO related forms as these seem kindoff incomplete ...
- So then the big one, I've overhauled the validation. (fixes #693) This was pretty simple luckily! A few bumps on the road:
  - many forms need form kwargs, so the ModelFormChecker now has a `get_form_kwargs()` method.
  - Then the other big thing: as individual sessions and tasks are not incorporated into the stepper (which I still think is good) these needed their own validation function. This is somewhat akin to the old validation system, but quite a lot simpler. I don't hate it.
  - To my great frustration, #691 still persists. I don't know 😭  ....
- Lastly I've made some simple implementations to get the correct css classes (incomplete, complete and disabled) This works ... but I feel like there might still be some bugs ... I think having this, mostly just really showcases inconsistencies/messiness in form's individual validation ... So it would be good to have a look at that ... (fixes #694, #693)

Sorry for the big PR ... One thing led to another and such. But overall I think this has improved the stepper quite a bit!


proposals/utils/validate_proposal.py
```diff
@@ -1,322 +1,82 @@
...
-            else:
-                forms[observation_key] = (
-                    ObservationForm,
-                    reverse("observations:create", args=[study.pk]),
-                    _("Het observatieonderzoek (traject {})").format(
-                        study.order,
-                    ),
-                    None,
-                )
+def validate_sessions_tasks(study):
```
miggol, line 18: I would really really prefer if we could somehow have this be part of the stepper/checker infrastructure, so that all we had to do was run a `stepper.get_all_errors` to get list of errors with titles and URLs.

For example, calling this code from the `SessionsChecker`. Or otherwise a `ModelFormItem` for each of the task pages that is hidden but does collect errors.

That would keep everything in one place and allow us to turn the Sessions item yellow if its subpages had errors.

miggol (changes requested):
You got a lot done with this PR! We still need to double check that all stepper items actually add errors, for example the "Financing" modelform, but that is more of an issue with the forms.

Also, this smaller part of #692 is not addressed:

* Remove the `<a />` tags from disabled stepper items

EdoStorm96:
Ok, So I've adressed your comments ... I guess this is fine for now, although I don´t love some aspects of the current design ... Mainly the validation of sessions and tasks ;p A small overview of what I've done:

- ensure disabled items are no longer <a> elements
- ensure ContainerItems check all their children and represented errors thusly.
- Mostly removed ContainerItems from the stepper, so that all major steps represent modelforms and the modelform's errors. ContainerItems are now only used for studies, when there are multiple studies.
- This also led to some slight reordering. It used to be:
  - StudyEnd > TranslatedForms > Documents > DMP
But now it's:
  - StudyEnd > Documents > TranslatedForms > DMP
- Improved the working of validation and (mostly ...) moved it inside of the stepper. I have mixed feelings about this, as this also leads to a lot more imports on stepper.py. But maybe you have suggestions for improvements of Stepper.get_form_errors() ... I did ensure all the problems with validation you pointed out are fixed ... But this also made the design a bit more complicated.
- I've added a possibility of providing a get_checker_errors function to ModelFormItem's in case we need to provide some more specific validation. I've implemented this on SessionsChecker, as an example. 
- And we're (for now) still stuck with this custom validator function for sessions and tasks ... Which gets called everytime the stepper gets rendered, just to give the Sessions stepper item the right color ... :skull: 

IMO this is fine for now though. Curious to hear your thoughts :+1:

miggol (approved):
> Ok, So I've adressed your comments ... I guess this is fine for now, although I don´t love some aspects of the current design ... Mainly the validation of sessions and tasks ;p A small overview of what I've done:
> 
>     * ensure disabled items are no longer  elements
> 
>     * ensure ContainerItems check all their children and represented errors thusly.
> 
>     * Mostly removed ContainerItems from the stepper, so that all major steps represent modelforms and the modelform's errors. ContainerItems are now only used for studies, when there are multiple studies.
> 
>     * This also led to some slight reordering. It used to be:
>       
>       * StudyEnd > TranslatedForms > Documents > DMP
>         But now it's:
>       * StudyEnd > Documents > TranslatedForms > DMP

This all looks great!

>     * Improved the working of validation and (mostly ...) moved it inside of the stepper. I have mixed feelings about this, as this also leads to a lot more imports on stepper.py. But maybe you have suggestions for improvements of Stepper.get_form_errors() ... I did ensure all the problems with validation you pointed out are fixed ... But this also made the design a bit more complicated.
> 
>     * I've added a possibility of providing a get_checker_errors function to ModelFormItem's in case we need to provide some more specific validation. I've implemented this on SessionsChecker, as an example.
> 
>     * And we're (for now) still stuck with this custom validator function for sessions and tasks ... Which gets called everytime the stepper gets rendered, just to give the Sessions stepper item the right color ... 💀
> 

I can totally live with this for now. There's a perfectly good future in which we just include items for every task and session that are just not rendered in the actual stepper. So I'm not too concerned about this bit of uglyness, as long as it's called from within the stepper (as it is now) and easy enough to debug. Nice work

> 
> IMO this is fine for now though. Curious to hear your thoughts 👍

There is one somewhat significant issue that I still want fixed, which is that errors in proposal validation don't actually prevent submission in ProposalSubmit. But expecting that that will be quick fix I'm approving this so you can merge right after

PR #702, opened 2024-09-19
feat: make stepper bubbles collapse and refine css working

EdoStorm96:
So, I was just puzzling a bit with the stepper, and to my surprise, came up with a solution for making the bubbles collapse ...

This is still a bit rough, but I thought I'd share it with you before the (my) weekend. Will add more notes on monday!


proposals/utils/stepper.py
```diff
@@ -84,8 +87,17 @@ def build_stepper(
...
+        #own styling
         for item in self.items:
+            #Only check item.is_current until there is a current item found
+            while not self.current_item:
+                if item.is_current(self.request):
+                    self.current_item = item
+                    self.current_item_ancestor = item.get_ancestor()
+                else:
+                    break
+            item.get_css_classes()
```
miggol, line 100: What is the reason that we need to get the CSS classes at this stage? Can't we just call `item.get_css_classes()` from the template?

proposals/utils/stepper_helpers.py
```diff
@@ -129,16 +148,14 @@ def is_current(self, request):
...
 
-    def css_classes(self):
-        css_classes = super().css_classes()
+    def get_css_classes(self):
         child_errors = []
         for child in self.children:
             child_errors += child.get_errors()
         if child_errors != []:
-            css_classes += "incomplete"
+            self.css_classes += "incomplete"
```
miggol, line 156: There isn't enough consistency in where the space goes. I think we might better avoid the issue entirely and provide a `list` of css classes, and only turn them into a string when needed for HTML.

miggol (changes requested):
Although I'm impressed with how you've got this feature working with so few lines, I'm not happy with the implementation. There are two main reasons for this.

First, I believe the representation of the stepper produced by `build_stepper()` should be "correct" and complete. Right now, we are collapsing stepper items by completely ommitting children that do exist in `stepper.items`. This means that there are items with parents that don't have children. Also, the resulting stepper's shape is very different depending on which item is current. This makes it unnecessarily hard to reason and potentially debug the output of `build_stepper`

Of course, the css classes and such are also different depending on which item is current, but I consider that to be part of the presentation rather than the internal representation. Collapsing of stepper items should in my view also be part of the presentation, not the internal representation.

Second, but somewhat related, this implementation is not very flexible. This implementation only works for top-level items: collapsing children of a child would get messy very quickly. We might also want to move to JS-expandable stepper items in the future, but ommitting children entirely does not allow for that. 

My recommendation would be to keep the collapsing part of the stepper to the presentation side of things, in this case I would suggest moving it to the template:

```django
{% if not item.is_collapsed %}
    {% if item.children %}
        ... Render children ...
    {% endif %}
{% endif %}
```

The `is_collapsed` flag can easily be repurposed into a CSS class if we move to JS collapsing. Also, we can inspect the complete structure of the stepper from within a breakpoint without needing to worry about children being ommitted.

What I do think we should keep is the one-time search for a current item, that's a great optimization. I don't think it belongs in `build_stepper()` which I would rather keep clean if possible. I would recommend a separate loop, or at least a separate method that checks for currentness.

EdoStorm96:
So here is the new version of this PR, some notes on what has been changed:
- StepperItem's now have an `is_available attribute. This is false by default. This relates to whether its children are displayed. A more correct name would be `is_not_collapsed` ...
- The new way of searching for the current item is retained, but this now happens in a separate method (`item_is_current_check()`) of the stepper, until a current item is found.
- When a current item is found, its `get_ancestors()` gets called, which returns a list of the item + all of its ancestors (so parent's, parent.parent's etc) These items' `is_available` are then set to True
- The get_errors function and get_css_classed have had a bit of an overhaul. get_errors, can now get an `include_children` boolean argument, so that the errors of an item's descendants are also taken into account. 
- The current design is, when an item is collapsed, its completeness gets evaluated with `include_children`, when an item `is_availabe` just the model_form errors and the optional `get_checker_errors` get evaluated.

This feels right to me :)

proposals/utils/stepper_helpers.py
```diff
@@ -68,25 +69,34 @@ def get_context_data(self, *args, **kwargs):
...
     def is_current(self, request):
         """
         Returns True if this item represents the page the user
-        is currently on.
+        is currently on and adds the active class to the item's
+        self.css_classes set. This gets called when building the stepper,
+        until is_current returns True.
         """
         if request.path == self.get_url():
+            self.css_classes.add("active")
```
miggol, line 99: This method was previously stateless, being more of a "check" and not an action with consequences. If we add the CSS class here, we have to remember to do that for every time we write a custom `is_current()`, like in the Session item (which currently doesn't do this correctly).

I see how this makes sense, but I would prefer if the css class gets added in `is_current_check()` so we don't have so many different places to think about CSS classes.

proposals/utils/stepper.py
```diff
@@ -36,6 +36,8 @@ def __init__(
         # which item is current
         self.request = request
         self.items = []
+        self.current_item_ancestors = None
```
miggol, line 39: To keep consistent with types, this should be an empty list.

main/templates/base/stepper.html
```diff
@@ -6,17 +6,18 @@
...
                 {% else %}
-                    <a class="stepper-item {{ item.css_classes }}" href="{{ item.get_url }}">
+                    <a class="stepper-item {{ item.get_css_classes }}"
+                       href="{{ item.get_url }}">
                         <span class="stepper-bubble {{ bubble_size.0 }}">{% counter counter value %}</span>
                         <span>{{ item.title }}</span>
                     </a>
                 {% endif %}
-                {% if item.children %}
+                {% if item.children and item.is_available %}
```
miggol, line 20: I like this collapsing implementation a lot more! 

If in testing we decide, for example, that we want to leave the Trajectories items expanded by default, we can easily implement that. Very flexible.

miggol (changes requested):
> So here is the new version of this PR, some notes on what has been changed:
> 
>     * StepperItem's now have an `is_available attribute. This is false by default. This relates to whether its children are displayed. A more correct name would be `is_not_collapsed` ...

How about `is_expanded`? No biggie either way

>     * The new way of searching for the current item is retained, but this now happens in a separate method (`item_is_current_check()`) of the stepper, until a current item is found.
> 
>     * When a current item is found, its `get_ancestors()` gets called, which returns a list of the item + all of its ancestors (so parent's, parent.parent's etc) These items' `is_available` are then set to True

See my code comment, I think it would be better if the current check or `build_css_classes` would check if it's the current item and set the css class there, rather than in `is_current()`.

Otherwise this is great.

>     * The get_errors function and get_css_classed have had a bit of an overhaul. get_errors, can now get an `include_children` boolean argument, so that the errors of an item's descendants are also taken into account.
> 
>     * The current design is, when an item is collapsed, its completeness gets evaluated with `include_children`, when an item `is_availabe` just the model_form errors and the optional `get_checker_errors` get evaluated.
> 
> 
> This feels right to me :)

Perfect!

EdoStorm96:
So, I've adressed your comments. Good ideas!

I also noticed that FundingForm did not really include validation, so I just added some cuz it was bugging me ...

proposals/utils/stepper_helpers.py
```diff
@@ -251,24 +282,38 @@ def instantiate_form(self):
...
+            if self.get_errors(include_children=True):
+                self.css_classes.add("incomplete")
+            else:
+                self.css_classes.add("complete")
+        elif self.get_errors():
+            self.css_classes.add("incomplete")
         else:
-            css_classes += " complete"
-        return css_classes
+            self.css_classes.add("complete")
```
miggol, line 316: I also feel like this could be simplified now that we have the `is_expanded` variable.

Wouldn't something like `if self.get_errors(include_children=self.is_expanded)` remove the need for the double `if` statements?

I might be missing something though, just a quick suggestion.

miggol (approved):
Very nice! I have some small suggestions but functionally this works very well. Approving.

proposals/utils/stepper_helpers.py
```diff
@@ -251,24 +282,38 @@ def instantiate_form(self):
...
+        errors += self.model_form.errors
 
-    def css_classes(self):
-        css_classes = super().css_classes()
+        if hasattr(self, "get_checker_errors"):
+            errors += self.get_checker_errors()
+        if include_children and self.children:
+            for child in self.children:
+                errors += child.get_errors(include_children=True)
+        return errors
```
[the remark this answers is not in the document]
EdoStorm96: No, I don't think so because the super() just goes to StepperItem's get_errors which just returns an empty list. This is where the behaviour of this function gets specified. Up the chain, include children becomes basically meaningless.

Issue #706, opened 2024-09-25
Do the diff ...


EdoStorm96:
The diff page still needs to get updated ...

Issue #705, opened 2024-09-25
Translations


EdoStorm96:
So ... I've not really been keeping django.po up to date ... So we need to get all translations in order.

PR #707, opened 2024-09-25
Feature/revise study end questions 4

### LINKED ISSUE #546: Extra question re knowledge security and safety researchers

EdoStorm96:
fixes #546. This was an olde branch I had worked on. This implementation is not entirely what we settled on, as we initially wanted to also have a small table representing the trajectories on the Knowledge Security page if there were multiple trajectories. I do still have a template lying around for this somewhere ... But I don't think it is the most important thing now. At least we have the updated questions/texts in here now. 

As far as I can tell, everything here is working well. Though there are a lot of changes, nothing too crazy happening.


miggol (changes requested):
Lots of small remarks but nothing major, excellent turnaround

proposals/utils/pdf_diff_logic.py
```diff
@@ -856,6 +854,30 @@ def get_row_fields(self):
...
+        "researcher_risk",
+        "researcher_risk_details",
+    ]
+
+    def get_row_fields(self):
+        rows = copy(self.row_fields)
+        obj = self.obj
+
+        rows_to_remove = []
+        for x in range(0, len(self.row_fields), 2):
```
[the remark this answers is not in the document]
EdoStorm96: Yeah, I have kindoff a love/hate relationship with this ... I often start thinking of a way to get rid of it, but then end up spending too much time on it (again ...) and I just move on. I'll get around to it someday ;p

EdoStorm96:
So, I've addressed your comments. The translations were a lot of work, because a lot of stuff was behind (#705 ...). I promise to keep translations up-to-date from now on 😔.

Most of the translations regarding this branch I've pulled from my old branch. Other things, which I've previously forgotten, I've just translated to my best intuition ... This should bring most of the translations for major/4 up to date.

EdoStorm96:
Ok, I've also adressed a dead link on index.html here. I ran into this during the translations. fixes #710.

EdoStorm96:
Woops, so I missed the issue with the tests you pointed out, which turned me onto the other issue with the tests, that we discussed earlier.

reviews/tests.py
```diff
@@ -258,18 +258,24 @@ def test_auto_review(self):
...
-        self.study.stressful = YesNoDoubt.YES
+        self.study.risk = YesNoDoubt.YES
         self.study.save()
 
         reasons = auto_review(self.proposal)
         self.assertEqual(len(reasons), 6)
 
-        self.study.risk = YesNoDoubt.YES
+        self.study.proposal.researcher_risk = YesNoDoubt.YES
         self.study.save()
```
miggol, line 268: The tests seem to pass fine without saving the object. But if you could save the proposal object here instead of the study at least it will be internally consistent, should there be a case (in the future) where saving does matter.

studies/models.py
miggol, line 1: Ah yes, way better like this. Backslashes stress me out

proposals/utils/pdf_diff_logic.py
```diff
@@ -1029,15 +1051,17 @@ def create_context_pdf(context, model):
                     sections.append(StudyOverviewSection(study))
                     sections.append(InformedConsentFormsSection(study.documents))
 
-                extra_documents = get_extra_documents(model)
+                sections.append(KnowledgeSecuritySection(proposal))
```
miggol, line 1054: When generating a PDF for a proposal with sessions, line 1050 causes an error. I suspect this is just a mismatched indent and that the line just needs one more indentation level.

This doesn't seem to be related to this PR in particular, so I couldn't select the specific line in this review. But if you could quickly fix it here that would be great.

miggol (changes requested):
With every PR things are looking better and better, on both the code and application side. Having a hard time keeping up with all the improvements.

I wanted to mention that the diff page needs a new lick of paint but I see there's already an issue for that (#706).

Because this branch as is does have breaking errors I'm not approving right now but I expect the fixes to be simple.

miggol (approved):
Looking good

PR #711, opened 2024-10-01
Fix/diff and other minor issues.

EdoStorm96:
So this branch is just some odds and ends. 

I firstly wanted to update the diff page, so that it makes use of Bootstrap's table class and such. I had to add some css to make the styling look like the old version. This is now in the diff.css static file.

When testing the diff, I noticed that different study types (sessions, interventions, observations) remain linked to the proposal, even though you might change your proposal's design to not include them. So I added the ability to StudyDesign to delete objects, as well as create them.

I found that TaskCreate did not receive a proposal, so I fixed this.

I also added the diff.css static file to the compare documents pages, so that it uses the same styling.

Finally I noticed that Session.net_duration(), did not take the repeats of the session into account. So I've implemented this as well.

Sorry for this chaotic PR ... I keep noticing small things and wanting to fix it right away ... Hope you can make sense of it. The commits help :)


tasks/models.py
```diff
@@ -49,7 +49,7 @@ def net_duration(self):
         if duration := self.task_set.annotate(
             total_duration=models.F("duration") * models.F("repeats")
         ).aggregate(models.Sum("total_duration"))["total_duration__sum"]:
-            return duration
+            return duration * self.repeats
```
EdoStorm96, line 52: Ok, now that I'm reading the help text again, I am actually thinking, the old version, which did not take session repeats into account is more accurate to the goal of this question. What do you think?

studies/views/study_views.py
```diff
@@ -106,6 +106,15 @@ def form_valid(self, form):
...
             form.instance.__setattr__(study_type[0], form_value)
+        if not form.instance.has_intervention and hasattr(
+            form.instance, "intervention"
+        ):
+            form.instance.intervention.delete()
+        if not form.instance.has_observation and hasattr(form.instance, "observation"):
+            form.instance.observation.delete()
+        if not form.instance.has_sessions and form.instance.session_set.all():
+            for session in form.instance.session_set.all():
+                session.delete()
```
miggol, line 117: It can be annoying to have these instances dangling around when the type is no longer relevant. But the unintentional feature that comes with it is that users can undo deletion by re-enabling it. Which is nice and we would probably want to keep.

Best to solve the issues with these objects appearing elsewhere symptomatically.

To do this right, we could move to a `Study.get_sessions()` function where needed. That function would return no sessions if not `study.as_sessions`.

Or we could assign custom managers to things like `study_set` and `sessions_set` that do the same.

Just a suggestion. A couple if statements to fix the diffs is fine by me as well, just a little harder to maintain.

miggol (changes requested):

EdoStorm96:
Ok, so I've adressed your comments. I've also merged major/4 into this branch to avoid conflicts.

So I implemented get_intervention, get_observation and get_sessions property methods on the Study model. For now we are only using these for the diff ... I guess it could remove some if-statements in other places ... What do you think? Have a look at other places where this can be used or just leave it?

miggol (commented):
I still had a pending review on this PR. I see it's already been merged so I hope this still comes through. Nothing major but if you could address my final comment in this same branch and open new PR (no need for another review) that would be great!

PR #715, opened 2024-10-08
Feature/legal justification question

EdoStorm96:
So this adds the legal justification question to every study. I've created a new page, after the Participants page, called Personal Data, here the new question gets asked.

There is nothing too crazy going on, I´d say. The basic idea works well. The only problem is that I am not sure about any textual elements of this PR ... I guess it is fine right now? But ideally the secretary has a look at this as well. I'll ask her for feedback too ...


miggol (changes requested):
Just the issue with error logic needs to be fixed. The rest can be addressed with a cleanup pass after testing or feedback from the secretary.

EdoStorm96:
All fixed!

PR #719, opened 2024-10-14
Revise attachments while preserving older versions

miggol:
If you edit an attachment that is already attached to something else, the portal will instead seamlessly save a copy so older proposals don't change.

To test:

Revise an application with an Attachment attached. Modify the attachment in the revision, and note that it gets a new pk while keeping all attributes that you would expect. The old application and its attachment will remain unchanged.

If this looks simple know that I played around with this a lot before getting it right, and the whole rest of the attachments system is pretty much built around this part being easy. Still proud of it, though!

### DIFF
--- a/proposals/views/attachment_views.py
+++ b/proposals/views/attachment_views.py
@@ -40,8 +40,31 @@ class AttachForm(
             del self.fields["kind"]
 
     def save(self,):
-        self.instance.kind = self.kind.db_name
-        self.instance.save()
+        # Set the kind if enforced by the view.
+        if self.kind:
+            self.instance.kind = self.kind.db_name
+        # Check we're not editing an attachment that is still in use
+        # by another object.
+        if self.instance.attached_to.count() > 1:
+            # Saving this instance might remove historical data. So
+            # we create a "revision" of sorts. First we save the old
+            # pk:
+            old_pk = self.instance.pk + 0
+            # The following means this instance will get saved under a
+            # new pk, effectively creating a copy.
+            self.instance.pk = None
+            self.instance.id = None
+            self.instance._state.adding = True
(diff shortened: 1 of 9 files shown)


proposals/views/attachment_views.py
```diff
@@ -39,11 +39,32 @@ def __init__(self, kind=None, other_object=None, extra=False, **kwargs):
...
+            # The following means this instance will get saved under a
+            # new pk, effectively creating a copy.
+            self.instance.pk = None
+            self.instance.id = None
+            self.instance._state.adding = True
+            self.instance.save()
+            # Retrieve and fix the old attachment.
+            instance_manager = type(self.instance).objects
+            old_attachment = instance_manager.get(pk=old_pk)
+            old_attachment.attached_to.remove(self.other_object)
```
EdoStorm96, line 62: I was kindoff confused about old_attachment not getting saved here ... But it seems to work as expected, as in, the new object gets removed from the old attachment. I've tested this in the shell.

EdoStorm96 (changes requested):
So besides the little creation bug, this works well AFAICT. Nice and simple. Good job!

proposals/views/attachment_views.py
miggol: I think elsewhere we do save in these cases "for good measure", but I think that we shouldn't. M2m relationships get updated instantly and it's good to be reminded of that now and again. I also sometimes forget if it's necessary to save them, but it isn't.

miggol:
I've fixed the creation bug, thanks for pointing that out. I've made the solution quite a bit more verbose to make more clear what's going on.

EdoStorm96 (approved):
Haven't tested ... But looks good!

PR #721, opened 2024-10-21
Feature/attachment download permissions

EdoStorm96:
So the main purpose of this branch is to add permissions to attachment downloads, but I also did some housekeeping along the way.

- I added cdh_files to .gitignore
- I removed this strange if-clause from UserOrGroupsAllowedMixin. At first I thought this was messing with my implementation, but I later realized this will never be True, because users have no attribute is_supervisor. I don´t know what this was doing or what it was intended to be doing, but it confused me so I got rid of it.
- I found an unused mixin in proposal_views, so I got rid of it.

The main meat of this PR is the final commit, where I actually implement to permission. Right now I implemented it so that:
- applicants have access
- the supervisor has access
- the secretary has access
- members of the relevant reviewing chamber have access

Right now these permissions are separate from the state of the review or proposal, so these people will always have access. In my mind, we can trust the people with permissions to not do anything too crazy with their permission, so for me this is fine.

Let me know what you think!


proposals/views/attachment_views.py
```diff
@@ -291,3 +291,24 @@ def get_file_response(self):
...
+            pk=self.kwargs.get("proposal_pk"),
+        )
+        allowed_users = list(self.proposal.applicants.all())
+        if self.proposal.supervisor:
+            allowed_users.append(self.proposal.supervisor)
+        return allowed_users
+
+    def get_group_required(self):
+        group_required = [
+            settings.GROUP_SECRETARY,
```
miggol, line 307: How about the voorzitters as well?

miggol (approved):
Looks good, just need the voorzitters in the groups list.

In the future I think it might be cool to have a proposal method that spits out groups and users that can access it. But that's not really a priority now.

PR #723, opened 2024-10-22
Fix errors for StudyEndForm

miggol:
If you create a study that has an intervention and/or an observation, but no sessions, there will always be an error in `StudyEndForm` ("Trajectory Overview"). This happens because the deception fields get removed *after* `super().__init__()` cleans and validates the form. So you'll have an error in a field you can't fill in.

This PR fixes that by cleaning the form again once all fields are set up.

Bonus feature: I got fed up of stashing and popping my fake local email settings every time so I set up a basic `debug_settings.py`. I've added it to gitignore so put anything in there you'd like.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ certs/
 fetc/ldap_settings.py
 fetc/saml_settings.py
 *.sqlite3
+debug_settings.py
 
 ### Coverage ###
 .coverage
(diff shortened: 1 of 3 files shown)


EdoStorm96 (approved):
Yess Love this!

I was also struggling with this bug. This is great!

PR #725, opened 2024-10-23
fix: create and implement attachment filename generator

EdoStorm96:
So here is a little filename generator. It took me a bit long to figure out, but with some tips from Ty, I got it to work. It did not quite make sense to use the old FilenameFactory, as the filename generator of cdh FileField, expects something different. 

So this function just receives the FileWrapper ... From there I search for the relevant ProposalAttachment or StudyAttachment and get the relevant info from there. This is a bit quick and dirty ... But it works for now.

Although there is some overlapping code between this and the FilenameFactory ... they are just different enough, that I thought let's just have them coexsist ... I guess that can be a helper function that removes the overlap, but this does not seem like a huge win to me, both for efficiency, or readability.

I have not though through all the possibilities for attachments resulting in the same filename ... I think I will first start implementing the slots and checkers a bit more, to get a grasp of all the possibilities for things like this to occur ... As a starting point, this is ok imo.

### DIFF
--- a/attachments/models.py
+++ b/attachments/models.py
@@ -2,11 +2,9 @@ from django.db import models
 from django.contrib.auth import get_user_model
 from django.utils.translation import gettext as _
 from main.utils import renderable
+from attachments.utils import attachment_filename_generator
 
 from cdh.files.db import FileField as CDHFileField
-
-# Create your models here.
-
 
 class Attachment(models.Model, renderable):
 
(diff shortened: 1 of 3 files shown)


miggol:
See the way I added an FETC-filename link in the detailed attachments review view (#752) for how I see the filename generator being used

EdoStorm96:
So this is working as it should now (I think?)

Nothing too crazy. The way to get to the proper attachments and related objects is a bit annoying in the generator function ... But I've discussed this with Ty, and this is the best we can do. The generated filename gets displayed in the proper place.

I also ran into one more bug, while testing, which I have now created a fix for here ... I have no idea how we did not run into this earlier, but Study.research_settings_contains_schools, after filling in the StudyDesignForm, causes a major breaking bug. I can also push the to the bugs branch if you'd prefer. It should make use of get_sessions/intervention/observation checks

miggol (commented):
Hi, sorry, this review was apparently still pending.

As discussed, I will be adopting the PR from you.

PR #743, opened 2024-11-05
Feature/extra attachments

miggol:
This branch allows for the adding for extra attachments to a proposal and its studies.

An additional change to this branch is that we now allow changing the kind ("Type") of an attachment in the attachment update views. 

Earlier I thought it would make sense to always enforce the kind if it was set by a slot. Now, when implementing extra files this felt like an overreach which would require deleting and re-adding of files by the user if they ever wanted to change the type.

Changing the type of a required attachment will simply move it from the required to an extra slot, as the attachment no longer fulfills the requirements for the required slot. Only if you are adding a **new** attachment to a required slot will the kind be enforced by the view.

I recommend taking a look at the changes to `Stepper.attachment_slots()`, which is now a property that matches slots on the fly. Slots generated by checkers now live in `Stepper._attachment_slots`, but that change should be of no consequence as long as the `Stepper.add_slot()` method is used.

### DIFF
--- a/attachments/kinds.py
+++ b/attachments/kinds.py
@@ -48,7 +48,7 @@ class DataManagementPlan(ProposalAttachmentKind):
 class OtherProposalAttachment(ProposalAttachmentKind):
 
     db_name = "other"
-    name = _("Overige bestanden")
+    name = _("Overig bestand")
     description = _("Voor alle overige soorten bestanden")
 
     def num_required(self):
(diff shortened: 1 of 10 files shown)


EdoStorm96 (approved):
Nice work overall! Some clever stuff here. I did find one crazy bug, but the fix is super simple, so, granted you fix this, I approve :)

Issue #733, opened 2024-10-24
Registration situation


EdoStorm96:
So as off the new attachments system, we want to check if a study measures using video or audio recording. Right now we can only check this clearly for observations and tasks ... the `measurement` field for interventions is just a text field. However that is not the only problem/annoyance: Both observations and tasks make use of a registration model, to provide options for picking the measurement method. However, these are separate models with the same name. Even though, their attributes basically fully overlap:

![image](https://github.com/user-attachments/assets/88ec6a6c-0f49-4aa5-9a01-c01e978aed5f)

The options are loaded as migrations (registrations.json and 00_registrations.json) and although there is a minor difference between the options, there should just be one registration model, covering for both these cases. We can then select the choices manually for different study types, if so desired.

I would prefer though, to have a question of registration method, for each study, regardless of study type, possibly on a seperate page. That would increase uniformity, and allow for more possibilities + checking if a study features audio or video recording would become way more efficient.

PR #753, opened 2024-11-07
Feature/all attachment kinds

EdoStorm96:
Woopee, there we are. While closely reading along the master instruction and [this page](https://intranet.uu.nl/en/knowledgebase/documents-ethics-assessment-committee-humanities?check_logged_in=1) I've implemented all the possible attachment kinds into the new attachment system. While writing up the different kinds was doable ... the logic for when certain slots are to be added with the right desiredness was quite a pain and led me down deep paths of the portal's architecture. The bulk of the action happens in the `StudyAttachmentChecker`. I am not sure if this is the most elegant solution for implementing all this logic ... But it becoming somewhat of a mess was unavoidable ... But I'm open to tips for improvements! I think the best way to understand all this is to keep the document/email with the master instruction close at hand.

Checking whether a study features recordings is especially annoying due to #733... I was also struggling a bit with making sure the other of a pair becomes optional when one is present. The wishes for this are also quite specific, so I could not find a nice way to generalize this behavior.

I added some pretty extensive descriptions taken from the page linked to above. These are currently not used, but might come in handy later.

I though I would also tackle translations, so I did this for the whole attachments branch.

Finally I implemented the attachments page into the general flow of the form, so now the back and forward buttons work correctly ... As of the current state of the portal this requires a form, so I added an empty form for enabling this.

Sorry for the big commits, (6654c4a42259e739f5f5a84af57af6091447d93e & ba2ddfdd3181e1aacddb07f9e546b1b4b5febec9) but it just made most sense as a whole imo.

### DIFF
--- a/proposals/utils/checkers.py
+++ b/proposals/utils/checkers.py
@@ -12,7 +12,7 @@ from tasks.views import task_views, session_views
 from tasks.models import Task, Session
 
 from attachments.utils import AttachmentSlot, desiredness
-from attachments.kinds import InformationLetter, DataManagementPlan
+from attachments.kinds import DataManagementPlan, LEGAL_BASIS_KIND_DICT
 
 from .stepper_helpers import (
     Checker,
@@ -472,12 +472,13 @@ class StudyAttachmentsChecker(
     def check(
         self,
     ):
-        kind = InformationLetter
-        info_slot = AttachmentSlot(
-            self.study,
-            kind=kind,
-        )
-        self.stepper.add_slot(info_slot)
+        if self.study.legal_basis is not None:
+            kind = LEGAL_BASIS_KIND_DICT[self.study.legal_basis]
+            info_slot = AttachmentSlot(
+                self.study,
+                kind=kind,
+            )
+            self.stepper.add_slot(info_slot)
         return []
 
 
--- a/proposals/utils/checkers.py
+++ b/proposals/utils/checkers.py
@@ -4,9 +4,12 @@ from django.urls import reverse
 from proposals import forms as proposal_forms
 from proposals.models import Wmo
 from studies import forms as study_forms
+from studies.models import Study
 from interventions import forms as intervention_forms
 from observations import forms as observation_forms
+from observations.models import Registration as observation_registration
 from tasks import forms as tasks_forms
+from tasks.models import Registration as task_registration
 
 from tasks.views import task_views, session_views
 from tasks.models import Task, Session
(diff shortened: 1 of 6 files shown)


proposals/utils/checkers.py
```diff
@@ -472,14 +486,125 @@ def __init__(
...
+
+    def at_least_one_fulfilled(self, slots):
+        """
+        Function which receives a list of slots and checks if at least one has
+        been fullfilled. It returns None, or the fullfiled slot
+        """
+        fullfilled_slot = None
+        for slot in slots:
+            # check if any of these slots have been fullfilled yet using match()
+            slot.match(exclude=[])
```
miggol, line 603: This is somewhat abusive of the match function, because `slot.match()` modifies the internal state of the slot, setting its kind and attachment based on what it finds.

I think this currently does no tangible harm, but it's definitely something that can potentially go wrong in the future. I would like to request the following change:

* Modify `slot.match()` to just return the matched attachment or False if nothing is found. It should no longer change the state of the slot.
* Add a new `slot.match_and_set()` method that calls `match()`, and if it finds something, it sets `self.kind` and `self.attachment`.

Like so we can use `slot.match()` for use cases like in this checker, and use `match_and_set()` where we actually desire to change the state of the slot, such as in `stepper.add_slot()` and `stepper.attachment_slots()`.

Otherwise reusing the match method here is a great solution, I just don't like all these slots getting different attachments and kinds set for every pass of the checker.

miggol (changes requested):
Awesome! I'm very happy with this.

At first I thought that putting everything in one checker would make it messy, but this works fine.

Please take a look at #755 for making it more clear when only one of a set of documents is required.

There are however some issues with the logic. Here's the first one:

> Als op dit punt bij de grondslag toestemming nog geen toestemmingsverklaring (voor kinderen) verplicht is, wordt de gewone toestemmingsverklaring grondslag toestemming verplicht gesteld.

I believe that in the current state, that final "toestemmingsverklaring grondslag toestemming" is made mandatory even if the consent forms for kids are already mandatory. That shouldn't be the case.

I believe the same should be the case here:

> In het geval dat er in het traject is aangegeven dat er bijzondere persoonsgegevens worden verzameld, en er nog geen akkoordverklaring voor opnames of script voor het maken van opnames in ingediend, wordt bij algemeen belang de toestemmingsverklaring algemeen belang met bijzondere persoonsgegevens verplicht gesteld.

Although here it's definitely an error of mine in the "law". I would reword the above as follows:

> In het geval dat er in het traject is aangegeven dat er bijzondere persoonsgegevens worden verzameld, en er nog geen akkoordverklaring voor opnames of script voor het maken van opnames **verplicht is gesteld**, wordt bij algemeen belang de toestemmingsverklaring algemeen belang met bijzondere persoonsgegevens verplicht gesteld.

Basically this avoids the case where an attachment is mandatory only if a different mandatory file has not yet been added. That makes no sense, because that means that the first file will always be optional, because the second file will always be added anyway.

I think this makes the logic simpler in the end, but sorry for the extra work cause by my mistake.

EdoStorm96:
I would say I have adressed all you comments. I have also made it the case, that a consentform for adults gets required if the study has adults and the stuff for children, if a study has children. These being interdependant did not make much sense, although we will check with Desiree on this.

attachments/utils.py
```diff
@@ -35,25 +35,36 @@ def __init__(
...
 
+    def match_and_set(self, exclude):
+        """
+        Uses self.match() to find a matching attachment. If it finds one, it
+        sets self.attachment and self.kind
+        """
+        matched_attachment = self.match(exclude=exclude)
+        if matched_attachment:
+            self.attachment = matched_attachment
+            self.kind = get_kind_from_str(matched_attachment.kind)
```
miggol, line 66: Match_and_set should actually still return True, or the instance, or anything truthy if it succeeds. Otherwise `stepper.attachment_slots()` will not work.

Although in that specific method we do actually want to modify internal slot state.

proposals/utils/stepper.py
```diff
@@ -66,7 +66,7 @@ def attachment_slots(
                     obj,
                     force_desiredness=desiredness.EXTRA,
                 )
-                success = empty_slot.match(exclude=exclude)
+                success = empty_slot.match_and_set(exclude=exclude)
```
miggol, line 69: This is why `match_and_set()` needs to return something. Its return value matters

miggol (approved):
Very nice! I left some comments for posterity but no need to do anything with them, I've fixed them as I went along. This branch is good now.

Issue #749, opened 2024-11-05
Selecting "This trajectory is fully anonymous." does not work


tijmenbaarda:
Path: `/studies/personal_data/`

When selecting this option, the choice is not saved in the database. This happens both when no choice was made yet and when altering a previous choice. The other two options work.

miggol:
Hoort bij 4.0, want is mogelijk al opgelost

Issue #748, opened 2024-11-05
Overzicht van alle sessies en taken - creating a session should be obligatory


djhcapel:
If I select 'Opslaan en volgende stap' on 'Overzicht van alle sessies en taken', I arrive at the page 'Overzicht en eigen beoordeling van het gehele traject', while I had expected to be forced to create a session first. I do see a warning on 'Overzicht en eigen beoordeling van het gehele traject' that no session has been created yet, but I should not have been able to get to that page in the first place, right?

EdoStorm96:
Mmm yes ... This is a bit more of a design choice. On the session overview you are able to click: create a session. To create sessions and tasks, you cannot just keep clicking save and go to next page, but you have to manually create them ...

miggol:
Decided in meeting to make the "Time to create session" text more angry/imperative. It should be made clear that this trajectory is *not finished* even though we do allow the user to continue.

Issue #742, opened 2024-11-05
Participants - hierarchy


djhcapel:
This question should be obligatory, since it is a criterion for long-route applications.

miggol:
High-priority, this should be obligatory.

Worth checking if other fields also have this issue, because it's a normal form field that *should* have been validated.

Issue #741, opened 2024-11-05
After failing validation of WMO application form, the upload box is visible even if "no" is selected


tijmenbaarda:
Path: `/proposals/wmo/application/`

See image. The upload box should be invisible, because "no" is selected. When selecting "yes" and then "no" again, the box disappears. Probably not a high priority though.

![image](https://github.com/user-attachments/assets/cd2294a3-a0e1-4c5a-ae6a-96ce5efc88ab)

miggol:
Worth fixing, JS issue, shouldn't be too big. But not high-priority for 4.0

PR #754, opened 2024-11-07
Fix/bugs from testing

EdoStorm96:
Let's keep this ball rolling!

So this covers most of the bugs found from testing. The only one left is regarding the stepper colors ...

#739:
So this was an oversight on my part (you might notice a pattern in this PR ... ;p) But yeah, some pretty funky functionality. I had to make sure wmo.js worked correctly and supply some ok-looking styling from check-wmo for the resulting little message. Seems to work fine now.

#742:
Ensure `hierarchy` gets validated.

#749:
Another oversight on my part. The PR (#715)  where I implemented this stuff featured an almost identical bug, which you pointed out ... But I didn't bother to check if I had made that same mistake again (`if not cleaned_data["{field}":`) which becomes buggy when validating falsey values. Woopsiee. Won't make this mistake again!

#748:
Did some style stuff on the warning for "you forget to make tasks/sessions" ... Nothing crazy. However ... I did run into an annoying issue. On SessionEndForm/session_end.html, there is a non_field_error provided, which is mostly used for validating in the stepper... SessionOverviewForm has a similar purpose. However, SessionOverviewForm, does not get rendered at all. On SessionEndForm, we do render this form, so this results in an ugly duplicate warning. I was figuring out a way to remove this. Got a bit annoyed and brute forced my way through with a script tag on the template, removing the warning by class name. It's very hacky, but it works for now ... Maybe you have a better solution?

#741:
So the WMO_application page now works as expected. This is also due to a tweak of depends_on_value, where it can now also clear radio input that were dependant on an earlier question. I've gone through every use of depends_on_value and this change does not cause any issues anywhere.

### Edit no. 2

I've also just pushed some other minor stuff. like using get_intervention on the StudyOverview and a minor translation fix.

The final thing is a bit bigger: I've converted StudyDesign to a normal ModelForm, to ease validation of this. It was a bit annoying and caused some of the Stepper issues ... This did require a custom rendering of this form, which is pretty crazy. But the backend stuff is much simpler now.

### Edit no. 3

Ok, so I've just pushed a few more things to this branch ... Sorry this is getting unwieldy. But I will leave it alone from now!

#645:
I've ensured that if there are no other applicants, applicants will always be reset to just the user, and if there are other_applicants, the user will always get added to applicants. This makes the form validation simpler, but, although it complicates the form_valid a bit ... The one bug that can still occur on this page, is that if you first say yes to other_applicants, then remove all applicants and change other_applicants to "No" again, you will get stuck. Because of the MultipleSelectWidget's error message, which is invisible ...

Sidenote: The MultipleSelectWidget seems to be unable to display error messages, which is why error's with the applicants field, get displayed on the other_applicants field, which leads me to:

#730:
So this was happening because the MultipleSelectWidget's inability to display errors ... See my [issue](https://github.com/CentreForDigitalHumanities/django-shared-core/issues/109) in the DSC for this. With this specific issue, I've removed most of the errors that could occur from this form, by forcing some correctness with my fix for #645. But the final form error for the applicants field, does display on the other_applicants field. As I proposed in my comment on #730.

#728:
I've added supervisor to `_soft_validation_fields`. Easy fix. Here, I've also made sure that any errors with the MultipleSelectWidget, get displayed on the "relation" field ... Bit annoying, but what can you do. (for now)

#729:
So, this was a similar issue & fix as with #741. I simply made sure that check_field_required, clears radio inputs, if it makes a field disappear. This ensures that fields whose visibility is dependant on multiple layers, get removed when desired.

PS I'm sorry about the big branch ... But at least it fixes a lot of issues (9 ... lol). But will try to avoid this many issues in a single branch in the future. It was just so tempting to keep on pushing :p

Happy reviewing :)


studies/forms.py
```diff
@@ -276,31 +260,19 @@ class Meta:
...
-        if not "study_types" in cleaned_data:
-            error = forms.ValidationError(
-                _("Je dient minstens een van de opties te selecteren."), code="required"
-            )
-            self.errors["study_types"] = error
-
-        # this checks the hidden fields, and could be used for validating this
-        # form elsewhere
-        if not any(cleaned_data.values()):
+        if not True in cleaned_data.values():
```
miggol, line 275: I bet this feels good

studies/templates/studies/study_design_form.html
```diff
@@ -0,0 +1,21 @@
...
+
+<div class="d-flex flex-column">
+    <label class="form-label d-block">
+        {% trans "Om welk type onderzoek gaat het hier? Je kan meerdere opties aankruisen." %}
+    </label>
+    <div class="uu-form-row">
+        <div class="uu-form-field">
+            {% for field, field_errors in fields %}
+                <input type="checkbox" class="form-check-input" name="{{ field.name }}">
+                <label class="form-check-label">{{ field.label }}</label>
```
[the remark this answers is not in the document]
miggol: Okay this form template is also not rendering initial values. Might be best to use a Django form template as a reference to make sure nothing else is missing.

Or maybe we can still just render the actual widget, and changing the form and div around it is enough?

tasks/templates/tasks/session_end.html
```diff
@@ -43,3 +43,12 @@ <h3 class="mt-3">{% trans "Taken" %}</h3>
     </div>
     <p>{% trans "Beantwoord op basis van dit overzicht de volgende vragen:" %}</p>
 {% endblock %}
+
+{% block post-form-pre-buttons %}
+    <script>
+    // So this form features a non-field error which gets used for validation
+    // However, it does not need to be visible here ...
+    // this is a very hacky way to remove it.
+    document.getElementsByClassName("alert-primary")[0].remove()
```
miggol, line 52: This does nothing for me currently. This class isn't used

EdoStorm96:
Allright, I've adressed your comments. These were mostly pretty easy fixes, also thanks to your suggestions :)

Some things to note:
- I've removed the SoftValidationMixin from StudyDesignForm, as that was messing with the validation for me.
- I've added the missing migration that has been around for a while now.
- So I've gone through all occurences of study.has_{study_type} and replaced it with get_{study_type}, when it made sense. I think it is alright like this, but this was a bit heady.

Everything seems to work well :)

miggol (approved):
Awesome, all good!

There's an issue where StudyDesign and StudyEnd validate as complete without having visited them. But that was there already, I think.

EdoStorm96:
> There's an issue where StudyDesign and StudyEnd validate as complete without having visited them. But that was there already, I think.
Ughh what is up with these?? So annoying ...

PR #755, opened 2024-11-13
Feature/all kinds suggestions

miggol:
This PR attempts to make relationships between optional files more clear.

image

Slots can now receive an `OptionalityGroup`, which in the general case does nothing except be a renderable. However, the `merge_groups` utility can use these groups to add some structure to a list of slots.

Currently only one layer of optionality groups is supported. However, multiple layers can be implemented with recursion if necessary.

Also I've added help texts to the side of the attachment slots.


EdoStorm96:
Super nice, imma take this and merge it

EdoStorm96 (approved):

Issue #756, opened 2024-11-13
Remove DMP page


miggol:
Now that the DMP is a regular attachment, the DMP page and step in the form is kind of pointless. It would be good to discuss what to do about the info on that page now. We could also just keep it with just the "I have spoken to the privacy officer" checkbox, maybe move it to before the attachments page.

PR #758, opened 2024-11-25
Fix/remove old dmp filefield

EdoStorm96:
Pretty straightforward PR, I've removed the old DMP file field and rearranged the order a bit, as decided upon in #756:

Now the order is:

knowledge security > DMP > attachments

As a small bonus, I've added a description to the DMP AttachmentKind.


miggol (approved):
Tasty!

PR #760, opened 2024-11-27
Feature/bubble colours

miggol:
The core change of this PR consists of the removal of the `incomplete` CSS class for stepper bubbles. Bubbles are now either grey, or green if completed.

Some additional changes were made to make this make sense:

* OtherResearchersForm had some strange defaults that prevented it from being incomplete. Those fields now have `null=True` and `default=None`
* ProposalSubmit now only turns green when there are no errors at all anywhere else. The function that prevents recursion within errors is a bit funky but works well enough
* New DisabledItem for placeholders inserted by the stepper. And PlaceholderItems are now not complete by default.

To review this PR effectively you may need to fix some `has_intervention` and friends. I didn't want to commit that here because I knew it was being worked on elsewhere.

Also StudyDesignForm and StudyOverview turn green by default but I recall those also being worked on so I left them as is for now.

image

It honestly looks more overzichtelijk like this anyway.


EdoStorm96 (approved):
I like this a lot! It retains the perks of the previous design, but is simpler and dare i say, easier on the eyes. It mostly works very well, so I'm quite pleased. I still have a few issues with this PR, which you can maybe take a look at:

- Now, when there are no required slots in Attachments, it will go green. Maybe you can add 
to AttachmentsItem.get_errors, a check where if there are no required slots, it also adds an error. Currently, when you start out making a study, there are not yet any required slots and it turns green.

- AttachmentsItem.get_errors also does not take its child into account when it is collapsed. I came up with this lil hack, but am sure you can find a better way:
```
        if not self.is_expanded:
            errors += [child.get_errors() for child in self.children][0]
```

- The submit form does not check its own form errors ... its submitable check works well otherwise (though I will not pretend to fully understand how ...). Buy the ProposalSubmitForm's own errors are not taken into account. This form's clean does not seem to get called. this is possibly a similar situation as with StudyDesign and StudyEnd.

Given you fix the stuff with the attachmentsItem, I approve. The Submit form issue, is maybe for another issue/PR, together with StudyEnd & StudyDesign

miggol:
Alright, the changes to this branch since your last review have proven to be larger than expected so I would like to request a new review.

`ProposalSubmitForm` now inserts an error if the proposal still has errors according to the stepper. This prevents the submission of an unfinished proposal, and is regulated by the `final_validation` form kwarg. Previously, we only hid the button, but the form would actually validate and submit if asked politely.

Additionally, `ProposalSubmitForm` has a custom template that does not show form errors, as these are already shown in the `get_form_errors` box. This template can be reused for other forms that should not show non-form errors. (`StudyDesign` perhaps?)

SoftValidationMixin has been added to a number of views to fix validation woes. I now better understand SVM and its `initial_clean` method, which allows us to validate forms despite them being unbound.

Initially I had added an `AttachmentsItem` error if no documents at all have been added. However, because I foresee annoying exceptions with non-standard proposals this was reverted. #wontfix

The `StepperItem.get_form_errors` method now follows the mixin paradigm and the `include_children` part is now inherited by default for all Items. So now we start with `errors = super().get_form_errors(**kwargs)` when overwriting it which optionally gets all child errors.

EdoStorm96 (changes requested):
Allright, Nice work! I'm very happy overall. Just two more forms are still causing some trouble ...

**OtherResearchersForm:**

I feel like I had done this at some point ... But I guess not. The other researcher form, with no data, 
validates as correct. Looking in OtherResearcherForm.clean() the other_applicants and other_stakeholders fields do no get validated.

Can be fixed with a simple `self.mark_soft_required(cleaned_data, "other_applicants", "other_stakeholders")` in the form's clean()

**StudyEndForm:**

StudyEndForm still goes green with no data. I'm pretty sure this is due to lines 351 & 352, which were I think a part of former attempts to fix this form's validation. Now it should get removed though.

Ah, but that then reintroduces the issue where deception gets validated again ... It does not get removed from self.fields, as it should in the form's `__init__` ...

One solution I came up with is to hide the field in the frond end:

- have has_sessions as a HiddenInput on this form
- do not attempt to remove the "deception" field anymore
- add this to the clean:
```
        if self.instance.has_sessions:
            self.mark_soft_required(cleaned_data, "deception")
```
- add this to the form's template:
```
            depends_on_hidden_value("has_sessions", "True", "deception");
```
Bit annoying, but it get the job done. Maybe you can come up with something better, but I think for now, this will have to do.

miggol:
Alright, I've addressed your issues.

`StudyEndForm` could be fixed by running `self.initial_clean()` rather than the regular `clean()`

`UpdateOrCreateChecker` now creates an error by default if the object is not created, e.g. interventions and observations.

Also stepper items now receive `get_checker_errors` by default. This isn't used here, but I was working on it as an option for the above and decided to keep it.

EdoStorm96 (approved):
Nice! Approved!

PR #780, opened 2024-12-02
Fix/textual changes

EdoStorm96:
**NOTE: DEPLOYING THIS REQUIRES RUNNING `loaddata`**

Woops I did it again ...

Yeah, so I thought let's fix all the textual changes in one branch, but it did become quite big, because there were more and more textual changes dripping in ... Because I wanted to fix the translations for the textual changes, I also did a lot of other unrelated translations, mostly concerning the attachments. There is nothing too crazy going on here, but it is a lot. Good luck with the review!

Something to note:
So I was working on django.po (the first time) and had everything done, but then I realised it would be nice for you to have a commit of the auto-generated django.po, for review purposes. So I stashed my edited django.po, committed the auto-generated one and applied my stash. This caused merge conflicts, where I always went with the changes from my stash of course. However I did not realise that where there were no merge conflicts, it would just merge them of course. Because of this, in my edited django.po, it looks like there is a whole bunch of untouched fuzzy translations, but these translations are actually correct. I only realised this on my second pass of django.po ... Which is when I removed them.


main/templates/main/index.html
```diff
@@ -71,7 +71,7 @@ <h2>{% trans "Voordat je begint" %}</h2>
                             Gebruik de juiste (meest recente)
                             <a href="https://intranet.uu.nl/documenten-ethische-toetsingscommissie-gw"
                                target="blank">voorbeelddocumenten</a>
-                            voor de <em>informed consent</em>. (de laatstse versie is van <u>december 2021</u>)
+                            voor de <em>bijlagen</em>. (de laatstse versie is van <u>december 2021</u>)
```
miggol, line 74: Best to update this date. The intranet site says last updated October 3rd 2024

observations/models.py
```diff
@@ -83,9 +83,9 @@ class Observation(SettingModel):
     is_nonpublic_space = models.BooleanField(
         _("Wordt er geobserveerd in een niet-openbare ruimte?"),
         help_text=_(
-            "Bijvoorbeeld er wordt geobserveerd bij iemand thuis, \
-tijdens een hypotheekgesprek, tijdens politieverhoren of een forum waar \
-een account voor moet worden aangemaakt."
+            "Bijvoorbeeld er wordt geobserveerd bij iemand thuis, tijdens een "
```
miggol, line 86: I would like a colon after "Bijvoorbeeld:", or perhaps "Er wordt bijvoorbeeld geobserveerd..."

miggol (approved):
Very very nice.

When reviewing I just went through the changes, didn't go through every issue to see if everything was addressed. I only found two minor issues so I'm approving.

Issue #781, opened 2024-12-02
New Documents: recording kids


EdoStorm96:
So if there are recordings being made of children, there needs to be an Akkoordverklaring - Algemeen belang - beeld & geluidsopnames for children, with parents or without parents.

So the forms we have now: 
- Akkoordverklaring - Algemeen Belang - beeld & geluidsopnames (adults) 
- a script voor mondelinge toestemming

For adults it will be either one of these.

Then if we have public interest, with children and recordings, we need either a:
- Akkoordverklaring - Algemeen Belang - beeld & geluidsopnames (children, no parents present)
- Akkoordverklaring - Algemeen Belang - beeld & geluidsopnames (children, parents present)

EdoStorm96:
Also, the script should also be an alternative to the consent form, if there are recording, legal justification is consent and it is with adults.

PR #783, opened 2024-12-02
Fix/recording attachmnt kinds

EdoStorm96:
Implemented #781.

Yeah the logic is getting pretty insane in StudyAttachmentsChecker ... just barely manageable, but in the future it would be better to streamline this mess ... For now, this works.

I am having one problem ... The order of items in an optionailtyGroup is not consistent ... Reloading the page, will yield the items in different orders and there is AFAICT no way to set this, because of the current design ... Any ideas on how to fix this are aprriciated. It is a little annoying, but not breaking IMO.


miggol (changes requested):
I have three main issues here. 

1. `check_has_recordings` should be a Study method. I know this method was added earlier and I didn't mention it there, which is sloppy of me, but I also don't think it's acceptable to match Registrations in the code by their name. A production app shouldn't be breakable by changing the name of a registration.
2. The setting of the desiredness for optionalitygroups should be done in the optionalitygroup.

I'm working on a suggestions branch so no need to do anything right now, you'll see the PR soon enough.

The rest of the checker logic is great though.

EdoStorm96:
Ok, yeah, those were some good suggestions :) cleaned up the checker considerably and also gave the optionality group more of a purpose :) Pretty happy with this.

Ps. I did find that the check has recordings was not actually working correctly ... I do agree that using strings here is quite bad ... But I do not feel like spending any more time on the mess that is the current registration models of the portal ... Getting these consistent across all study types would be a priority for 4.1 I'd say.

miggol (approved):

PR #785, opened 2024-12-04
Recording PR suggestions

miggol:
Requires loaddata/registration editing on server

This is what optionality groups look like now:
image

Fixtures have been dumped from prod and loaded back into this branch, then `is_recording` was set accordingly, then dumped once again to end up with these fixtures.


miggol:
In the end I did still find "voldoende" to be a nicer wording. "Verplicht" didn't sit right here

EdoStorm96 (approved):
All looks good!

Issue #765, opened 2024-11-29
Gatekeeper documents


djhcapel:
See also issue #391. 

Now, only if 'school/onderwijsinstelling' or ' peuterspeelzaal/dagopvang/buitenschoolse opvang' is ticked, the gatekeeper documents pop up on the Documents page. However, this should happen in other instances as well:

-  'logopediepraktijk' and 
- 'zorginstelling'

When 'other' is ticked, gatekeeper documents are potentially required, e.g. in case of a museum. Could we add the gatekeeper  documents when 'other' is ticked, saying 'Mogelijk' instead of 'Aangeraden'?

miggol:
Will do!

djhcapel:
And of course 'mogelijk' should be 'optioneel'....(this is the term that is in use already, right?).

PR #803, opened 2024-12-05
Feat/attachment validation

miggol:
This fixes the following:

* #773
* #765 
* #779 

And requires reloading of the `main.setting` fixture, which has been updated to reflect prod beforehand.


EdoStorm96 (approved):
Nice! Approve

Issue #802, opened 2024-12-05
Make the "verder" button use the Stepper


EdoStorm96:
The "verder" button now uses available_urls, but this can still lead to the old documents page. So we should make this button use the stepper to find the first form with errors ...

Then we can probably get rid of available_urls entirely.

PR #804, opened 2024-12-06
Remove StudyConsent and available_urls

### LINKED ISSUE #761: Get rid of the old documents stuff
Of course the views, templates etc, but also eg. describe_documents.py

Good riddance!

miggol:
This fixes #761, although some vestigial Consent stuff still remains, like the option to edit legacy documents if you're the secretary.

This also removes `available_urls`, and proposal.continue_url now uses the Stepper to get the next best page to continue on. A required change was to make the Stepper work without a request, which is done by manufacturing a request with an applicant attached so the forms validate.


miggol:
For the transition period, I've chosen to keep StudyConsent available if you know the exact URL for testing purposes.

EdoStorm96:
fixes #802

PR #806, opened 2024-12-09
Feat/attachments pdf

EdoStorm96:
So this PR incorporates attachments into the pdf system. Ugh, it was not entirely pleasant to revisit this part of the portal ... But i got it working. I did quite a lot, so the PR is a bit overwhelming ... Sorry about that.

This PR only implements the PDF. I will also create a PR into this branch for implementing attachments into the diff.

A short overview of what I did:
- I reorganized the code a bit and restructured the PDF to be more in line with the current order of the form.
- I retained the PDF infrastructure for legacy documents, but these will only get rendered if there are legacy docs. As of now, there is no warning à la: "These documents use the old documents system". Could easily add that, but I figured, it will be phased out anyway, so it should be fine.
- I also implemented a PersonalDataSection, as that was still missing.
- I fixed an unrelated bug (3a3475878eecaf26d5e9534fd7a5e93c098662f2), which I think is what is causing the server error the secretary mentioned regarding the diff. In the upcoming diff branch, I've improved this line of code further, so regard this as a temporary fix.
- I implements a section object for a single attachment. This works nicely within the current system and is quite simple. However, getting all sections together is more complex ...
- I've made a little titlesection Just for adding Titles separate from normal sections. I could just use the new `renderable` mixin for this I realized later ... But this is also fine i guess.
- Then I have the giant class for collecting and organizing all attachments of a proposal into PDF sections ... This is the real meat of this PR. It is partly written as a baseclass for doing this for the diff as well, which is why there are some weird design choices (like often needing the proposal as a method argument, instead of just using self.proposal.) It works fine now imho, but it was written rather fast and messily, so the design could probably be improved.

Best of luck reviewing this @miggol ... If you have any question about this (or about how the hell the PDF system works in general) HMU.


proposals/utils/pdf_diff_sections.py
```diff
@@ -1067,39 +786,106 @@ def create_context_pdf(context, proposal):
...
+
+        objects = self._all_owners_list(proposal)
+        slot_dict = {obj: [] for obj in objects}
+        for slot in slots:
+            relevant_owner = slot.attachment.get_owner_for_proposal(proposal)
+            slot_dict[relevant_owner].append(slot)
+
+        return slot_dict
+
+    def _create_object_heading(self, owner, proposal):
```
miggol, line 850: I believe the "All Attachments" setting still needs a heading of its own. Perhaps just some kind of "TitleSection".

miggol (changes requested):
Okay, I have some minor comments that can be addressed now. I like the new file structure.

As discussed, we should confer with the secretary on what to do with the legacy documents.

miggol (changes requested):
Presuming the diff PR was up to date, I only have one remark before this can be merged. I would like to see somewhat better/more consistent spacing between sections in the Diff


(This first one is nice)
image

vs. 

image

vs.

image

But these are annoying.

Because the layout stuff mainly lived in this PR I'm bringing this up here.

miggol:
Oh and another reminder to us both to check the migrations before merging this.

EdoStorm96:
So the margins have been fixed, with a css if statement in table_with_header_diff.html, getting rid of the stupid `SectionTitle` util thing, and instead adding the title for an owner to the first section's section_title attribute.

Also the upload_date is now displayed for attachments, and the flag "gewijzigd voor revisie" got removed. This did require another very custom Row ...  `UploadDateRow`.

miggol (approved):
Added translations and approved!

PR #807, opened 2024-12-09
Feat/attachments diff

EdoStorm96:
So this is the diff part of incorporating the attachments into the pdf/diff system.

It build upon how the pdf is created, however it gets a lot more complicated with matching attachments together.

I think this solution is ok ... but it is very _mad scientist-y_ ... But it was quite a logical challenge to match attachments to a study.

This PR also assumes some knowledge of how Diffsections work (which is also quite _mad scientist-y_. A basic rundown:

DiffSections receive two sections, the first for the old_p, the second for the new_p. If either of these objects is missing, it gets replaced with None.

Good luck with this review as well!

### DIFF
--- a/proposals/utils/pdf_diff_sections.py
+++ b/proposals/utils/pdf_diff_sections.py
@@ -18,6 +18,7 @@ from proposals.utils.pdf_diff_utils import (
     TitleSection,
     get_all_related,
     get_all_related_set,
+    get_all_sessions,
     get_extra_documents,
     multi_sections,
 )
@@ -963,8 +964,8 @@ def create_context_diff(context, old_proposal, new_proposal):
                         or new_study is not None
                         and new_study.get_sessions()
                     ):
-                        old_sessions_set, new_sessions_set = get_all_related_set(
-                            both_studies, "session_set"
+                        old_sessions_set, new_sessions_set = get_all_sessions(
+                            both_studies,
                         )
 
                         for both_sessions in zip_equalize_lists(
--- a/proposals/utils/pdf_diff_sections.py
+++ b/proposals/utils/pdf_diff_sections.py
@@ -553,7 +553,8 @@ class TranslatedFormsSection(BaseSection):
             rows.remove("translated_forms_languages")
 
         return rows
-    
+
+
 class AttachmentSection(BaseSection):
 
     section_title = ""
@@ -568,7 +569,7 @@ class AttachmentSection(BaseSection):
         super().__init__(obj)
         self.sub_title = sub_title
         self.proposal = proposal
-    
+
     def make_row_for_field(self, field):
         if field in self.get_row_fields():
             return Row(self.obj, field, self.proposal)
@@ -1032,6 +1033,12 @@ def create_context_diff(context, old_proposal, new_proposal):
                         ):
                             sections.append(DiffSection(*multi_sections()))
 
+                sections.extend(
+                    AllAttachmentSectionsDiff(
+                        old_proposal, new_proposal
+                    ).get_all_attachment_sections_diff()
+                )
+
                 sections.append(
                     DiffSection(
                         TranslatedFormsSection(old_proposal),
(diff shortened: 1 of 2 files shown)


miggol (changes requested):
Okay I'm not completely done with this review yet, but I wanted to put down these comments before calling it quits for today.

I wouldn't have attacked the logic in this way, and I'm not sure it's exactly robust, but great work in getting this to work at all.

I think the attachments in the diff are a bit confusing with the different kinds of titles and lack of spacing. Perhaps rather than two titles right after each other we could just have one title containing "Documents for the proposal as a whole", "Documents for trajectory 1" etc...

Also something went wrong with the merge and this branch still references `has_legacy_documents` which causes an error.

miggol:
Recursive slot matching algorithm. Less concise than a for loop, but I liked making it. No need to use this, just wanted to share

```python3

def get_order(slot):
    if type(slot.attached_object) is Proposal:
        return 0
    return slot.attached_object.order

def insert_into_matches(old_slot, new_slot, matches):
    if new_slot:
        order = get_order(new_slot)
    else:
        order = get_order(old_slot)
    if order not in matches.keys():
        matches[order] = []
    matches[order].append(
        (old_slot, new_slot),
    )

def match_slot(slot, slots):
    if slots == []:
        # If we've run out of slots, no match was found
        return False
    potential_match = slots[0]
    if slot.attachment is potential_match.attachment:
        return potential_match
    if slot.attachment.parent is potential_match.attachment:
        return potential_match
    # Continue with the next potential match
    return match_slot(
        slot, slots[1:]
    )

def get_matches_from_slots(old_slots, new_slots, matches=dict()):
    """
    Tail recursive function that returns a dictionary of integers to tuples
    of (old, new) attachment slot sets, either of which may be None, but not
    both.
    """
    if new_slots == []:
        # If we've run out of new slots, start going through the yet
        # unmatched old slots.
        if old_slots == []:
            # If not old slots remain, return the matches
            return matches
        unmatched_old_slot = old_slots[0]
        # We've already run out of new slots, so these must be
        # unmatched slots and we can just insert them.
        insert_into_matches(
            unmatched_old_slot, None, matches,
        )
        # Continue until old slots run out
        return get_matches_from_slots(
            old_slots[1:], new_slots, matches=matches,
        )
    current_slot = new_slots[0]lots)
    if match:
        # If we have a match, insert it and remove the matched
        # old slot from the pool
        insert_into_matches(
            match, current_slot,
        )
        old_slots.remove(match)
    # Continue with the next new slot
    return get_matches_from_slots(
        old_slots, new_slots[1:], matches=matches,
    )
```

proposals/utils/pdf_diff_sections.py
```diff
@@ -570,8 +580,14 @@ def __init__(self, obj, sub_title, proposal):
         self.proposal = proposal
 
     def make_row_for_field(self, field):
+
+        if field == "kind":
+            return KindRow(self.obj, self.obj.attachment, field)
+        else:
+            obj = self.obj.attachment
+
         if field in self.get_row_fields():
```
miggol, line 589: Not to wake sleeping dogs here, but why do we check if the requested field is in `self.get_row_fields()` here? Is this method called anywhere else than from `self.make_rows()`, which just loops over those exact fields?

proposals/utils/pdf_diff_sections.py
```diff
@@ -554,14 +556,22 @@ def get_row_fields(self):
         return rows
 
 
+
+
 class AttachmentSection(BaseSection):
```
miggol, line 561: I think creating a subclass here is a good way of doing things. You want subclasses to be stable at the interface, which is the case here.

This isn't as clean as it could be because of the `Row` class. `BaseSection` is nice to subclass because it has a generic interface. But `Row` is kind of annoying because it's purpose-built for Django models, without a generic interface, which makes you work around it.

So in the future my advice would be to keep the generic interface in mind from the start. So you would have a `BaseRow` where you define how other classes should talk to it, and then a `DjangoModelRow` that implements that interface. Even if you only ever use django model rows, your classes still become easier to talk to.

miggol (changes requested):
Just one major flaw in the match function

miggol (approved):
Looks good to me!

Issue #663, opened 2024-04-24
Problem with editing of pre aproved proposals by supervisors


miggol:
How to reproduce:

* Create and fill in a pre-approved proposal with a relation that requires a supervisor
* Submit the proposal
* Switch users to the selected supervisor
* Visit the decide screen and attempt to edit the proposal
* Go back to the first page of the proposal form
 
You will now be stuck, because the portal won't let you leave as long as you've selected yourself as the supervisor, which is forbidden. You can select the UD relation to remove the requirement, but that information would then be false. You can not select a different supervisor (the form does not permit this)

Todo:

* Check if this doesn't happen elsewhere (normal proposals, preassessments)
* Fix it. I don't expect this to be hard but it's late and I don't feel like starting this now.

tymees:
> Check if this doesn't happen elsewhere (normal proposals, preassessments)

It does, we got a bug report over mail. I think all proposal types are affected

PR #808, opened 2024-12-09
Feat/nonstandard proposals

miggol:
This adds the Attachments page to nonstandard proposals. It also somewhat changes the logic with which we construct non-standard proposals.

I have chosen to *not* define the preassessed/preapproved proposal flow in the "previous" checkers, i.e. where the next checker is returned. Now instead, we do the following:

```python3
def check(
        self,
    ):
        proposal = self.proposal
        if proposal.is_pre_assessment or proposal.is_pre_approved:
            return self.remaining_checkers()
```

So it's the trajectories checker that decides for itself that it is not needed for these proposals, and just continues. I think this is easier to keep track of. Wondering why WMO isn't showing up for preapproved proposals? Look in the `WMOChecker`, not the end of the Basic Details section.

DMP is also no longer recommended for preassessments and some intro text is added to the Attachments page.


EdoStorm96 (changes requested):
That's a nice idea, to have each checker evaluate the proposal type! I like it. There is still some stuff not working with this PR though.

- We have specific views for submitting of PreApproved and PreAss ... So I guess we should use those for submitting the alt paths.
- Although the stepper is working, the forward and back buttons are not yet configured correctly.
- I noticed that ProposalSubmit does not redirect back to translations. This is an unrelated error, but could you fix it here as well?

miggol:
Alright, I have addressed your comments.

I have also fixed that PreAssessments were not validated in ProposalSubmit. My blame diagnosis is that that specific if statement was there because of document validation, and the Embargo error checking was erroneously included inside of it after.

I will be addressing #663 in a separate PR.

miggol:
Thank you again for being thorough. Some time could have been saved if I had done a full check myself, which I have now done. Found some more stuff I missed

EdoStorm96 (changes requested):
Haha ok ... almost done here. The bug i outlined still persists. 

I should have explained it better. So if you start a new PreAss proposal, and you come to the first wmo page for the first time (/proposals/wmo/create/{pk}/pre/) and then you click the back button, you get an attribute error. Because WMOCreatePreAssessment, is trying to do PreAssesmentMixin.get_back_url(), but it doesn't have a self.object. All views inheriting from this method do have access to a get_proposal method, so just replace self.object.proposal with get_proposal() and we're good.

Issue #817, opened 2024-12-10
Server error https://fetc.tst.hum.uu.nl/reviews/show/12/


EdoStorm96:

EdoStorm96:
Mostly likely caused by some incomplete logic on a review action:
see [Sentry](https://uu-cdh.sentry.io/issues/13959581/?alert_rule_id=69106&alert_timestamp=1733840915788&alert_type=email&environment=major_4&notification_uuid=99fcf8cf-290f-419f-b66d-35555ecfaea6&project=4508008290123856&referrer=alert_email)

However, this also showcases that it is now possible to submit a proposal with `embargo = True` and `embargo_end_date = None` so fix this in the form as well.

Issue #814, opened 2024-12-09
Comment reviewer - no hard returns visible


djhcapel:
It would be very nice and helpful if hard returns take effect in the comments field of a reviewer. It will be easier to see lists for instance.

Issue #813, opened 2024-12-09
View all - no documents found


djhcapel:
This happens in application LC Test 1 - 20-001-02. Perhaps an 'old' application?

BTW: text in Dutch and English:

Application in general

Geen documenten gevonden.
The main trajectory

Geen documenten gevonden.

Issue #811, opened 2024-12-09
Length name trajectory


djhcapel:
The name of a trajectory now has max 15 characters; this should be 30 characters.

PR #818, opened 2024-12-10
Fix/post testing requests

EdoStorm96:
This branch features a lot of tiny issues, that have been uncovered following the testing of 4.0.

#789: Apparently the observation table still included fields from v1 observations and we never noticed. It now includes the new fields.

#809: The Stepper bubbles were untranslated ... My first hunch was to replace gettext with gettext_lazy and this works. Don't ask me why ...

#811: Increase the max length of trajectory names from 15 to 30.

#814: Include linebreaks in review comments for the Secretary to read comfortably.

No issue: Remove more remnant of the old AvailableURL system ...

#813: Untranslated text.

#817: A server errors caused by some shoddy validation and logic regarding embargo and embargo_end_date.


reviews/utils/review_actions.py
```diff
@@ -245,14 +245,17 @@ def is_available(self):
             return False
 
         if (
-            review.proposal.embargo == True
+            review.proposal.embargo_end_date is not None
```
miggol, line 248: As far as I know we don't reset `embargo_end_date` when an applicant sets `embargo` back to `False`. So there are potentially proposals with no embargo but yes embargo_end_date.

Therefore you must check both here.

reviews/templates/reviews/review_table.html
```diff
@@ -17,7 +17,7 @@
                 <td>{{ decision.reviewer.get_full_name }}</td>
                 <td>{{ decision.get_go_display|default:_("open") }}</td>
                 <td data-order="{{ decision.date_decision|date:'c' }}">{{ decision.date_decision|date:"j F Y, G:i" }}</td>
-                <td>{{ decision.comments }}</td>
+                <td>{{ decision.comments | linebreaks }}</td>
```
miggol, line 20: I never knew about this filter. Now I feel like an idiot for `<br />`ifying everything by hand in grant-tool. Good find

miggol (changes requested):
Awesome you got to all of this. Changes requested are minor

PR #820, opened 2024-12-12
Fix/more textual changes

EdoStorm96:
**Requires reloading fixtures!**

So this is all the textual changes and translations. Not much to say. It was a bit annoying, but I got it done :)

Best to review the translations using the translations commit.

Also, I think this branch will cause some merge issues with the migrations ... I could remove those from this branch and we can make migrations after merge ... I am not sure what migrations are currently in review.


miggol (approved):
Very nice! Just a minor textual comment so approving for now.

Awesome that you went through all of this.

Make sure when merging that the migration order is still good. Even if it's just text Django will get mad if the numbering/order of operations is wrong.

PR #821, opened 2024-12-16
Fix/archive access bug

### LINKED ISSUE #815: Archive - error message
If you press 'Archive' on the start page, you get the error message: "The page isn’t redirecting properly".

However, this does NOT happen if you log in as secretary!

EdoStorm96:
fixes #815.

A small PR for a change ;p

The mixin used for archive access check `HumanitiesOrPrivilegeRequiredMixin` was not raising a PermissionDenied. It's default behaviour is to ask for login again (which apparently was also not working on the test server). This is now fixed.

The archive button is also removed from the homepage if the user cannot access it.


miggol (approved):
Excellent!

Issue #796, opened 2024-12-05
HTML not marked safe in ProposalSubmit


miggol:
![image](https://github.com/user-attachments/assets/7187d258-de6b-4e4a-9440-5c6403afb4cf)

EdoStorm96:
It looks like this has been fixed by now, but will have to double check

PR #822, opened 2024-12-18
Fix/supervisor editing

miggol:
This should allow supervisors to edit proposals they are supervising, without errors about selecting themselves as the supervisor. This now also works for nonstandard proposals and proposals that have not yet been submitted.

Also an unsuccessful attempt at fixing #796

### DIFF
--- a/proposals/forms.py
+++ b/proposals/forms.py
@@ -103,7 +103,6 @@ class ResearcherForm(
         - If this is a practice Proposal, limit the relation choices
         """
 
-
         self.supervisor_editing_flag = False
         super(ResearcherForm, self).__init__(*args, **kwargs)
         self.fields["relation"].empty_label = None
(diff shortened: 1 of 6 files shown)


EdoStorm96 (changes requested):
As I said, there is still a problem with self.supervisor_editing_flag, as, thanks to SoftValidationMixin, clean() happens before the __init__ so self.supervisor_editing_flag is always False. You can add it as a form from the view though, and then pop it before the super __init__ in the form à la `UserFormModelKwargMixin`. Like so:

**ProposalResearcherFormView**
```
    def get_form_kwargs(self):
        kwargs = super().get_form_kwargs()

        if kwargs["instance"].supervisor == kwargs["user"]:
            kwargs["supervisor_editing_flag"] = True

        return kwargs
```

EdoStorm96 (approved):
Good!

PR #833, opened 2025-01-06
Fix/textual bugs

### LINKED ISSUE #829: Homepage: problem with the template?
Some template tags are visible on the homepage:

afbeelding

EdoStorm96:
fixes #829

So this fixes two bugs.

**The homepage trans tags:**
Djlint made the html inside the trans tags multi-line, so it required blocktrans tags.

**the inform_local_staff label:**
So ... the big one. I had a closer look at how another label was marked safe, namely `other_stakeholders` ... I've just gone and implemented this the same way. It is really ugly ... but it does work (for now ... seemingly ...), which is nice at this point!

So basically, you just override the label in the form and mark it safe there. This means this label exists in two places. The label in the model is just there for reference. This kinda sucks, but I guess this is how we did it before?

I found that the `other_stakeholders` label in the form was `mark_safe`'d twice, which AFAICT is redundant, so I removed this.

Let me know what you think :)

### DIFF
--- a/proposals/forms.py
+++ b/proposals/forms.py
@@ -795,6 +795,26 @@ class ProposalSubmitForm(
     class Meta:
         model = Proposal
         fields = ["comments", "inform_local_staff", "embargo", "embargo_end_date"]
+        labels = {
+            "inform_local_staff": mark_safe_lazy(
+                _(
+                    "<p>Je hebt aangegeven dat je gebruik wilt gaan maken van één "
+                    "van de faciliteiten van het ILS, namelijk de database, Zep software "
+                    "en/of het ILS lab. Het lab supportteam van het ILS zou graag op "
+                    "de hoogte willen worden gesteld van aankomende onderzoeken. "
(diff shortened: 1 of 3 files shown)


proposals/forms.py
```diff
@@ -816,11 +836,9 @@ def __init__(self, *args, **kwargs):
...
         super(ProposalSubmitForm, self).__init__(*args, **kwargs)
 
-        self.fields["inform_local_staff"].label_suffix = ""
-
-        self.fields["inform_local_staff"].label = mark_safe(
-            self.fields["inform_local_staff"].label
-        )
+        self.fields["inform_local_staff"].label = self.fields[
+            "inform_local_staff"
+        ].label
```
miggol, line 841: For me, this part is not necessary to fix the bug.

But it's also harmless.

miggol (approved):
Very glad you could crack this!

proposals/forms.py
miggol: The same goes for the `other_stakeholders` part, which never gave us any trouble. I have no rational reasoning but feel like keeping it in just to be safe.

PR #834, opened 2025-01-06
fix: avoid double validation of submit form

EdoStorm96:
So, the fix you suggested, @miggol:

When validating the submit form using the stepper, if only the submit form has an error, we don't add an extra error, as this can cause the form to get stuck, because it is unable to save.

Nothing fancy.

This does kindoff showcase why having `Stepper.get_form_errors()` return a list of dicts was a bit of a bad idea ... my bad hehe


proposals/forms.py
```diff
@@ -880,7 +880,14 @@ def clean(self):
...
 
         validator = Stepper(self.instance, request=self.request)
-        if validator.get_form_errors():
+
+        form_errors = validator.get_form_errors()
+
+        if form_errors:
+            # If there are only errors on the submit form, we override this
+            # validation. Otherwise saving the updated data becomes impossible
+            if len(form_errors) == 1 and form_errors[0]["page_name"] == _("Indienen"):
```
miggol, line 889: Works surprisingly well, but I would much rather have a test like

```python3
"proposals/submit" in error["url"].lower()"
```

The URL just feels a lot more dependable.

miggol (approved):
If matching the URL works just as well, I would prefer that. But otherwise approved!

PR #835, opened 2025-01-07
Temp/embargo and tests wip

EdoStorm96:
Make tests pass :)

Just some final issues to ensure major/4 can pass tests ...

Besides the stuff you did already, I made the PreAss and PreAppr proposals use the test_pdf.pdf file, when applicable, and solved the issue with adding a WMO object to PreAss proposals in the test db.


proposals/forms.py
```diff
@@ -832,7 +832,8 @@ def __init__(self, *args, **kwargs):
         self.request = kwargs.pop("request", None)
 
         # Needed for validation
-        self.final_validation = kwargs.pop("final_validation", None)
+        final_validation = kwargs.pop("final_validation", None)
+        self.final_validation = None
```
miggol, line 836: I think I intended to add a comment along the lines of

"we save this variable for later, but want it set for None for this part of the validation"

Because as it stands this is kind of confusing.

proposals/tests/proposal_submission_tests.py
```diff
@@ -77,6 +77,12 @@ def setup_proposal(self):
         Please note, this currently uses a mish-mash of both fixtures
         and programmatically created users and objects.
         """
+        self.wmo = Wmo(
+            status=0,
+            metc="N",
+            is_medical="N",
+        )
+        self.wmo.save()
```
miggol, line 85: This wmo must be implicitly getting used for the regular `self.proposal.wmo` without getting explicitly assigned. I don't know why (it's very confusing), but perhaps it would be better to add `self.proposal.wmo = self.wmo` down here:

```python3
    def setup_proposal(self):
        """
        Load our test proposals from a fixture, then add our user as
        an applicant to each of them.

        Please note, this currently uses a mish-mash of both fixtures
        and programmatically created users and objects.
        """
        self.wmo = Wmo(
            status=0,
            metc="N",
            is_medical="N",
        )
        self.wmo.save()
        self.proposal = Proposal.objects.get(pk=1)
        self.proposal.applicants.add(self.user)
        self.proposal.wmo = self.wmo
        self.proposal.save()
```

Right before the proposal gets saved. Just to make it explicit and less magic.

miggol (approved):
Approved with some minor requests.

PR #837, opened 2025-01-08
Feat/fetc filename in pdf

EdoStorm96:
So this is the simplest way I could think of. Just have another custom row, which receives a slot. It ain't pretty, but it does the job. The earlier implementation does become fully obsolete, as that also only worked for this specific case, so I removed this from the RowValue class.


miggol (approved):
Awesome, thank you for this.

PR #851, opened 2025-01-14
Fix is_new()

miggol:
This changes two things:

* The order with which flags like `slot.comparable()` and `slot.is_new()` are checked. Comparable now always comes first, and if it returns True then there is no more checking for is_new. So documents are no longer *both* comparable and new. This is decided in the template.
* `AttachmentSlot.is_new()` has been fixed. It returns True for every attachment in an unrevised proposal. It returns False for any attachment that is also attached to a parent proposal or study object. Finally it returns True for anything left over.

Please let this logic be right this time.


EdoStorm96 (approved):
Good good

PR #859, opened 2025-01-15
Feat/exclude submit from final validation

EdoStorm96:
Ok so, hear me out. To me this is a nice fix for removing the submit from the 'all errors' display on the submit page. It can also be reused in the submit form.

I know you did not want this, but consider it :)


miggol (approved):
> I know you did not want this, but consider it :)

Well I think your judgement has rightly prevailed here. Can't argue with the fact that it just works and works well.

Issue #855, opened 2025-01-14
Legacy documents - vergelijkingsknop onnodig and legacy documents visible for secretary, unnecessary?


djhcapel:
I have submitted a revision (25-002-02) and noticed two things:
1. If I log in as secretary, I still see the legacy documents. But I do not need to see these, do I? The legacy documents are only there for the researcher, when submitting the (next) version, right?
2. There is a vergelijkingsknop for each legacy document, but obviously, there are no changes. Can this vergelijkingsknop be deleted?

miggol:
As discussed, (1) is actually necessary because it's important to be able to see legacy documents for past proposals. For example, if examining the review process of an application from 2024.

As for (2), I understand that the feature being requested is that the compare icon is not shown for legacy documents that have no changes. This is a significant feature to implement for a documents system that is no longer in use. I would suggest that we don't do this, as long as @djhcapel agrees

djhcapel:
Perhaps there is a misunderstanding here. Legacy documents are the documents from the previous version, say version 01, and they appear as legacy documents in version 02. What does the compare icon do here? Does it do anything? I thought no, because the legacy documents have remained the same. Or does the compare icon lead to comparison of the legacy document X from version 01 to the new document X of version 02? But that is not possible, right, because we do not know which document from 01 to compare to which document from version 02? So I thought that the compare icon did not have a function for the legacy documents and could be deleted to avoid the expectation that you actually can get a comparison. Discuss via Teams?

miggol:
The compare icon does the exact same thing that it always did in the previous version. It allows you to view differences between (legacy) documents from one review to the next. If we were to remove it, and someone in the committee wanted to see the changes between versions of an old proposal, that link would be missing.

In the previous version, the compare link would always show up for theoretically comparable documents, even if they were exactly the same. I just now found a way to not display the icon for documents that are exactly the same, which I have implemented in PR #860 . If this PR is merged, then the compare icon will no longer show up for identical legacy documents. Because all future legacy documents should never change, this compare link will be hidden for all future reviews, but not for old reviews where the documents had changes.

miggol:
This has been discussed in person to the point where I believe this can be closed now.

PR #860, opened 2025-01-15
feat: Do not offer to compare files with identical urls

miggol:
Apparently the comparison issue in #855 is surprisingly fixable. This is a bit of a dangerous change to be making right now, so I'm open to discussion on if we should fix this at all.

This PR presents the fix in question that I'm carefully optimistic about. But please try and break it.


EdoStorm96 (approved):
Nice one! I could not fault it

PR #868, opened 2025-01-16
fix: exclude_submit_bug

EdoStorm96:
This fixes a breaking bug for the new exclude_submit option I added to Stepper.get_form_errors.

If this if statement receives StepperItem's without a form_class attribuge eg. ContainerItem's (used when there are multiple studies), an AttributeError gets raised. This fixes it.

### DIFF
--- a/proposals/utils/stepper.py
+++ b/proposals/utils/stepper.py
@@ -262,7 +262,7 @@ class Stepper(renderable):
                 # When using this method on the submit page, we
                 # use the form's own validation, otherwise it gets
                 # validated twice
-                if exclude_submit and getattr(item, "form_class", None) == ProposalSubmitForm:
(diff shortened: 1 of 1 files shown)


proposals/utils/stepper.py
```diff
@@ -262,7 +262,7 @@ def get_form_errors(self, exclude_submit=False):
                 # When using this method on the submit page, we
                 # use the form's own validation, otherwise it gets
                 # validated twice
-                if exclude_submit and item.form_class == ProposalSubmitForm:
+                if exclude_submit and getattr(item, "form_class", None) == ProposalSubmitForm:
```
miggol, line 265: Test for equal identity of class types should be `x is some_class` not `x == some_class`. The latter is for eqaulity of value and is not well-defined for Forms.

We missed both these issues on the first pass, rip

miggol (approved):
Approved with one correctness suggestion

Issue #864, opened 2025-01-15
Minor issue - edit legacy documents - not return to applicaton but to home page


djhcapel:
In order to test the compare icon, I edited the legacy documents, and chose a different file for one of the documents. After pressing 'change forms' I did not return to the overview page of that application (23-001-04) but to the home page. No major problem as this will rarely happen. And if it happens, it will be to Katharina or myself.

Issue #838, opened 2025-01-08
Implement 404 rather than a system error on missing attachments


miggol:

EdoStorm96:
This is the case currently/has already been fixed, right?

EdoStorm96:
No, it was not! but it was similar to #726

Issue #800, opened 2024-12-05
Edit start date - button 'save'


djhcapel:
Very nice we can do this as secretaries! 

If I press 'edit start date', the page 'Edit start date' appears. I can adapt the date, but I thought I could not save it, because the button also says: 'Edit start date'. I realize this is similar to the current version of the portal (e.g. 'documenten aanpassen'), but I'd rather have this button just reads 'Save' (and if applicable, also for the 'Change documents', sth like just 'Save'.

Issue #727, opened 2024-10-23
Validation error appears twice


tijmenbaarda:
When the field "Students (who do person-oriented research as part of their program) do, in principle, not need to submit a proposal with the ethical committee..." is left empty and the form is saved, the validation error "This field is required" appears twice.

miggol:
Worth looking into, hopefully softvalidationmixin can fix this. But otherwise not a priority right now

tijmenbaarda:
Here the same happens:
![image](https://github.com/user-attachments/assets/e532117d-8e0f-4633-a534-97c0228c37a6)

EdoStorm96:
I could not reproduce the first one. For the second one, it seems that the field simply gets validated twice. One time as a standard django procedure, and once by us.

PR #869, opened 2025-01-16
Fix/tiny bugs

EdoStorm96:
Bugfixes for tiny bugs and issues, mostly out of boredom/ doing housekeeping.

#727:
Study.recruitment got validated twice in the form, which caused the double error message.

#726 & #838:
When creating objects that have the pk of a related object in the url, it causes an DoesNotExist when loading these with non-existant pk's. I fixed this with a try/except.

#864:
Change the succes url for editing legacy docs

#800:
Changed text on buttons, pretty much only seen by the secretary. Gotta please the boss ;p


miggol (approved):
Nice work adding all those 404s. No changes requested just some general comments

PR #870, opened 2025-01-20
Fix/404 on attachment download

### LINKED ISSUE #838: Implement 404 rather than a system error on missing attachments

miggol:
Closes #838 

Not only would `AttachmentDownloadView` previously not return a 404, it also allowed the downloading of any existing attachment given any proposal that the requesting user had access to. Because permissions are checked based on the given proposal, this was a security issue.

To fix this, `Attachment.get_owner_for_proposal()` now actually does something (for proposal attachments) and returns `None` if the given attachment does not have a valid owner for the given proposal. The download view will raise a 403 (`PermissionDenied`) if no owner can be found.

`AttachmentDownloadView` now only gets attachments and proposals from their respective `self.get_` methods, which can each raise a 404, and also cache their object so they only produce one query.

### DIFF
--- a/attachments/models.py
+++ b/attachments/models.py
@@ -4,7 +4,6 @@ from django.utils.translation import gettext as _
 from django.urls import reverse
 from main.utils import renderable
 
-
 from cdh.files.db import FileField as CDHFileField
 
 # Create your models here.
@@ -161,6 +160,6 @@ class StudyAttachment(
         """
         try:
             owner = self.attached_to.get(proposal=proposal)
-        except self.DoesNotExist:
+        except self.attached_to.model.DoesNotExist:
             owner = None
         return owner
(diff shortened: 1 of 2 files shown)


attachments/models.py
```diff
@@ -153,6 +156,11 @@ def get_owner_for_proposal(
...
     ):
         """
-        Gets the owner study based on given proposal.
+        Gets the owner study based on given proposal, or None
+        if no such study exists.
         """
-        return self.attached_to.get(proposal=proposal)
+        try:
+            owner = self.attached_to.get(proposal=proposal)
+        except self.DoesNotExist:
```
EdoStorm96, line 164: Should be Study.DoesNotExist. Otherwise this try except does not catch the expected error.

EdoStorm96 (approved):
Nice work! Just the minor fix and we're good!

EdoStorm96 (approved):
Yes this works!

PR #879, opened 2025-02-06
fix: make date format alphabetically sortable

EdoStorm96:
Thought I would just tackle this as a little snack.

So, I just converted the date format in the export view to be alphabetically sortable (YYYY-mm-dd) See the [docs](https://docs.djangoproject.com/en/5.1/ref/templates/builtins/#date) for more info ...

Think this would be nice to deploy as a little treat for the secretary ;p


miggol (approved):
So thoughtful, what a pleasant surprise

IMO this can go straight to acc

PR #890, opened 2025-02-19
Suggestion: deny supervisor editing on certain views

miggol:
It's caused us enough pain already. Does it really matter if supervisors can't edit the pages that cause major headaches for us?

### DIFF
--- a/locale/en/LC_MESSAGES/django.po
+++ b/locale/en/LC_MESSAGES/django.po
@@ -6,8 +6,8 @@ msgid ""
 msgstr ""
 "Project-Id-Version: \n"
 "Report-Msgid-Bugs-To: \n"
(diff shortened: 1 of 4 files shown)


EdoStorm96 (commented):
I think this implementation is perfectly sound and quite nice :)

I am kindoff on the fence about whether we want this though ... On the one hand, yes ... but on the other hand, if a supervisor just wants to add an applicant or stakeholder, this would be a bit annoying ... So to me, having the mixin on the researcherform makes sense ... but on the other researcher form it can be annoying. So what do you think about removing the mixin there, and we just rely on my fix in #889 to avoid chaos? ;p

miggol:
No longer relevant

Issue #877, opened 2025-01-31
Documents - Recordings & public interest - no request for 'agreement' document


djhcapel:
See #753. I just checked the situation of 'public interest' and audio/video recording, but on the Documents page, no slot for the 'agreement recordings' document becomes available.

I know this could happen for an intervention study, since 'recording' is not a specific field but a general text field, but I assumed the 'agreement recording' would be suggested (if not mandatory) in all other cases, i.e. observation and task/interviews?

There is a discussion now with the privacy officer about the necessity / desirability of such agreement document, so no need for immediate action, but I'd like to know if the document could at least be optional on the Documents page, or that we can only refer to this document only at the revision stage.

djhcapel:
Discussed this last week in the Linguistics Chamber and we concluded that the document 'Agreement - legal basis Public interest - video/audio recordings' should in principle be uploaded unless there are good arguments not to do so. So, if 'audio recording' OR 'video recording' is ticked in Obervation OR 'Task/Interviews', this document should be listed as optional document.

Under 'meer info':
Als er audio- en/of video-opnames gemaakt worden, en de AVG-grondslag is 'algemeen belang', dan moet dit document ingediend worden. Als je dit document niet indient, dan moet je de reden hiervoor beschrijven in het laatste vrije-tekstveld.

djhcapel:
See also #733.

EdoStorm96:
I'm pretty sure this issue is deprecated, but am waiting to discuss this with the secretary

djhcapel:
See also #904; I think we need to discuss this further, because this document is required. 

The question about recordings was a free text field under 'Intervation' I think, whereas it was a tick box for 'observation' and 'tasks/interviews' and that this caused difficulties in requesting the document 'Agreement for recording' on the Documents page?

EdoStorm96:
Currently we have this. if public interest and there are recordings (In this case the study has both adults and children):

![Image](https://github.com/user-attachments/assets/8614354c-df76-4e66-ae4f-4ffe336f8cb0)

I feel like that is sufficient for what you are asking for here?

djhcapel:
Since #904 has been merged, can this issue be closed?

PR #904, opened 2025-03-24
Feat/registration question

EdoStorm96:
Okay, so this is a basic version of creating a unified registration question for Studies. I just took the Registration and Restrationkinds fields and fixtures from tasks, as these were the most detailed. I think this system will work fine for any kind of study. I think this is quite nice, but there is some things to consider. Mainly, do we want to attempt a data migration? I think it would be pretty ugly, as you'd have to get the migrate order right to make it work ... So I would prefer asking researchers to fill it in again. Although this is also annoying for sure ... But we cannot get a perfect data migration anyway.

I also looked again at #877, as I was planning to include this. But I feel like this issue is deprecated, as we have plenty of attachmentkinds to ask about recordings. I discussed this with the secretary, but she did not get back to me yet. If she insists we need another attachmentKind for registration, I will add that to this PR.

Good luck reviewing this :) The migrations and stuff are pretty funky, but the result is much better imho.


EdoStorm96:
Ok, just realizing I missed some stuff. Back to work.

miggol:
> Having some mention of the registration (or Personal Data info for that matter) on StudyEnd.

This is more than worthy of its own issue which the secretary may then prioritize

miggol (changes requested):
It seems that after your first dramatic realization this is quite complete. There are some things I still found to be missing:

- The table in `task_list.html` still references registrations and messes up the table.

- We need an explicit "no registration" option in the fixtures.

### Data migration
I think that we are throwing away a lot of data here. I'm not worried at all about researchers having to re-enter some information for a running application. But We're getting rid of all kinds of historical information by just deleting these models and fields without any kind of data migration.

Getting the order of migrations right isn't that big of a deal because you can just specify the dependencies of each migration. But even easier would be to do it in multiple steps.

1. Add the new model
2. Makemigrations
3. Create new blank migration in the new Registrations app (so it will depend on the new model migration)
4. Write and test the data migration
5. Remove the old models and fields
6. Makemigrations again

That should set the order of operations correctly. 

> removing the measurement field from Intervention ... I feel like the registration question replaces this. But not entirely sure.

Because this was unstructured data I feel less bad about tossing it. But you could also stick this information in the `registration_details` field for "Otherwise...".

Of course, the so far unspoken alternative to all of this is to just leave the old models and fields to rot until we're confident we won't need the data / forever. Not a good solution but to me slightly better than throwing everything out immediately.

studies/migrations/0034_registration_study_registration_kinds_details_and_more.py
miggol, line 1: These migrations are a good place to load fixtures IMO. Or create a new blank migration after this one that loads in the initial data.

EdoStorm96:
Oke, there was one more bug with the migrations, but all good now.

studies/fixtures/00_registrations.json
```diff
@@ -128,10 +128,42 @@
...
+        "description": "geen registratie",
+        "description_nl": "geen registratie",
+        "description_en": "no registration",
+        "is_local": false,
+        "needs_details": false,
+        "needs_kind": false,
+        "requires_review": true,
+        "age_min": null,
+        "is_recording": false
+    }
```
miggol, line 160: A nice to have would be an `excludes_others` boolean for this registration in particular. It's a bit strange that you can select "no registration" and other registrations at the same time.

studies/migrations/0035_load_registration_fixtures.py
```diff
@@ -0,0 +1,25 @@
...
+
+# Custom migration to load new fixtures for Registration model
+
+
+registration_fixtures = "00_registrations", "01_registrationkinds"
+
+
+def load_registration_fixtures(apps, schema_editor):
+    for fixture in registration_fixtures:
+        call_command("loaddata", fixture, app_label="studies")
```
miggol, line 14: Calling external commands here isn't great. Migrations are supposed to remain stable through time, but here we're relying on the `loaddata` command and specific fixture names. 

If in the future we add new fields to the `registration` models and update these fixtures to include that field, this migration will break, because it doesn't know about that future field yet.

On the other hand we *do* have historical models available, so we can just add instances as python objects without worrying about future updates breaking. So I would prefer that we add them either through python:

```python
audio = Registration(
    pk=1,
    order=1,
    description="audio_opname",
    ...
)
audio.save()
```

Or by dumping the JSON directly into this migration as a multi-line string (I'm serious). Use the JSONL format from the `dumpdata` command to do this in a more succinct form.

studies/migrations/0036_translate_old_registrations.py
```diff
@@ -0,0 +1,169 @@
...
+    1: 8,
+    # audio-recording
+    2: 1,
+    # video-recording
+    3: 2,
+    # notes
+    4: 9,
+    # other
+    5: 10,
+}
```
miggol, line 28: Using the object pk in this dict is a bit scary, because we don't know for sure what's in the database.

I think that in this case it would be better to match based on description string. It's self-documenting, and on the FROM-side it provides an assertion that we're working with the right data. We can just fail if a description is not found in the table.

On the TO-side using pk's is fine though, we just defined those so we can be pretty sure about them.

miggol (changes requested):
Top notch work on the data migration. I really like how the details table translates into the other field. Sorry this review took ages, I forgot about it.

I just don't agree loading the external fixture file like this.

studies/fixtures/00_registrations.json
EdoStorm96: Mmmm yes, that would be nice ... But implementing this functionality on a widget and model level is a whole different story ... so I'm leaving this be.

EdoStorm96:
Alright, I've brought this branch up-to-date with develop and addressed your comments. It still is working fine. Using the description for the observation registration is a little bit more clunky in practice, as I had to load the model into the migration, but it's not a huge deal. Seems to work fine!

The nice to have would be quite a bit of work in practice I think ... So I'll leave that be for now.

EdoStorm96:
Alright so I tested your scenario and you are indeed right. Nice catch! I did find however that mock has become part of the python standard library. You can do `from unittest.mock import patch` and then 
```
    with patch(
        "django.core.serializers.python.apps",
        apps,
    ):
```
Works fine. And then we can keep the requirements as they were.

Could you change this? Then I think we're all good!

PR #906, opened 2025-03-25
Fix/remove ldap support

### LINKED ISSUE #599: Remove LDAP support
We've migrated to SAML, all things that are still using LDAP are broken, so let's just rip it out completely

EdoStorm96:
fixes #599 

Long overdue, as this was pretty simple actually. I think it should all be fine. Let's see what the tests say ...


miggol (approved):
Nice! Including a requirements bump, I would recommend merging and deploying this to tst right away to see if everything still works on a HumIT server.

Hold off on merging other things into develop while doing so

PR #907, opened 2025-03-31
Merge "develop" into "acceptation"

EdoStorm96:
Alright, so develop is live on tst. It seems to work perfectly fine with LDAP support removes, so I'd say let's merge this to acc.


EdoStorm96:
Imma merge this.

PR #912, opened 2025-04-17
fix: update qs to exclude supervisor reviews

EdoStorm96:
Just a small bugfix, so that the in_revision list view does not exclude reviews, with children that are supervisor reviews.


miggol (approved):
Approved and merging to acceptation

PR #913, opened 2025-04-22
update docs

RoordinkFrank:
update with load_fixtures command and clarified required python3-dev …

### DIFF
--- a/docs/developing/install/steps.rst
+++ b/docs/developing/install/steps.rst
@@ -7,6 +7,11 @@ Installation of this Django project is quite straightforward:
 OS Dependencies
 ---------------
 
+It is assumed you have:
+
++ python3 (Server runs on python 3.9)
++ python3-dev (python3.9-dev when using python3.9, relevant when running more then one python version on your computer)
+
 Your host OS needs some packages, below is a list of debian packages:
 
 + libtiff5-dev
@@ -67,6 +72,16 @@ Run the database migrations:
 .. note::
     This will also create a sqlitedb file if it's missing
 
+Setting up the fixtures.
+There are two ways to do this.
+
+.. code:: shell
+
+    python manage.py load_fixtures
+
+If this works you can skip to create a superuser.
+If for some reason the command does not work you can try the old way to load the fixtures
+
 Load all fixtures using ``python manage.py loaddata``.
 
 This command requires you to specify each fixture file, as it doesn't auto detect them.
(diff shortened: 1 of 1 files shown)


docs/developing/install/steps.rst
```diff
@@ -117,6 +132,17 @@ You can set this is the admin interface.
...
+   In a non-production enviroment it is advised to change the email in the settings.py file to
+        # E-mail settings
+        EMAIL_BACKEND = "django.core.mail.backends.filebased.EmailBackend"
+        EMAIL_FILE_PATH = "email/"
+        # EMAIL_HOST = "localhost"
+        # EMAIL_PORT = 2525
+        EMAIL_FROM = "T.D.Mees@uu.nl"
+        EMAIL_LOCAL_STAFF = "T.D.Mees@uu.nl"
+   Emails will now be send to a local directory instead of generating an error.
+
```
miggol, line 145: Markup in this paragraph isn't exactly right. Take a look at the version that Github rendered here: https://github.com/DH-IT-Portal-Development/ethics/blob/f728ab7e9cd097159f651e13a315495ae9da0170/docs/developing/install/steps.rst

It appears you can't place a code block inside a tip, or at least not like this. Check out [the RST primer](https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html) for a quickstart on how to write RST.

It's better to build the docs locally and check everything is correct before committing and pushing. See the top-level README.md on how to build the docs. You can also use a tool like `restview` for live updates while editing RST.

miggol: Second, I would recommend putting the above email suggestion in `debug_settings.py` rather than in `settings.py`. The former is included in the global `.gitignore` so that the changes in there never get pushed and don't get lost when switching branches.

docs/developing/install/steps.rst
```diff
@@ -67,6 +72,16 @@ Run the database migrations:
 .. note::
     This will also create a sqlitedb file if it's missing
 
+Setting up the fixtures.
```
miggol, line 75: This could be a heading.

docs/developing/install/steps.rst
```diff
@@ -67,6 +72,16 @@ Run the database migrations:
...
+There are two ways to do this.
+
+.. code:: shell
+
+    python manage.py load_fixtures
+
+If this works you can skip to create a superuser.
+If for some reason the command does not work you can try the old way to load the fixtures
```
miggol, line 83: It isn't very clear what exactly are the two ways to load the fixtures are, and where the instructions of both methods start and end. You wouldn't want to load the fixtures twice!

Perhaps try to make it more clear with headings

docs/developing/install/steps.rst
```diff
@@ -7,6 +7,11 @@ Installation of this Django project is quite straightforward:
 OS Dependencies
 ---------------
 
+It is assumed you have:
+
++ python3 (Server runs on python 3.9)
++ python3-dev (python3.9-dev when using python3.9, relevant when running more then one python version on your computer)
+
```
miggol, line 14: Great inclusion

miggol (changes requested):
Congrats on the first PR!! I would like to request some changes :)

miggol (approved):
Presuming the heading fix we discussed is included, I approve!

Issue #894, opened 2025-02-24
Textual change - page 'Informatie over financiering'


djhcapel:
Here, 'toestemmingsbrief' should read 'goedkeuringsbrief'. Not sure if it appears somewhere else as well?
This is the letter of approval that is sent to an applicant once the application has been approved by the FEtC-H.
There is also a template 'Informatiebrief toestemming', but that is a letter for participants which they should receive when their personal data are processed on the legal basis 'consent/toestemming'.

PR #914, opened 2025-04-22
"toestemming" naar "goedkeuring" veranderd bij proposals/funding.

RoordinkFrank:


RoordinkFrank:
Does this also have the same translation problem?

miggol (changes requested):
Hi, this PR also has a fuzzy translation.

```gettext
#: proposals/models.py:295
#, fuzzy
#| msgid ""
#| "De titel die je hier opgeeft zal in de formele toestemmingsbrief gebruikt "
#| "worden."
msgid ""
"De titel die je hier opgeeft zal in de formele goedkeuringsbrief gebruikt "
"worden."
msgstr "This title will be used in the formal letter of approval."
```

In this case, I would say that the English string is still fine and just needs the fuzzy tag removed.

The reason for this is that although Django will detect these strings being similar and mark them as fuzzy versions of each other, it won't actually use this translation until a human confirms that they are in fact correct translations of each other. Presumably because the Fuzzy algorithm gets it wrong more often than right.

RoordinkFrank:
Issue #894 gemerged in this pullrequest. This request now has #894 and #908

tasks/models.py
```diff
@@ -124,7 +124,7 @@ class Task(models.Model):
     )
 
     repeats = models.PositiveBigIntegerField(
-        _("Hoe vaak wordt deze taak uitgevoerd binnen deze sessie?"),
+        _("Hoe vaak wordt deze taak uitgevoerd binnen deze sessie (per deelnemer)?"),
```
miggol, line 127: I think this is overly terse wording. How about: "Hoe vaak voert een deelnemer deze taak uit binnen deze sessie?"

The English translation could be "How often will a participant perform this task within this session?"

miggol (changes requested):
Nice work! Translations are working :)

I do in fact have a comment about the specific wording. Sorry, I could have caught this in my previous review. But I really do think the wording is confusing, while the whole point of the issue was to make things less confusing.

miggol (approved):
Excellent, good work!

PR #915, opened 2025-04-24
model aangepast zoals gevraagd.

RoordinkFrank:


miggol:
Again, we don't know which issue this references :)

Also needs translations. I will get back to you on Tuesday

miggol:
After running `python manage.py makemessages -a`, there is a fuzzy translated string in `django.po`, which should be fixed. Take a look at `i18n.rst` over in docs to figure out how to do that!

Issue #893, opened 2025-02-21
PDF - order of two questions should be the same as in the portal


djhcapel:
I just noticed that in the portal, the question "Give a clear, concise description of the research question or questions. Use a maximum of 200 words." is followed by the question: "What are the most important ethical considerations in this research and please describe briefly how you will address them. Use a maximum of a 1000 words." However, in the pdf, the order of these two questions is the other way around, which is not logical. 

Could these two questions be listed in the pdf in the same order as in the portal?

PR #916, opened 2025-04-24
PDF Order veranderd zoals gevraagd.

RoordinkFrank:


miggol:
Even with a small PR that doesn't merit a wordy description, it's nice to mention the issue number. In this case that would be #893, github will link (and close it automatically on merge) it automatically if you say something like "Fixes 893"

miggol (approved):
Very nice!

PR #917, opened 2025-04-28
Removed Sphix auto build files from remote

RoordinkFrank:
Added docs/reference and docs/_build to gitignore as it's part of the Sphix auto build which were to be removed from remote. Also untracked currently existing files that already were on remote with the git rm --cached reference/ -r commando.


miggol (approved):
Looks good! I approve

Issue #878, opened 2025-02-04
Automatic mail long route - textual change


djhcapel:
In the automatic mail for long-route applications, it says something like ‘zeer kort’ after the meeting, comments will be sent. This 'very short' can be deleted: just that comments will be sent after the meeting.

PR #922, opened 2025-05-09
Fix/automatic mail revision

RoordinkFrank:
Issues #683 and #878

### DIFF
--- a/reviews/templates/mail/submitted.txt
+++ b/reviews/templates/mail/submitted.txt
@@ -1,7 +1,7 @@
 {% extends "mail/base-internal.txt" %}
 
 {% block content %}
-Er is een nieuwe aanvraag ingediend (referentienummer: {{ review.proposal.reference_number }}, eindverantwoordelijke: {{ review.accountable_user.get_full_name }}, kamer: {{ review.proposal.reviewing_committee.name }}).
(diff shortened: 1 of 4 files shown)


reviews/templates/mail/submitted.txt
```diff
@@ -1,7 +1,7 @@
 {% extends "mail/base-internal.txt" %}
 
 {% block content %}
-Er is een nieuwe aanvraag ingediend (referentienummer: {{ review.proposal.reference_number }}, eindverantwoordelijke: {{ review.accountable_user.get_full_name }}, kamer: {{ review.proposal.reviewing_committee.name }}).
+Er is {% if was_revised %}een revisie{% else %}een nieuwe aanvraag{% endif %} ingediend (referentienummer: {{ review.proposal.reference_number }}, eindverantwoordelijke: {{ review.accountable_user.get_full_name }}, kamer: {{ review.proposal.reviewing_committee.name }}).
```
miggol, line 4: Nice touch

miggol (changes requested):
The HTML and plaintext versions of the emails are separate files. It would be nice if they were both based on the same plaintext, but sadly that is not the case and the HTML emails need to be edited as well.

You can search the project for "zeer kort" to find the email HTML files.

PR #927, opened 2025-05-13
fix/several_small_bugs

### LINKED ISSUE #874: Download attachments with FETC-filename by default?
The (adjunct) secretaries would like to have the 'old' file names when you download files from the left menu. We now get files with the names provided by the applicants, which are not always informative. We would, however, like to keep the double names in the View all. 

Is there a reason why downloaded files from the left menu keep the name provided by the applicant?

### LINKED ISSUE #919: Error with logic of student_context_details in PDF
So there was some incorrect logic in the student context related logic in the PDF ... It is now:

_pdf_diff_sections.py, line 85_
```
        if not obj.relation.check_in_course:
            rows.remove("student_program")
            rows.remove("student_context")
            if obj.student_context is not None:
                if not obj.student_context.needs_details:
                    rows.remove("student_context_details")
            else:
                rows.remove("student_context_details")
            rows.remove("student_justification")
```
this is incorrect, because it still keeps student_context_details around when it should not be there. It only checks for this when check_in_course is False, but it should check for this when check_in_course is True.

Something like this would fix it:
```
        if not obj.relation.check_in_course:
            rows.remove("student_program")
            rows.remove("student_context")
            rows.remove("student_context_details")
            rows.remove("student_justification")
        elif not obj.student_context.needs_details:
            rows.remove("student_context_details")
```

### LINKED ISSUE #925: Documents page English - still some text in Dutch
Image
See the Documents page in English for "research approved elsewhere". This page still contains some text in Dutch:

- Aangeraden
- Voeg optioneel bestand toe aan aanvraag.

We just had an applicant who had not uploaded some required documents, perhaps because of this.

### LINKED ISSUE #926: Supervisor reviews appear on InRevisionApiView
This is a bug that the secretary mentioned a while ago ... But it basically comes down to Supervisor reviews that have been closed, due to some complicated logic, sometimes appear on the In Revision Review list.

The fix for this is a on liner. Basically, we need to use the is_comittee_review attribute in the filter for this list.

EdoStorm96:
This PR addresses a few minor issues.

fixes (not) #926:
Because of an old unupdated QuerySet, supervisor reviews appeared on the In Revision Review list view.

fixes #874:
Ensure that the FETC filename is the default when downloading attachments.

fixes #919:
I noticed the logic for making student_context_details appear in the PDF was incorrect ... So I rewrote this.

fixes #925:
There was some stuff that wasn't translated for documents. One string in a template simply had no trans tag and for attachment.kind.desiredness, we had to use gettext_lazy instead of gettext to make it work.

Each commit of this branch adresses a seperate issue in the same order as written above, so best to review it per commit. Good luck!


EdoStorm96:
I am now realizing that the fix for #926 is already on acceptation ... So reviewing it here is kindoff redundant. Sorry, a bit unorganized of me to reuse this branch ... I didn't realized this was already merged with acc.

proposals/utils/pdf_diff_sections.py
```diff
@@ -86,12 +86,10 @@ def get_row_fields(self):
         if not obj.relation.check_in_course:
             rows.remove("student_program")
             rows.remove("student_context")
-            if obj.student_context is not None:
```
RoordinkFrank, line 89: if false:
     do something
else (so if true)
    do something

it is slightly cleaner to read if as
if true
     do something
else
     do something

RoordinkFrank (changes requested):
if false:
do something
else (so if true)
do something

it is slightly cleaner to read if as
if true
do something
else
do something

RoordinkFrank (approved):

PR #928, opened 2025-05-14
chore: Requirements bump for acc

miggol:
It's been a while, so best to update requirements before we move the train along to prod.


EdoStorm96 (approved):
Good call! I think this should all be fine.

PR #931, opened 2025-05-15
Fix/question about consultation data manager

RoordinkFrank:


RoordinkFrank:
if cleaned_data["data_manager_conversation"] is None: voelt redunant aan. Is er een manier om het netter op te schrijven of is dit wat het is?

EdoStorm96 (changes requested):
Ok, so good work! You thought off everything, even the PDF, which I often forget ... I just have a few comments, so have a look at that.

In general, I am wondering if this is the correct po file for this branch, as it looks a bit funky. Maybe that partly has to do with the unresolved merge conflicts. But you also forgot to add translations to the new texts.

Then a more subjective thing ... I think it is nicer to just have one migration for a PR like this. Often you have multiple migrations as you are working on a branch, but then for the PR, I like to reset the migrations and regenerate all the changes into one migration.

In this case you can just do:
```
python manage.py migrate proposals 0059
```
This resets your db to migration 0059 of proposals. Then delete your migrations. It is important to do it in this order! 
And then:
```
python manage.py makemigrations
```

locale/en/LC_MESSAGES/django.po
```diff
@@ -1889,6 +1889,27 @@ msgstr ""
 
 #: proposals/models.py:451
 msgid ""
+"Ik heb mijn Data Management Plan (DMP) besproken met de data manager van de "
+"Faculteit Geesteswetenschappen."
+msgstr ""
+"I discussed my application and Data Management Plan (DMP) with the data manager of Faculty of Humanities"
```
EdoStorm96, line 1895: Ok, so tiny tiny nitpick. I would prefer to have "Faculty" and "Faculteit" lowercase. I know this contradicts Desiree's text in the issue, but I think, in general, we write it like this. See [this](https://www.uu.nl/organisatie/faculteit-geesteswetenschappen/over-de-faculteit) page for example.

proposals/forms.py
```diff
@@ -758,18 +758,33 @@ def clean(self):
...
+            if cleaned_data[field] is None:
+                if (
+                    field is "privacy_officer_conversation"
+                    or field is "data_manager_conversation"
+                    or field is "research_data_management_conversation"
+                ):
+                    self.add_error(
+                        field,
+                        _("Dit veld is verplicht om verder te gaan."),
+                    )
```
EdoStorm96, line 788: Alright, so this can still be improved. In my opinion, it is not really necessary to anticipate the adding of more fields to this form. I think this is a bit over-engineered. I think **if** a new field were to be added to this form (which I do not see happening any time soon), that needs a different cleaning solution, we can fix it at that time. We don't have to plan ahead for a fairly unlikely scenario in this way. So for me, just removing lines 779-783 is my preferred solution. 

However, if you were to insist we need to check that this loop only covers the specific fields we desire, there is a few ways we could improve this. Since you have already hardcoded the strings, it makes sense to just create a list with these strings to work with. Doing this, we could replace lines 779-783 with:
```
if field in ["privacy_officer_conversation", "data_manager_conversation", "research_data_management_conversation"]:
```
Then running black, would make it look a bit nicer. `if variable in list:` is a very useful python pattern, so it is good to know.

But then, if we already have the list in place, we might as well just loop over that. So then the best way would be:
```
for field in ["privacy_officer_conversation", "data_manager_conversation", "research_data_management_conversation"]:
    if cleaned_data[field] is None:
        self.add_error( 
            etc.
```
So again, I say just loop over `self.fields`. But if you must specify the fields, the solution above is cleaner.

EdoStorm96 (changes requested):
Alright, this is almost there :) Good job!

Just two more tiny changes I'd like to see.

Issue #875, opened 2025-01-30
Documents - incorporate link to templates


djhcapel:
Members of the FEtC-H suggested to include links to templates on the Documents page. That would indeed be very user-friendly. 

Is this possible? And if so, how would you like to get the URLs, just as a list of templates plus the correct links?

RoordinkFrank:
@djhcapel Can you elaborate? You wish to include links on proposals/attachments that link to templates?
I assume those templates are for example an example for Data Management Plan or a fillable pdf.

Are the templates already online somewhere or do they need to be stored on the server?

djhcapel:
The idea is to have links to particular documents, e.g. 'Informatiebrief toestemming' to that document at https://intranet.uu.nl/kennisbank/documenten-ethische-toetsingscommissie-gw. I could provide you with a list of document names and corresponding links if that is convenient. Perhaps we could also link to the template for DMPs, but I will need to discuss this further.

djhcapel:
All researchers who submit applications should hold a PhD and are therefore employees of Humanities and thus have access to intranet, so this is not a problem.

The link cannot be added to the names of the documents? That would be easiest for users I think. If not, then could we change the text from: "More info" to: "More info and link" ?

Here is the list:

NL: 
Datamanagementplan: https://intranet.uu.nl/system/files/2025%20-%20Dataset-list.docx

[Informatiebrief anoniem onderzoek](https://intranet.uu.nl/media/34426)
[Informatiebrief Algemeen belang](https://intranet.uu.nl/media/34425)
[Informatiebrief Toestemming]](https://intranet.uu.nl/media/34428)

[Akkoordverklaring - grondslag Algemeen belang - beeld- en geluidsopnames](https://intranet.uu.nl/system/files/gw_Akkoordverklaring%20-%20grondslag%20Algemeen%20belang%20-%20beeld-%20en%20geluidsopnames.docx)
[Toestemmingsverklaring - grondslag Algemeen belang - bijzondere persoonsgegevens](https://intranet.uu.nl/media/34431)
[Toestemmingsverklaring Toestemming](https://intranet.uu.nl/media/34434)
[Toestemmingsverklaring - grondslag Toestemming - kinderen tot 16 jaar - ouders-voogden aanwezig bij onderzoek](https://intranet.uu.nl/media/34432)
[...]

PR #934, opened 2025-05-20
Fix/incorporate link to templates

RoordinkFrank:
Issue #875 
If you want to test this manually.
Make sure "Uit welke leeftijdscategorie(ën) bestaat je deelnemersgroep?" has
above 16 and under 16 participants

The documents asked also change greatly with
¨¨Wat is de AVG grondslag voor het verzamelen van persoonsgegevens?
Dit traject is volledig anoniem.
De AVG grondslag is 'algemeen belang'.
De AVG grondslag is 'toestemming'.


attachments/utils.py
```diff
@@ -27,6 +27,8 @@ class AttachmentKind:
     max_num = None
     attached_field = "attachments"
     desiredness = desiredness.OPTIONAL
+    more_info_link = "" #optional
```
EdoStorm96, line 30: I would prefer `model_document_url`. On the home page, we have a link to the page where these files can be found with the label "Model documents" or "Voorbeelddocumenten". Although, these linked documents may not all be model documents, I think that is their general purpose and that therefore `model_document_url` is a more descriptive variable name.

attachments/utils.py
```diff
@@ -27,6 +27,8 @@ class AttachmentKind:
...
+    more_info_link_name = "" #defaults to name in template unless overridden.
```
EdoStorm96, line 31: I think that for now, this variable is not needed as AttachmentKind.name suffices. If we do ever need to distinguish between the label of an AttachmentKind and the label for the url to the model template, we can reintroduce this variable, but right now it is just one extra if-statement in my eyes.

attachments/templates/attachments/slot.html
```diff
@@ -31,6 +31,11 @@ <h5>{{ slot.kind.name }}</h5>
             <details>
                 <summary class="text-muted">{% trans "Meer info" %}</summary>
                 {{ slot.kind.description }}
+                {% if slot.kind.more_info_link_name == "" %}
```
EdoStorm96, line 34: If we get rid of more_info_link_name, this if-statement can be avoided, which I would prefer.

However, I do think it would be good to account for AttachmentKinds which do not have a more_info_link. Right now we have `OtherAttachment` for which there can be no model document because it is built for exceptions. It just has the empty string as the href, resulting in a broken link. So we do need to check if there is a more_info_link. I would suggest to just not render this part if there is no link, so doing this check like so:
```
                {% if slot.kind.more_info_link %}
                    <br>{% trans "Link: " %}<a href="{{slot.kind.more_info_link}}" target="_blank">{{ slot.kind.name }}</a>
                {% endif %}
```

EdoStorm96 (changes requested):
Ok, I have checked all the links and they all work. So hats off for that! That must have been really annoying ...

I also should have pointed you earlier to [this page](https://intranet.uu.nl/en/knowledgebase/documents-ethics-assessment-committee-humanities). This is where the secretary gets the model documents from as well, and we also link to this page on the FETC home page. This page also translates well (including links)

I do have some more issues with the design, mainly with the existence of the more_info_link_name variable. I think this is just a bit messy and unnecessary on the code side. The instances in which you've used it are not really different from the name attribute, so I think we can just use the name for the label to the links. I see that the wording is slightly different in some cases, but I think you are overestimating with how much intent the secretary wrote these labels in the issue. 

If she does get upset about this the slightly different wording you can blame me, but from the code point of view, I don't like having 2 variables which do almost the same thing and of which one gets rarely used.

Be careful when merging the .po file ... We don't want to have to redo these ... If you want we can do this together.

locale/en/LC_MESSAGES/django.po
```diff
@@ -164,19 +218,61 @@ msgstr ""
...
 
-#: attachments/kinds.py:190
+#: attachments/kinds.py:191
+msgid ""
+"Je baseert de verwerking van persoonsgegevens op de wettelijke grondslag "
+"algemeen belang, maar je maakt beeld- en/of geluidsopnames van kinderen in "
+"afwezigheid van diens ouders of voogden. Daar moeten de ouders of voogden "
+"van de deelnemers op ethische gronden mee instemmen. Ook voor het verdere "
+"gebruik van die opnames kun je de afspraken schriftelijk vastleggen."
+
```
EdoStorm96, line 250: I don't like this newline ...

PR #935, opened 2025-05-20
Fix/date end

### LINKED ISSUE #933: Some In Revision Review do not have a date_end on In Revision Reviews list
The secretary emailed us about this, but I am making an issue for it now. I think I've found the culprit. We moved around how date_end in this commit: 2231ce66d64c667e0bc0b4f0005fd16d528abc08. So therefore, Reviews only receive a date_end when they are discontinued (`ReviewDiscontinueView` -> `discontinue_review()`), whereas in the past, they also received one when all decisions had been made, through update_go ... Do we want to make this feature return?

EdoStorm96:
fixes #933.

So there was an issue where most often reviews did not receive a date end. This was due to a commit, which I've linked in the issue. The secretary noticed the lack of end dates on the InRevision list.

I considered calling update_go through a data migration, but sadly migrations do not have access to model methods. So when deploying this, we should go to the shell and run something like:
```
>>> no_date_review = Review.objects.filter(date_end__isnull=True)
>>> for review in no_date_reviews:
...     review.update_go()
...     review.save()
... 
```
This does seem to work fine in my limited testing


RoordinkFrank (approved):
Do not forget to communicate to users that they may receive old emails when
we are executing review.update_go() during deployment phase.
I believe that should be the secretary only.

EdoStorm96:
Alright, so after the comments by Frank and looking more thoroughly at `update_go()`, I figured it is not ideal to use this function for setting a date_end, so I went back to the data migration idea. I manually reworked the parts of update_go() that we need to set a date_end into a data_migration, so now this won't have any unwanted side effects.

RoordinkFrank (approved):

miggol (approved):
This looks good! I have a sole requested change but otherwise approve.

This is an important migration to keep track of, so I would like to ask that we wait with merging this into develop until we move the *current*  develop along to acc.

Then we can keep the registration (#904) and this PR together to only have one deployment we really have to keep an eye on.

Issue #938, opened 2025-05-23
Refactor Decision-for-secretary-exists cache


RoordinkFrank:
Decision-for-secretary-exists cache inside reviews/api/views.py now has partly duplicate code.
Refactor to remove this.
Overlapping refactoring with Issue @631 so wait until that Issue is closed.

EdoStorm96:
@631 is not an issue :) #631

RoordinkFrank:
Duplicate code is still there but logic is harder than expected. For now I have given up. Not worth the effort of refactoring right now.

PR #939, opened 2025-05-23
Fix/refactor opensupervisordecision queryset

RoordinkFrank:
Duplicate code found and refactored
added review__is_committee_review=False to queryset.
Further duplications found. Seperate Issue made in #938


reviews/api/views.py
```diff
@@ -296,20 +306,11 @@ def get_queryset(self):
         """Returns all proposals that still need to be reviewed by the supervisor"""
         objects = Decision.objects.filter(
             go="",
-            review__stage=Review.Stages.SUPERVISOR,
-            review__proposal__status=Proposal.Statuses.SUBMITTED_TO_SUPERVISOR,
+            review__is_committee_review=False,
```
miggol, line 309: This API endpoint should return *open* supervisor reviews, so we still need to check the stage and/or proposal status. Or figure out if the review is still open some other way.

reviews/api/views.py
```diff
@@ -16,6 +16,32 @@
 from .serializers import DecisionSerializer, ReviewSerializer
 
 
+def return_latest_decisions(objects):
+    # return latest decisions from each proposal
```
miggol, line 20: I would say: return *only* the latest decision for each proposal in objects.

This could be a docstring rather than a comment.

miggol (changes requested):
The Open supervisor reviews filter isn't right, but otherwise nice deduplication!

miggol (approved):
Approved!

PR #940, opened 2025-05-26
fix/children_criterion_missing

RoordinkFrank:
Can be manually checked if you start a review as a secretary. Long route reasons should show up in the middle of the screen.


reviews/utils/review_utils.py
```diff
@@ -510,6 +510,12 @@ def auto_review(proposal: Proposal):
     reasons = []
 
     for study in proposal.study_set.all():
+        aanvraag_minderjarigen_added = False
+        for age_group in study.age_groups.all():
```
miggol, line 514: There are a couple of things I would do differently here.

First, the for loop. When you start iterating over a Django QuerySet, Django converts that QuerySet into actual SQL and executes it in the database. Usually you want to do as much as possible in your QuerySet, because Django is very good at optimizing QuerySets.

In this vein, you can optimize this loop by just checking for minors with `study.age_groups.filter(is_adult=False).exists()`, which will return `True` if at least one minor age group is selected. [Link](https://docs.djangoproject.com/en/5.2/ref/models/querysets/#django.db.models.query.QuerySet.exists)

Second, if you were to use a for loop here (which I wouldn't recommend here, but anyway), you might want to have a look at the `break` keyword. [Link](https://docs.python.org/3/tutorial/controlflow.html#break-and-continue-statements)

With it you could remove the temporary variable:

```
    for study in proposal.study_set.all():
        for age_group in study.age_groups.all():
            if not age_group.is_adult:
                reasons.append(_("De aanvraag bevat minderjarigen"))
                break  # The for loop ends here
```

Because why would you check the other age groups, if you've found what you were looking for already?

Finally, I wanted to mention that even for temporary variables, we prefer to keep things in English.

reviews/tests.py
```diff
@@ -331,15 +331,15 @@ def test_auto_review_age_groups(self):
...
+        self.assertEqual(len(reasons), 1)  # 1 from toddlers are minors
 
         s1_t2.duration = 30
         s1_t2.save()
         self.study.save()
         self.assertEqual(s1.net_duration(), 50)
 
         reasons = auto_review(self.proposal)
-        self.assertEqual(len(reasons), 1)
+        self.assertEqual(len(reasons), 2)  # totale duur en minors
```
miggol, line 342: Try to keep comments in English, as well :)

locale/en/LC_MESSAGES/django.po
```diff
@@ -5775,3 +5775,6 @@ msgstr "Task edited"
 #: tasks/views/task_views.py:84
 msgid "Taak verwijderd"
 msgstr "Task deleted"
+
+msgid "De aanvraag bevat minderjarigen"
+msgstr "The application has contains minors"
```
miggol, line 5780: The application either has or contains minors, not both

miggol (changes requested):
I have requested some changes

reviews/utils/review_utils.py
RoordinkFrank: I tend to avoid break and continue statements as they can cause possible bugs. That said the query solution is way more elegant.

RoordinkFrank: On another note. This logic is inside of a for loop and uses the reasons variable. Does it still optimize the query at that point?

EdoStorm96 (changes requested):
Alright nice work! good you updated the tests as well. I just have a few minor things I'd like to see changed.

EdoStorm96 (approved):
Yes, looks good!

reviews/tests.py
```diff
@@ -266,6 +266,13 @@ def test_decision_commission(self):
 
 
 class AutoReviewTests(BaseReviewTestCase):
+
+    def setUp(self):
+        super().setUp()
+        self.toddlers = AgeGroup.objects.filter(pk=2)
+        self.adolescents = AgeGroup.objects.filter(pk=4)
+        self.adults = AgeGroup.objects.filter(pk=5)
```
miggol, line 274: I like this

reviews/tests.py
```diff
@@ -362,21 +382,21 @@ def test_auto_review_observation(self):
...
         reasons = auto_review(self.proposal)
-        self.assertEqual(len(reasons), 0)
+        self.assertEqual(len(reasons), 1)  # adolescents are minors
 
         s1 = Session.objects.create(study=self.study, order=1)
         s1_t1 = Task.objects.create(session=s1, order=1)
-        s1_t1.registrations.set(
-            Registration.objects.filter(pk=6)
-        )  # psychofysiological measurements
+        psychofysiological_measurement = Registration.objects.filter(pk=6)
```
miggol, line 395: Um, I would rather we match by name of the registration here. At least if the name changes, the test will be guaranteed to fail. 

If the pk changes, that could go unnoticed for a long time.

miggol (approved):
Very good! I have a minor comment but I do approve either way.

PR #942, opened 2025-05-27
fix/choosing_an_institute

RoordinkFrank:
Help text added.
Alternative is adding a widget to extend django core so that the institutes can be seperated with group by. this is the easier solution


locale/en/LC_MESSAGES/django.po
```diff
@@ -5891,4 +5891,11 @@ msgstr "Task edited"
...
-msgstr "Task deleted"
\ No newline at end of file
+msgstr "Task deleted"
+
+msgid ""
+"Algemene Kamer: OFR, OGK, ICON \n"
+"Linguïstieke Kamer: ILS"
+msgstr ""
+"General Chamber: OFR, OGK, ICON \n"
+"Linguistics Chamber: Ils"
```
EdoStorm96, line 5901: It looks like you manually added this. I think it would be better for consistency to add new translations using makemessages, to ensure we don't miss anything. Also, I think 'ILS' should be spelled in allcapps in the englisch translation.

EdoStorm96 (changes requested):
Just a request regarding the translation. Otherwise I agree with @miggol.

PR #943, opened 2025-05-27
Fix/broken pdf filenames

### LINKED ISSUE #941: PDF not systematically named anymore.
This recently occurred for the same application (I think version 01), but I cannot find an issue on Github. Since it now happens again for Algemene Kamer 25-111-02, I am submitting this issue. When downloading the pdf of this particular application, the name is not systematic but the file gets a strange code each time you download it.

EdoStorm96:
fixes #941.

The issue should provide plenty of context. Have a look at the [list of Unicode characters](https://en.wikipedia.org/wiki/List_of_Unicode_characters) and their decimals to get an idea of which chars work and which will not.

This will ensure the bug does not occur, but it is still kindoff strange and unexplained why having an unwanted char results in using the filename from OverwriteStorage ...

### DIFF
--- a/proposals/utils/proposal_utils.py
+++ b/proposals/utils/proposal_utils.py
@@ -365,8 +365,16 @@ class FilenameFactory:
                 trajectory = "Extra" + str(extra_index)
 
         chamber = proposal.reviewing_committee.name
-        lastname = proposal.created_by.last_name
         refnum = proposal.reference_number
+
+        # We need to make sure that we only use characters that
+        # that are in ISO 8859-1 AKA the first 256 chars of unicode. Otherwise, 
+        # a bug occurs, where the OverwriteStorage filename gets served to the
+        # user. If there are special characters in a user's name, we replace
+        # them with '_'.
+
+        lastname = proposal.created_by.last_name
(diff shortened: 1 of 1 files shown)


proposals/utils/proposal_utils.py
```diff
@@ -407,3 +412,12 @@ def get_available_name(self, name, **kwargs):
         if self.exists(name):
             os.remove(os.path.join(settings.MEDIA_ROOT, name))
         return super(OverwriteStorage, self).get_available_name(name, **kwargs)
+
+
```
RoordinkFrank, line 416: New def looks great.
The Issue is Fix/broken pdf filenames and yet this sits in utils/proposal_utils.py
Does this not belong in a pdf util file? If you ever need this function for pdf generation that is not proposal related you would not expect the function to be here, let alone know of it's existence.

RoordinkFrank (changes requested):

miggol (approved):
Looks good! 

Sorry this took a while to review, I was trying to find the root cause. I will create a new issue for that because I do think it's suboptimal that only 256 characters are allowed in our filenames.

PR #945, opened 2025-06-02
Fix/menus user attribute error

### LINKED ISSUE #944: WSGIRequest has no attribute "user"
Related Sentry issue: [ETHICS-1](https://sentry.it.hum.uu.nl/organizations/centre-for-digital-humanities/issues/28/events/ca616e1785784991936c25bd5510aba8/)

Whenever a client tries to visit the portal through a disallowed host, an error occurs. Django tries to serve our client an error page, but crashes because `request.user` is unset. `request.user` is required for the rendering of our menus.

How to reproduce:

1. Set `DEBUG = False`
2. Set `ALLOWED_HOSTS`
3. visit your local server through a URL that is not specified in `ALLOWED_HOSTS`
4. Note the basic error page and traceback in console

This is not a high priority issue were it not for the hundreds of sentry reports we get for it, which are annoying and make it harder to figure out what actually does require attention.

miggol:
Fixes #944 

I saw four possible ways of fixing this:

1. Change error templates to no longer render menus, which would require modifying `tool-base.html` in DSC.
2. Modify all existing auth helper functions to access `request.user` safely.
3. Subclass MenuItems to do all the processing in a safe manner.
4. Add a couple wrapper functions that access `request.user` safely

I ended up going with number 4 as the least intrusive. But if we were engineering this from scratch I may have preferred a form of solution 1 or 3.

### DIFF
--- a/main/utils.py
+++ b/main/utils.py
@@ -4,6 +4,7 @@ from django.contrib.auth.models import Group
 from django.contrib.staticfiles.storage import staticfiles_storage
 from django.db.models import Q
 from django.db.models.fields.files import FieldFile
+from django.contrib.auth.models import AnonymousUser
 from django.utils.translation import gettext_lazy as _
 from django.template import loader, Template, Context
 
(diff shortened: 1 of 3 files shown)


miggol:
Descriptive argument names are a good idea! But `lambda x: is_authenticated(x)` really rubs me the wrong way

Issue #950, opened 2025-06-12
Discontinued reviews do not get closed ...


EdoStorm96:
"Niet verder in behandeling genomen" AKA `Review.continuations.DISCONTINUED` does not get checked for in `ReviewCloseForm.form_valid()` ...

So these reviews are never closed, and do not receive a `date_reviewed`

I think it should just be included in the first if-statement of the form_valid.

EdoStorm96:
Possibly related to #677

Issue #677, opened 2024-06-17
Niet verder in behandeling genomen - dates are swopped


djhcapel:
In cases where the assessment procedure is stopped without a formal decision (e.g. withdrawal of application), something goes wrong with the dates:

The date "Laatste beoordeling ontvangen op" seems to be swopped with the date "Afhandeling: Niet verder in behandeling genomen". This also leads to errors in the statistics.

tymees:
For something to be swapped, there actually needs two things that swap places. That is not the case here, there is only one date with a misleading label. A correcter name would be 'Date review ended'; This normally corresponds to 'date last decision', but in this case the review wasn't ended because the last decision was received.

(Internally, the field is actually called `date_ended`).

That being said, that's just a explanation for this behavior. It also points us to a potential fix: a new field `date_last_decision`. This field should be updated every time a decision is made (hint: `Decision.update_go`); That way, the field will correspond to `date_end` when the review is closed normally, but will have a sensical name in the `discontinue` scenario.

djhcapel:
Not sure if I understand this correctly. There can be two dates in case of 'niet verder in behandeling genomen', but this does not necessarily have to be the case. In principle, an application can be 'niet verder in behandeling genomen' at any time, also right after submission. In some cases, a first decision has already been sent, and that date should be the date the applicant was informed about that decision. The date 'niet verder in behandeling genomen' can be an entirely different date.

djhcapel:
Another example (and this information is just wrong, so it should be corrected at some point): see GC 24-086-01:
 Laatste beoordeling ontvangen op
28 februari 2025, 14:40.
Afhandeling:
Niet verder in behandeling genomen,
6 juni 2024, 12:32. 

This should read:
 Laatste beoordeling ontvangen op
**6 juni 2024, 12:32**. 
Afhandeling:
Niet verder in behandeling genomen,
**28 februari 2025, 14:40.**

djhcapel:
Apart from GC 24-086-01, other examples of where the dates go wrong are:
GC 24-087-01
GC 24-154-01
GC 24-156-01

According to the correspondence, these applications were under revision, but I do not see that anymore in the portal. What happened to these applications is that during the revision stage, it appeared that applicants wanted to withdraw their application for various reasons.

So the most recent date should be the date at which the assessment procedure was stopped (Niet verder in behandeling genomen), and the other date belongs to the step 'Laatste beoordeling ontvangen op'.

Two things:
[...]

PR #952, opened 2025-06-12
Dates are swapped 

RoordinkFrank:
#950 fixed, not #677


RoordinkFrank:
I have gotten an answer in teams from Michael, I go into to much detail in trying to rework the end_date. Fair enough.

"What changes are we making to the discontinue_review() function and why are we changing it at all? I don't think discontinued proposals should get a date_reviewed, because they were never reviewed. Also why are we no longer setting date_end? These are major changes to how this continuation functions and I see no justification for them. Please explain!"

I have done to much of an information dump in the above post to make it more clear
image
 "Last decision received on" is no longer updated and "Not to be assessed further" is now updated.
There is also this warning
image

miggol:
So Frank and I had a nice long chat today covering the comments above. After which two questions remained:

1. Through which view can proposals be discontinued? We currently have two views making for three options: ReviewCloseView, ReviewDiscontinueView, and keeping both.
2. Can we implement @djhcapel's feature request of having a discontinue date displayed somewhere using an existing date field or should we put that in a new field?

My answer to (1) is clear: we should just keep the discontinue view and remove the discontinuation from the options in ReviewCloseView. ReviewCloseView is only available for reviews that have ended, whereas any review or proposal should be able to be discontinued. This means that removing ReviewDiscontinueView is not an option.

If we allow both views to co-exist, there is a risk of only changing one view but not the other resulting in bugs. Also ReviewCloseView does not show the same description and warnings as ReviewDiscontinueView, which I think are important.

And then for (2): although I was initially apprehensive to use `Proposal.mark_reviewed` for this purpose, I see now that this is a great option. This attribute already serves a similar purpose and would fix the lack of discontinue date with minimal effort.

Unlike what I initially thought, `date_reviewed` is not used elsewhere to confirm a proposal was reviewed or anything like that, and it never shows up in a PDF or similar. So it's a good option to put the discontinue date in.

RoordinkFrank:
> So Frank and I had a nice long chat today covering the comments above. After which two questions remained:
> 
>     1. Through which view can proposals be discontinued? We currently have two views making for three options: ReviewCloseView, ReviewDiscontinueView, and keeping both.
> 
>     2. Can we implement @djhcapel's feature request of having a discontinue date displayed somewhere using an existing date field or should we put that in a new field?
> 
> 
> My answer to (1) is clear: we should just keep the discontinue view and remove the discontinuation from the options in ReviewCloseView. ReviewCloseView is only available for reviews that have ended, whereas any review or proposal should be able to be discontinued. This means that removing ReviewDiscontinueView is not an option.
> 
> If we allow both views to co-exist, there is a risk of only changing one view but not the other resulting in bugs. Also ReviewCloseView does not show the same description and warnings as ReviewDiscontinueView, which I think are important.
> 
> And then for (2): although I was initially apprehensive to use `Proposal.mark_reviewed` for this purpose, I see now that this is a great option. This attribute already serves a similar purpose and would fix the lack of discontinue date with minimal effort.
> 
> Unlike what I initially thought, `date_reviewed` is not used elsewhere to confirm a proposal was reviewed or anything like that, and it never shows up in a PDF or similar. So it's a good option to put the discontinue date in.

Then I will remove the ReviewCloseView option to discontinue the view.

miggol:
I think we should consider avoiding `mark_reviewed` when a proposal gets discontinued. The processes are different and I can see a risk of future changes to `mark_reviewed` unintentionally affecting `discontinue_review`

reviews/tests.py
miggol, line 1: These imports are now no longer needed

reviews/forms.py
```diff
@@ -116,10 +113,16 @@ def __init__(self, *args, **kwargs):
...
+            if x[0] != Review.Continuations.DISCONTINUED
+        ]
+
         if not allow_long_route_continuation:
             self.fields["continuation"].choices = [
                 x
-                for x in Review.Continuations.choices
+                for x in self.fields["continuation"].choices
                 if x[0] != Review.Continuations.LONG_ROUTE
             ]
```
miggol, line 127: 1. I think it would be good to have a comment along the lines of `# Reviews should only be discontinued with ReviewDiscontinueView`
2. (optional) This may be a question of taste but two multiline list comprehensions just to remove two options is a bit awkward. An exclude list or a solution with `pop()` would look nicer.

reviews/views.py
```diff
@@ -422,6 +422,7 @@ def get_success_url(self):
     def form_valid(self, form):
         "Sets the discontinued continuation on the review"
         review = form.instance
+        # mark_reviewed should already have been done at this point.
```
miggol, line 425: `porposal.mark_reviewed()` is not called when a proposal gets discontinued.

Mark reviewed sets status_review and generates a pdf, which are things I don't think we want to happen to discontinued reviews.

Best to remove this comment to reduce confusion I think. Or am I the one who is confused here?

miggol (approved):
Okay, sorry for the late review! I think this is good now, though I do have some optional comments that would make this PR even nicer.

PR #967, opened 2025-07-10
added expected end date to research

RoordinkFrank:
fixed duplication in django.po file
added expected end date to research
field is currently not required, I have asked the DJ how that sounds.


EdoStorm96 (approved):
Nicely done! You thought of everything and it all looks good to me.The wording of the question looks a bit funky to me, but it came straight from the secretary, so we'll accept it ;p

PR #970, opened 2025-07-11
Fix/export script

RoordinkFrank:
Test this with: python manage.py export_csv 2025

One could argue registration would be a better name then registrations as the rest of the model variables is named that way. Regardless it is fixed.

Warning: subbranched from another Issue where the double translations error is fixed. That is why you see a bit more code. Should be fixed once that Issue is merged.

"f"{registration.description}: {study.registration_details}"
                    f"{registration.description}: {study.registrations_details}"

This is the only change on this branch.

### DIFF
--- a/attachments/templates/attachments/slot.html
+++ b/attachments/templates/attachments/slot.html
@@ -29,7 +29,7 @@
         </div>
         <div class="uu-form-help">
             <details>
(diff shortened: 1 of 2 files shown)


EdoStorm96 (approved):
Looks good!

PR #982, opened 2025-07-28
Feat/uu list 4

### LINKED ISSUE #557: Implement UU-List
Replaces issues #225 (That's a really low number...) and #500 

The new UU-List can be used in ethics, as the Python dependencies needed do not include `cdh.core`. 

It will, however, need a CSS hack to get the styling working as UU-List is a UU Bootstrap component. 
Luckily, we can just do a partial compile of UUB to only include the UU-List component. It will probably need a few more tweaks, and it will never look as nice as when used as part of UUB, but it will be functional. (And a way better UX for the secretary).

I would also like propose removing all the existing applicant list views with just one UU-List implementation. The variants we have are just very simple QS changes on a single field (status). That's just a filter. 

Something similar can, probably, be done with the committee lists. However, the queries there are more complex. So it might not be feasible.

RoordinkFrank:


main/serializers.py
```diff
@@ -8,3 +8,12 @@ class Meta:
         fields = ["pk", "fullname", "first_name", "last_name", "email"]
 
     fullname = serializers.CharField(source="get_full_name")
+
+
+# redundant text with above, is it possible to override thhe fields?
+class UserMinimalSerializer(serializers.ModelSerializer):
```
RoordinkFrank, line 14: I do prefer to be able to write this comment before a draft pull request (or commit). So if anyone knows how to do that.

Back to the original comment. The minimal serializer gives three value, fullname for display, first and last name for the filter. I suspect the filter getś it's data from somehwere else and is unneccesary.

proposals/api/serializers.py
```diff
@@ -76,31 +88,126 @@ class Meta:
...
+        users = ""
+        for user in proposal.applicants.all():
+            if users != "":
+                users += ", "
+            users += user.first_name + " " + user.last_name
+        return users
+
+    @staticmethod
+    def get_state(proposal: Proposal):
+        # state = status with decisions added so this is still lacking
```
RoordinkFrank, line 210: status displayed still needs more, decision of the ethics commision should be displayed here as well.

proposals/views/proposal_views.py
```diff
@@ -86,18 +96,76 @@ def get_context_data(self, **kwargs):
...
+            field="state",  # may need to be shown in different forma
+            label="state",
+        ),
+        DDVDate(
+            # incoming format: 2025-04-24T11:59:45.583305+02:00
+            field="date_submitted",  # may need to be shown in different format, time is now gone
+            # compared to original data shown.
+            label="date submitted",
+        ),
+        DDVDate(  # DDVDate is an option, however the default for their empty dat is 1970 instead of empty.
```
RoordinkFrank, line 135: DDVDate has 1970 as a default option when there is an empty field. That is different with current behaviour.

proposals/models.py
```diff
@@ -776,11 +786,18 @@ def supervisor_decision(self):
 
             return self.supervisor_decision()
 
+    @property
     def latest_review(self):
         from reviews.models import Review
 
         return Review.objects.filter(proposal=self).last()
```
RoordinkFrank, line 794: the serializer seems to want a direct field of the model. And a field of a field is not accepted. Still this is not a great solution as it clutters the proposal model even more than it already is.
I could make a separate model matching the serializer where there are methods like these. That would remove this from proposal. I think it may end up slightly better but not ideal either.

proposals/api/views.py
```diff
@@ -2,16 +2,91 @@
...
+    def get_queryset(self):
+        return Proposal.objects.filter(
+            Q(applicants=self.request.user),
+            Q(in_course=False) & Q(is_exploration=False),
+        )
+
+
+class MyPracticeApiView(ProposalApiView):
+    """Gets all practise applications including supervisor."""
+
```
RoordinkFrank, line 64: # practise proposals for supervisor currently do appear in MySupervised to match the action buttons.
    # Uses probably do expect it in this view but I am unsure how to implement that.

proposals/api/serializers.py
```diff
@@ -99,8 +112,174 @@ class Meta:
...
+        return ""
+
+    @staticmethod
+    def get_date_submitted(proposal: Proposal):
+        if proposal.date_submitted is not None:
+            return proposal.date_submitted.date()
+        return ""
+
+    # logic works but unsure if it should be in the serializer
+    @staticmethod
```
RoordinkFrank, line 231: logic works but unsure if it should be in the serializer

proposals/api/serializers.py
```diff
@@ -99,8 +112,174 @@ class Meta:
...
+    action_delete = DDVLinkField(
+        text=_("Verwijderen"),
+        link="proposals:delete",
+        link_attr="pk",
+        check=lambda proposal: ProposalActions.action_allowed_delete(proposal),
+    )
+
+    action_make_revision = DDVLinkField(
+        text=_("Maak revisie"),
+        link="proposals:copy",  # there is also a copy revison and copy amendment,
```
RoordinkFrank, line 173: there is also a copy revison and copy amendment,
         not sure when those two are supposed to be shown

proposals/utils/proposal_actions.py
```diff
@@ -0,0 +1,91 @@
...
+     {% if submitted %} 
+        <img src="{{ img_pdf }}" title="Inzien">
+        <img src="{{ img_revise }}" title="Maak revisie">
+      {% if supervised %}
+          <img src="{{ img_decide }}" title="Beslissen">
+       {% if is_secretary %}
+             <img src="{{ img_hide }}" title="Verbergen">
+               <img src="{{ img_add }}" title="Toevoegen">
+"""
+
```
RoordinkFrank, line 27: import datetime

# actiontree, implementation in proosal_actions.py
"""
    {% if modifiable %}
        
        
        
     {% if submitted %} 
        
        
      {% if supervised %}
          
       {% if is_secretary %}
             
               
"""

proposals/utils/proposal_actions.py
```diff
@@ -0,0 +1,91 @@
...
+
+# Does not check user permissions as of now
```
RoordinkFrank, line 29: Does not check user permissions as of now

proposals/views/proposal_views.py
```diff
@@ -86,21 +96,115 @@ def get_context_data(self, **kwargs):
...
+    # I do want to do something like below to avoid duplicate DDVcolumns but I can not get it to work
+    # the subclass does not seem to know the superclass variable if they are class variables.
+    # I have a simular problem in the serializer, if there is someone who might know a solution please say so.
+    # columns.append(
+    #    DDVActions(
+    #        field="my_supervised_actions",
+    #        label=_("Acties"),
+    #    ),
+    # )
+    columns = [
```
RoordinkFrank, line 167: # I do want to do something like below to avoid duplicate DDVcolumns but I can not get it to work
    # the subclass does not seem to know the superclass variable if they are class variables.
    # I have a simular problem in the serializer, if there is someone who might know a solution please say so.
    # columns.append(
    #    DDVActions(
    #        field="my_supervised_actions",
    #        label=_("Acties"),
    #    ),
    # )

proposals/models.py
```diff
@@ -622,6 +621,14 @@ class PracticeReasons(models.IntegerChoices):
         related_name="applicants",
     )
 
+    def get_applicants_names(self):
+        users = ""
+        for user in self.applicants.all():
+            if users != "":
```
RoordinkFrank, line 627: .all() has a yellow error. Unsolved attribute. It works fine but I have no idea how to remove the error.

proposals/utils/proposal_actions.py
```diff
@@ -0,0 +1,66 @@
...
+
+    @staticmethod
+    def action_allowed_edit(proposal: Proposal):
+        # if_modifiable return true
+        return proposal.status == Proposal.Statuses.DRAFT
+
+    @staticmethod
+    def action_allowed_show_difference(proposal: Proposal):
+        # if_modifiable return true
+        if proposal.is_revision:  # amendment_or_revision is also an option?
```
RoordinkFrank, line 23: amendment_or_revision is also an option?

proposals/utils/proposal_actions.py
```diff
@@ -0,0 +1,66 @@
...
+    @staticmethod
+    def action_allowed_make_supervise_decision(proposal: Proposal):
+        # if_supervised return true
+        check = lambda o: o.status == Proposal.Statuses.SUBMITTED_TO_SUPERVISOR
+        return check(proposal)
+
+    # hid and add to archive may not belong here, but I will not remove them without confirmation that they do not belong
+    # in a proposal view.
+    @staticmethod
+    def action_allowed_hide_from_archive(proposal: Proposal):
```
RoordinkFrank, line 57: Hide and show will be implemented during the archive DDL view.

proposals/views/proposal_views.py
```diff
@@ -86,21 +96,115 @@ def get_context_data(self, **kwargs):
...
+        DDVString(
+            field="usernames",
+            label=_("Indieners"),
+        ),
+        DDVString(
+            field="state_or_decision",
+            label=_("Status"),
+        ),
+        DDVString(
+            field="date_modified",  # in the old version this is either laatst bijgewerkt or besloten op.
```
RoordinkFrank, line 128: in the old version this is either laatst bijgewerkt or besloten op.
That last interface part is now lost.

proposals/utils/proposal_actions.py
```diff
@@ -0,0 +1,50 @@
...
+    @staticmethod
+    def action_allowed_make_supervise_decision(proposal: Proposal):
+        check = lambda o: o.status == Proposal.Statuses.SUBMITTED_TO_SUPERVISOR
+        return check(proposal)
+
+    @staticmethod
+    def action_allowed_hide_from_archive(proposal: Proposal):
+        check = lambda o: o.status == Proposal.Statuses.SUBMITTED
+        return check(proposal)
+
```
RoordinkFrank, line 46: lamda logic is left intact so it can be ported back to the serializer if ProposalActions gets removed.

proposals/utils/proposal_actions.py
```diff
@@ -0,0 +1,66 @@
...
+    @staticmethod
+    def action_allowed_make_revision(proposal: Proposal):
+        # if_submitted return true
+        return proposal.is_revisable
+
+    @staticmethod
+    def action_allowed_view_pdf(proposal: Proposal):
+        # if_submitted return true
+        # does seem to be always available, why is there a submitted requirement?
+        return proposal.status == Proposal.Statuses.SUBMITTED
```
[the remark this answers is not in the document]
EdoStorm96: The pdf might be always availabe ... But we only link to it after it gets submitted. I think that makes sense so let's keep it that way.

proposals/utils/proposal_actions.py
```diff
@@ -0,0 +1,50 @@
...
+from django.conf import settings
+from django.utils.safestring import mark_safe
+
+from proposals.models import Proposal
+from reviews.models import Review, Decision
+
+import datetime
+
+
+class ProposalActions:
```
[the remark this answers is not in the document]
EdoStorm96: I do see how this is a problem ... But I think we can overcome this by carefully constructing querysets and splitting out in a few lists, like you did with my_proposals and my_supervised. Because we can access the user in the api views, we can ensure that only relevant proposals appear in a certain list. Then we can only make actions that should only be available to a specific user, only available in the specific list where we know only that specific user can come. I think taken together, this will create a solid system.

EdoStorm96 (commented):
Ok, so really good work! Massive props for figuring all of this out by yourself. That must have been challenging. I think it is looking really snazzy! I think really the main stuff is done, but we are just missing some functionality and polish. But overall this is really on the right track!

I tried to be as thorough as possible with my review, but I might have missed some stuff. I just found a lot of little things which is to be expected with such a large PR. I hope you find my comments useful. But in general you can keep going in this direction I think!

I have one general comment: ensure al references to removed list views are also removed eg. proposal_submitted.html.

It will probably take quite a while for this to be fully finished, as it is a really big feature, but do not be discouraged by that. We will need to iterate over this a few times, before it gets perfect (also on the review side) and hopefully @miggol can take a look as well. But in general this is going in the right direction!

proposals/models.py
```diff
@@ -741,6 +748,7 @@ def amendment_or_revision(self):
         if self.is_revision and self.parent:
             return _("Amendement") if self.parent.status_review else _("Revisie")
 
+    @property
```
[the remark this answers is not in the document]
RoordinkFrank: The added @property? Serializer does not recognize methods and gives back a yellow error without it.

proposals/api/views.py
```diff
@@ -2,16 +2,88 @@
...
+    status_choices = Proposal.Statuses.choices
+    for choice in status_choices:
+        if choice[0] == 60:
+            status_choices.remove(choice)
+    status = filters.MultipleChoiceFilter(
+        label="Status",
+        field_name="status",
+        choices=status_choices,
+    )
```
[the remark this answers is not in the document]
RoordinkFrank: Definitely looks cleaner already.

proposals/api/serializers.py
RoordinkFrank: Moved logic to proposal

proposals/api/serializers.py
```diff
@@ -99,8 +112,174 @@ class Meta:
...
+            "state_or_decision",
+            "usernames",
+            "info",
+            "my_supervised_actions",
+        ]
+
+    info = serializers.SerializerMethodField()
+
+    # I am having trouble finding out what info best to display to the supervisor
+    @staticmethod
```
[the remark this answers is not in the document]
RoordinkFrank: done

reviews/utils/review_actions.py
```diff
@@ -9,6 +9,7 @@
 import datetime
 
 
+# revviewactions are here
```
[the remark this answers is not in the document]
RoordinkFrank: done

proposals/utils/proposal_actions.py
RoordinkFrank: ah so we have a templates/proposals/proposal_list.html and a templates/proposals/vue_templates/proposal_list.html

proposals/api/serializers.py
```diff
@@ -99,8 +112,157 @@ class Meta:
...
+    )
+
+    date_modified = serializers.SerializerMethodField()
+    date_submitted = serializers.SerializerMethodField()
+    detailed_state = serializers.SerializerMethodField()
+    usernames = serializers.SerializerMethodField()
+
+    # DDVDate returns 1970 by default so we have to use a DDVString and format the date ourselves
+    # incoming format: 2025-04-24T11:59:45.583305+02:00
+    @staticmethod
```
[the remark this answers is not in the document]
RoordinkFrank: <div class="ufl-undertitle-line">
                <span v-if="proposal.latest_review && proposal.latest_review.continuation == context.review.REVISION">
                    {% trans "Besloten op" %}:
                    {% verbatim %}
                    {{ proposal.date_reviewed | date("DD-MM-YYYY, HH:mm") }}
                    {% endverbatim %}
                </span>
                <span
                    v-else-if="proposal.date_confirmed"
                >
                    {% trans "Besloten op" %}:
                    {% verbatim %}
                    {{ proposal.date_confirmed | date("DD-MM-YYYY, HH:mm") }}
                    {% endverbatim %}
                </span>
                <span v-else>
                    {% trans "Laatst bijgewerkt" %}:
                    {% verbatim %}
                    {{ proposal.date_modified | date("DD-MM-YYYY, HH:mm") }}
                    {% endverbatim %}
                </span>
            </div>

Basicly fusing these three, we only have place for one label though.

proposals/api/serializers.py
```diff
@@ -99,8 +112,157 @@ class Meta:
...
+    def get_date_modified(proposal: Proposal):
+        if proposal.date_modified is not None:
+            return proposal.date_modified.date()
+        return ""
+
```
RoordinkFrank, line 215: date_submitted is not date submitted to supervisor but date submitted to commission?
the wording of date_submitted seems to be wrong.

miggol (commented):
This is really cool! Congrats on getting this to work. There's still a way to go. I have some comments to start:

1. I don't see much utility in filtering by status. I think the left-hand checkbox filters should be "supervised", "draft", "processed", etc. This can supplant the need for all the separate views we have today.
2. There are too many columns displaying too much text. Text density is high while information density is low. This can be fixed by having a detail view, or having a row "expand" when clicked like with the current lsits. Or perhaps use icons or colours to represent certain information.
3. I'm a bit conflicted about the actions implementation. I like that they work and are understandable. But they are hard to reuse and maintain in this way. Perhaps we should have a talk about this in person, I'm keen to hear your opinion on doing it either way.

UU-list-4 will not be ready for merging into develop anytime soon, this is more of a long term feature. I would recommend going further as follows:

* This branch becomes the new base for uu-list development
* Further changes come from PRs into this branch, which can come from anyone and hopefully be smaller changesets. I see you've already started on this!
* Perhaps we can put this branch on the test server. Edo knows how to do this. This does require that it is kept to a certain degree of usability.
* I think that when we get close to merging this, we will want a fresh PR without dusty old comments. So IMO this PR can be closed while we keep working on the branch. But I don't mind keeping it either.

proposals/api/serializers.py
```diff
@@ -99,8 +112,160 @@ class Meta:
...
+    # A small DDV explanation:
+    # link_attr= variable in the model. For example proposal.pk is used in action_view_pdf.
+    # the lambda function which we give also uses the model object parameter
+    # which is otherwise not available as far as I can see.
+    action_view_pdf = DDVLinkField(
+        text=_("PDF Inzien"),
+        link="proposals:pdf",
+        link_attr="pk",
+        new_tab=True,
+        check=lambda proposal: ProposalActions.action_allowed_view_pdf(proposal),
```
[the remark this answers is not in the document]
RoordinkFrank: That makes way more sense. It's just that I don´t know how. My tries didn´t work so I did leave it as it is.

RoordinkFrank:
Will be on hold until summer at least, likely permanently.

PR #1001, opened 2025-09-24
add pre assessment revisions

### LINKED ISSUE #999: Pre-assessment proposal should also be able to get revisions
Recently a pre-assessment proposal was sent for revision, but looking at the queryset which fetches revisable proposals, this is currently not possible as we filter out pre-assessment proposals. This should not happen.

So I would say `BaseProposalCopyForm._get_parent_queryset()` and `RevisionProposalCopyForm._get_parent_queryset()` should no longer filter for `is_pre_assessment=False`.

RoordinkFrank:
After a pre assesment revision has been submitted the secretary view looks like this:
image
change comparison looks like this:
image
It does look in order as far as I can see.

backup comments before git reset
image
image


proposals/forms.py
```diff
@@ -500,12 +500,11 @@ def __init__(self, *args, **kwargs):
         )
 
     def _get_parent_queryset(self):
-        # Select non-pre-assessments that have been reviewed and rejected and
+        # Select proposals that have been reviewed and rejected and
```
RoordinkFrank, line 503: I am still not sure what exactly the difference is between RevisionProposalCopyForm and AmendmentProposalCopyForm . I assume for now that I should make this change for both of them.

EdoStorm96 (approved):
Nice catch that you found `is_revisable`! I'd forgotten all about it ...

Looks all good!

proposals/forms.py
miggol: Yes... I suppose preassessments can be amended as well now!

proposals/forms.py
```diff
@@ -459,9 +459,9 @@ def __init__(self, *args, **kwargs):
         self.fields["parent"].queryset = self._get_parent_queryset()
 
     def _get_parent_queryset(self):
-        # Return all non-pre-assessments, that are not currently in review
+        # Return all proposals, that are not currently in review
         return (
-            Proposal.objects.filter(is_pre_assessment=False)
+            Proposal.objects.filter()
```
miggol, line 464: Now that we're touching this, could you make this part of the `ProposalQuerySet` manager in `models.py`? What I want is that the manager provides the queryset of all revisable proposals, and this form only filters on the applicant part.

miggol (changes requested):
I haven't elaborately tested the copying mechanism, but assuming that works, this should be almost ready for merge. I found some small things that need to be addressed:

- Logic in email templates isn't designed for revised preassessments, for example `assignment_shortroute.txt`
- Emails to secretary will always be for a "New preassessment proposal" rather than reflecting the revised status
- It would be nice to use the Proposal QS manager to return revisable applications and the filter on that basis. But that's ultimately nice-to-have and not important

RoordinkFrank:
I have yet to add the tests. Something we discussed verbally. As for your comments I will look into them.

reviews/tests.py
```diff
@@ -151,6 +152,80 @@ def test_start_review(self):
...
+
+    def setUp(self):
+        super().setUp()
+        self.student: Relation = Relation.objects.get(pk=4)
+        self.language_sciences_pk = 1
+        self.direct_funding = Funding.objects.get(pk=2)
+        print("ReviewPreAssessmentTestCase: Creating pre-assessment proposal")
+        self.setup_pre_assessment()
+
+    def setup_pre_assessment(self):
```
RoordinkFrank, line 165: Follows the same setup as above. That said if we want to have more then one example data we may need to load this from an external file (SQL perhaps?)

Goes outside the scope of this issue but something to think about

reviews/tests.py
```diff
@@ -107,15 +108,78 @@ def refresh(self):
...
+
+class BasePreAssessmentTestCase(BaseReviewTestCase):
+    def setUp(self):
+        super().setUp()
+        self.student: Relation = Relation.objects.get(pk=4)
+        self.language_sciences_pk = 1
+        self.direct_funding: Funding = Funding.objects.get(pk=2)
+        self.setup_pre_assessment()
+
+    def setup_pre_assessment(self):
```
RoordinkFrank, line 130: If it turns out this data will be needed in all test classes we may just as well move this to BaseReviewTestCase. Still I am not sure if that's the case.

reviews/tests.py
```diff
@@ -107,15 +108,78 @@ def refresh(self):
...
+        self.pre_assessment.wmo = Wmo.objects.create(
+            proposal=self.pre_assessment,
+            metc=YesNoDoubt.NO,
+        )
+        # 3. Documents
+        # Optional File
+        # No Data yet added
+        # 4. Submit
+        # Space for possible comments. Use a maximum of 1000 words.
+        self.pre_assessment.comments = "Space for possible comments"
```
RoordinkFrank, line 177: Not sure if pre_assessment study is needed.

EdoStorm96 (changes requested):
Well this is a good start and there are some good ideas here. However this can be streamlined a lot. I'd like you to have a look at `proposal_submission_tests.py` and try to replicate the pattern seen there. This can make the code a lot more concise.

One of the main differences there is that the proposals are loaded from fixtures instead of created in the tests. I think you might be able to simply repurpose `BaseProposalTestCase` from the submission test file for the tests that are being run here. From the fixtures we already create a pre_assessment and a pre_approved proposal, so you can the also test pre_approved proposals while you're at it.

To test the different types we don't create multiple functions, but instead create a new subclass where we override self.proposal. I think this is pretty clean. It will be a bit of work to get these tests working with the setup from `BaseProposalTestCase`, but it will be very nice once it does become compatible! (In fact, I was surprised we do not repurpose `BaseProposalTestCase` for these tests already ...)

Let me know if you have any other questions and keep up the good work :)

proposals/forms.py
RoordinkFrank: Do you want something like this? And if so just for this method or for others as well?

proposals/models.py
```diff
@@ -90,6 +90,12 @@ def __str__(self):
 class ProposalQuerySet(models.QuerySet):
     DECISION_MADE = 55
 
+    def all_proposals_not_in_review(self):
+        return self.filter(
```
RoordinkFrank, line 95: the .distinct now happens in the middle of the filter instead at the end. I am also not 100% sure if it is needed here.

proposals/forms.py
miggol: What do you mean?

RoordinkFrank: This is about commit https://github.com/DH-IT-Portal-Development/ethics/pull/1001/commits/27c18da1a7dc41f6c2d8abd7cc8340638ef3d661

proposals/models.py
```diff
@@ -90,6 +90,12 @@ def __str__(self):
 class ProposalQuerySet(models.QuerySet):
     DECISION_MADE = 55
 
+    def all_proposals_not_in_review(self):
```
miggol: The name of this method obscures its semantics: `copyable_proposals()` would be better. This is where the queryset of copyable proposals lives.

proposals/models.py
miggol: I don't think it is needed, but maybe I'm missing something. What makes you think it is needed?

reviews/templates/mail/concept_supervisor.txt
```diff
@@ -11,7 +11,9 @@ Je kunt deze concept-aanvraag bekijken via {{ proposal_url }}. Je kunt deze, eve
 Als {{ creator }} de aanpassingen moet maken, dan moet je de aanvraag formeel afkeuren en zelf {{ creator }} hiervan op de hoogte stellen.
 
 Je kunt al je aanvragen als eindverantwoordelijke onderzoeker en hun status via {{ my_supervised }} bekijken.
-
+{% if is_pre_assessment %}
+Let op: het gaat hierbij om een pre-assessment.
+{% endif %}
```
miggol, line 16: This variable isn't provided by `start_supervisor_phase()`. I'm guessing you didn't test this?

Remember to also modify the HTML emails when editing the text emails :)

miggol (changes requested):

proposals/models.py
RoordinkFrank: agreed

proposals/models.py
RoordinkFrank: .distinct was in the original query. Except it was at the end and now in the middle. That is a change of logic, that is why I highlighted it. 

As for .distinct itself I am not sure if it is needed either.

edit: removed .distinct() to end of query as was the original logic. No changes to thing I do not fully understand why/if they are needed.

reviews/templates/mail/concept_supervisor.txt
RoordinkFrank: caught out. I indeed did not test this properly. Also good catch to have found this.

reviews/tests.py
```diff
@@ -76,6 +78,26 @@ def setup_proposal(self):
...
+            institution_id=1,
+        )
+        self.pre_assessment.is_pre_assessment = True
+        self.pre_assessment.applicants.add(self.user)
+        self.pre_assessment.wmo = Wmo.objects.create(
+            proposal=self.pre_assessment,
+            metc=YesNoDoubt.NO,
+        )
+        self.pre_assessment.supervisor = self.supervisor
+        self.pre_assessment.relation = Relation.objects.get(description_en=PHD_STUDENT)
```
EdoStorm96, line 98: I think just getting the relation via the PK is better ... It is more likely to be constant and it is how we roll for the rest of the tests. Then you can remove the test_constants file.

proposals/forms.py
```diff
@@ -459,19 +459,15 @@ def __init__(self, *args, **kwargs):
         self.fields["parent"].queryset = self._get_parent_queryset()
 
     def _get_parent_queryset(self):
-        # Return all non-pre-assessments, that are not currently in review
+        # Return all proposals, that are not currently in review
         return (
-            Proposal.objects.filter(is_pre_assessment=False)
+            Proposal.objects.copyable_proposals()
```
EdoStorm96: Right now I don't really like this logic being in a manager, as this only gets used in spot and now I have to visit a different file to see what is happening in that one spot ...

EdoStorm96: I do however see a use for a `can_be_copied_by(user)` filter, as that gets used in all these querysets? Maybe in more places.

That would replace this part:
```
            .filter(
                Q(
                    applicants=self.user,
                )
                | Q(supervisor=self.user)
            )
```
But you can do this if you want, it's not a must for me.

proposals/models.py
```diff
@@ -90,6 +90,12 @@ def __str__(self):
 class ProposalQuerySet(models.QuerySet):
     DECISION_MADE = 55
 
+    def copyable_proposals(self):
+        return self.filter(
```
EdoStorm96, line 95: As I mention in another comment, I think having a filter which only gets used in one spot is not so useful. I say, get rid of it.

proposals/models.py
```diff
@@ -681,8 +681,7 @@ def is_revisable(self):
         keep in mind to also check if the user is one of the applicants
         or the supervisor."""
         if (
-            not self.is_pre_assessment
```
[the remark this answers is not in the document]
EdoStorm96: Nice

reviews/tests.py
```diff
@@ -76,6 +81,26 @@ def setup_proposal(self):
...
+            title="p2",
+            reference_number=generate_ref_number(),
+            date_start=date.today(),
+            created_by=self.user,
+            reviewing_committee=Group.objects.get(
+                name=settings.GROUP_LINGUISTICS_CHAMBER
+            ),
+            institution_id=1,
+        )
+        self.pre_assessment.is_pre_assessment = True
```
EdoStorm96: Why is this outside of the creation? It can also be done during the creation right?

reviews/tests.py
```diff
@@ -401,6 +523,44 @@ def test_auto_review_registration_age_min(self):
         self.assertEqual(len(reasons), 2)
 
 
+class PreAssessmentAutoReviewTestCase(AutoReviewTests):
```
EdoStorm96: Pre assessments don't have a study, so I think this whole thing can go.

reviews/tests.py
```diff
@@ -409,6 +569,9 @@ class ReviewCloseTestCase(
 
     def setUp(self):
         super().setUp()
+        self.startReview()
+
+    def startReview(self):
```
EdoStorm96: Naming convention for python class methods is `lower_case_with_underscores`.

reviews/tests.py
```diff
@@ -493,6 +656,7 @@ def test_long_route(self):
         self.assertEqual(
             self.review.stage,
             self.review.Stages.CLOSED,
+            f"Pre-assessment proposal stage is incorrect. We expect {Review.Stages.CLOSED.name}, but instead we got {Review.Stages(self.review.stage).name}",
```
EdoStorm96: This is not PreAss right?

EdoStorm96 (changes requested):
There is some small stuff I'd like to see changed. Almost there :)

reviews/tests.py
RoordinkFrank: Switching between python en typescript is hard:(

proposals/forms.py
RoordinkFrank: I can see your point but I do get a bit of conflicting feedback here. I will take a look at can_be_copied_by(user).

image

reviews/tests.py
RoordinkFrank: Conflicting feedback here. Though that feedback was in feat/pre-assessment_tests I believe.

Edit found it:

image

reviews/tests.py
RoordinkFrank: This does make Sense.
I added a comment `# pre-assessment does not have a study so PreAssessmentAutoReviewTestCase should and not be added.` If you feel it's unnecessary it can be removed.

reviews/tests.py
RoordinkFrank: I have no idea why I wrote Pre-assessment there. It's not even a proposal stage

reviews/tests.py
RoordinkFrank: Done.

EdoStorm96 (approved):
Looks good!

PR #1002, opened 2025-09-26
refactor review views

RoordinkFrank:
Issue #938

If everything is done correct no logic should have been changed.

### DIFF
--- a/reviews/api/views.py
+++ b/reviews/api/views.py
@@ -19,7 +19,7 @@ from .serializers import DecisionSerializer, ReviewSerializer
 
 def return_latest_decisions(decisions: QuerySet[Decision]) -> list[Decision]:
     """
-        filters decisions with duplicate proposals to only give back the most recent decision for each proposal
+    filters decisions with duplicate proposals to only give back the most recent decision for each proposal
     """
     filtered_decisions: OrderedDict[int, Decision] = (
         OrderedDict()
@@ -128,8 +128,9 @@ class BaseDecisionApiView(GroupRequiredMixin, CommitteeMixin, FancyListApiView):
             # for this user, skip!
             # The template handles adding a different button for creating
             # a new decision for a secretary that doesn't have one yet.
-            if dfse_cache[proposal.pk] and decision.reviewer != self.request.user:
-                continue
+            if dfse_cache[proposal.pk]:
+                if decision.reviewer != self.request.user:
+                    continue
 
             if proposal.pk not in filtered_decisions:
                 filtered_decisions[proposal.pk] = decision
@@ -137,8 +138,8 @@ class BaseDecisionApiView(GroupRequiredMixin, CommitteeMixin, FancyListApiView):
                 if filtered_decisions[proposal.pk].pk < decision.pk:
                     filtered_decisions[proposal.pk] = decision
 
-        hello: list[Decision] = [value for key, value in filtered_decisions.items()]
-        return hello
+        decisions: list[Decision] = [value for key, value in filtered_decisions.items()]
+        return decisions
 
 
 class MyDecisionsApiView(BaseDecisionApiView):
--- a/reviews/api/views.py
+++ b/reviews/api/views.py
@@ -103,6 +103,60 @@ class BaseDecisionApiView(GroupRequiredMixin, CommitteeMixin, FancyListApiView):
 
         return context
 
+    def method_name(self, objects):
+        decisions = OrderedDict()
+        dfse_cache = {}  # Decision-for-secretary-exists cache.
+
+        for obj in objects:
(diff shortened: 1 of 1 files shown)


reviews/api/views.py
```diff
@@ -103,6 +104,45 @@ def get_context(self):
...
+                if dfse_cache[proposal.pk] and decision.reviewer == self.request.user:
+                    continue
+            else:
+                # If there is a decision for this user, but this decision isn't
+                # for this user, skip!
+                # The template handles adding a different button for creating
+                # a new decision for a secretary that doesn't have one yet.
+                if dfse_cache[proposal.pk] and decision.reviewer != self.request.user:
+                    continue
+
```
RoordinkFrank, line 137: I can not find out in what scenario the double seems to trigger. Does this not filtered in the queryset beforehand?

reviews/api/views.py
```diff
@@ -16,20 +17,21 @@
 from .serializers import DecisionSerializer, ReviewSerializer
```
RoordinkFrank, line 19: I can not find out why QuerySet is any here. It may have to do with my editor. Is this still any on vscode?

reviews/api/views.py
```diff
@@ -103,6 +105,41 @@ def get_context(self):
...
+                dfse_cache[proposal.pk] = Decision.objects.filter(
+                    review__proposal=proposal, reviewer=self.request.user
+                ).exists()
+
+            # If there is a decision for this user, but this decision isn't
+            # for this user, skip!
+            # The template handles adding a different button for creating
+            # a new decision for a secretary that doesn't have one yet.
+            if dfse_cache[proposal.pk] and decision.reviewer != self.request.user:
+                continue
```
RoordinkFrank, line 132: I can not get this to trigger. I feel like dfse_cache[proposal.pk] is redunant. and reviever != request.user should be in the query?

reviews/api/views.py
```diff
@@ -103,6 +105,41 @@ def get_context(self):
 
         return context
 
+    # filters decisions with duplicate proposals to only give back the most recent decision for any given proposal
+    def filter_decisions(self, decision_queryset: QuerySet[Decision]) -> list[Decision]:
+        filtered_decisions: OrderedDict[int, Decision] = (
+            OrderedDict()
+        )  # proposal.pk, decision
+        # Decision-for-secretary-exists cache.
+        dfse_cache: dict[int, bool] = (
```
RoordinkFrank, line 114: I have been wondering for a while why it checks for existence. It should not get a non existing decision as far as I can see.
Unless this method does not trust the QuerySet input and this is added security. If so why does this security not exist for the query's that call return_latest_decisions?

reviews/api/views.py
```diff
@@ -16,20 +17,21 @@
 from .serializers import DecisionSerializer, ReviewSerializer
 
 
-def return_latest_decisions(objects):
+def return_latest_decisions(decisions: QuerySet[Decision]) -> list[Decision]:
     """
```
RoordinkFrank, line 21: Until now I have only refactored and if everything is done correct. I do wonder however if this logic could be a query. It should be faster that way.

reviews/api/views.py
```diff
@@ -103,6 +105,42 @@ def get_context(self):
...
+                    review__proposal=proposal, reviewer=self.request.user
+                ).exists()
+
+            # If there is a decision for this user, but this decision isn't
+            # for this user, skip!
+            # The template handles adding a different button for creating
+            # a new decision for a secretary that doesn't have one yet.
+            if dfse_cache[proposal.pk]:
+                if decision.reviewer != self.request.user:
+                    continue
```
RoordinkFrank, line 133: if dfse_cache always is True and and next if statement is always false for me.

reviews/api/views.py
```diff
@@ -270,16 +263,17 @@ def get_queryset(self):
             # for this user, skip!
             # The template handles adding a different button for creating
             # a new decision for a secretary that doesn't have one yet.
-            if dfse_cache[proposal.pk] and obj.reviewer == self.request.user:
+            if dfse_cache[proposal.pk] and decision.reviewer == self.request.user:
```
RoordinkFrank, line 266: This method is nearly identical as well, this if statement is now == instead of !=. Still comment is confusing.

RoordinkFrank:
> Hey, what are we looking at here? Just some renaming of variables?
> 
> A description of this PR's goals would be helpful! And is there an issue related to this?


Issue #938, I should have posted that in the initial description instead of linking the issue.

That said the more I look at it the more I suspect I should just close this branch entirely. The refactoring did not go as expected and sure I removed some duplicate code but it may end up more confusing then it was.

PR #1007, opened 2025-10-15
Feat/faster tests

RoordinkFrank:
branched from #1001

I noticed tests getting from 24s to 8s during pre-assessment as soon as test data transferred to fixtures. Those 8s are with the added pre-assessment tests.

After feedback that  this pull request is to big I will try to split this up into different branches. This branch will not change but as other branches are slowly be merged into develop this one should become smaller.

Code with a different branch:
- [x] Pre-Assessment tests. Branch is  #1016, not dependent on the branches below. 
- [x] Proposal tests are faster as a lot of code has been replaced with fixtures.  Branch is #1009
- [x] Review tests are faster as a lot of code has been replaced with fixtures.  Branch is #1010. 
- [x] Test data from review and proposal tests have been merged Branch is #1011. Dependency to "feat/proposal tests additional fixtures" and "Review tests are faster". The two checkboxes just above.
- [x] unused code such as generate pdf has been removed Branch is #1012

These branches are branched from each other from top to bottom. To merge test data this is needed.

-----------------------------------
image

To clarify, these helper methods are mostly methods that already did exist. Just moved up to parent classes. That or duplicate code refactored to a method.


reviews/tests.py
```diff
@@ -526,3 +576,17 @@ def test_metc(self):
...
+    def setUp(self):
+        super().setUp()
+        self.proposal = self.pre_assessment
+
+    def startReview(self):
+        # super().startReview()
+        self.review = start_review(self.pre_assessment)
+
+    def test_long_route(self):
+        pass
```
RoordinkFrank: test is failing here. pre assessment should not reach long route but I do not understand everything here.

reviews/tests.py
```diff
@@ -526,3 +576,17 @@ def test_metc(self):
...
+
+class PreAssessmentReviewCloseTestCase(ReviewCloseTestCase):
+
+    def setUp(self):
+        super().setUp()
+        self.proposal = self.pre_assessment
+
+    def startReview(self):
+        # super().startReview()
+        self.review = start_review(self.pre_assessment)
```
RoordinkFrank, line 526: setUp review -> setUp actual tests -> setUp pre assessment is the order that triggers and it has to remain this way. Yet I do want review -> pre assessment -> actual tests which is imposible in current setup so i have to override startReview here.

Possible solution is the same as in one of my previous comments about proposal inheriting the same was as preAssessment is doing now.

reviews/tests.py
```diff
@@ -283,43 +288,68 @@ def test_auto_review(self):
...
 
         self.study.risk = YesNoDoubt.YES
         self.study.save()
 
         reasons = auto_review(self.proposal)
         self.assertEqual(len(reasons), 6)
+        self.assertEqual(
+            reasons[-1],
+            "De onderzoeker geeft aan dat er mogelijk kwesties zijn rondom de veiligheid van de deelnemers tijdens of na het onderzoek.",
+        )
```
RoordinkFrank, line 270: changing self.study.proposal to self.proposal makes it possible to load studies with fixtures. Why this is needed I still do not know, but it works.

reviews/tests.py
```diff
@@ -17,95 +17,49 @@
...
 
     def setUp(self):
-        """
-        Sets up the Users and a default Proposal to use in the tests below.
-        """
-        self.setup_users()
-        self.setup_proposal()
         super().setUp()
 
     def setup_proposal(self):
```
RoordinkFrank: generate pdf removed here. It was not actually used and significantly slowed down tests as it was done for every test. If it is needed somewhere in the future I propose adding it locally.

reviews/tests.py
```diff
@@ -264,62 +247,96 @@ def test_decision_commission(self):
...
-            description="psychofysiologische meting (bijv. EEG, fMRI, EMA)"
-        )
+        self.study = Study.objects.get(proposal=f"{self.proposal.pk}")
+
+    def test_auto_review_pre_assessment_short_route(self):
+        reasons = auto_review(self.proposal)
+        self.assertEqual(len(reasons), 0)
 
     def test_auto_review(self):
+        if self.proposal.is_pre_assessment:
```
RoordinkFrank, line 209: return statement is not great here as the pre assessment tests still make a setup for these tests that are basically not used as these tests are only meant for base proposal.

Only solution I can think of is by letting the base proposal inherit in the same way as pre assessment now inherits AutoReviewTests. And adding the normal proposal specific tests there.

reviews/tests.py
```diff
@@ -17,95 +17,40 @@
...
-        "compensations",
-        "00_registrations",
-        "01_registrationkinds",
-        "agegroups",
-        "groups",
-        "institutions",
-    ]
+class BaseReviewTestCase(BaseProposalTestCase):
     relation_pk = 1
+    review = None
```
RoordinkFrank, line 27: relation_pk seems no longer used.

proposals/tests/proposal_submission_tests.py
```diff
@@ -44,60 +47,43 @@ def setUp(self):
...
+
+    def add_supervisor_to_proposal(self):
+        self.proposal.relation = Relation.objects.get(
+            description_nl="als AIO / promovendus verbonden"
+        )
+        self.proposal.supervisor = self.supervisor
+        self.proposal.save()
+
+    def remove_study(self):
+        Study.objects.filter(proposal=f"{self.proposal.pk}").delete()
```
RoordinkFrank, line 86: Seems to be needed in two tests.

proposals/tests/proposal_submission_tests.py
```diff
@@ -44,60 +47,53 @@ def setUp(self):
...
-            metc="N",
-            is_medical="N",
+        self.proposal = Proposal.objects.get(
+            reference_number="24-009-01", title="Normal test proposal"
+        )
+        self.pre_assessment = Proposal.objects.get(
+            reference_number="24-010-01", title="Preassessment test proposal"
+        )
+        self.pre_approval = Proposal.objects.get(
+            reference_number="24-011-01", title="Preapproved test pdf"
```
miggol, line 67: Just in passing, I must say I disagree with black on this one...

proposals/tests/proposal_submission_tests.py
```diff
@@ -44,60 +47,53 @@ def setUp(self):
...
+
+    def add_supervisor_to_proposal(self):
+        self.proposal.relation = Relation.objects.get(
+            description_nl="als AIO / promovendus verbonden"
+        )
+        self.proposal.supervisor = self.supervisor
+        self.proposal.save()
+
+    def remove_study(self):
+        Study.objects.filter(proposal=f"{self.proposal.pk}").delete()
```
miggol, line 86: Does `self.proposal.study_set.all().delete()` not work? As written your version seems a bit omslachtig. Also it removes all studies so best to call it `remove_studies()` (plural)

I would also be down with removing this method entirely. Calling it doesn't even save a single line.

proposals/tests/proposal_submission_tests.py
```diff
@@ -44,60 +47,53 @@ def setUp(self):
...
         self.proposal.refresh_from_db()
+        self.pre_assessment.refresh_from_db()
+        self.pre_approval.refresh_from_db()
+
+    def add_supervisor_to_proposal(self):
+        self.proposal.relation = Relation.objects.get(
+            description_nl="als AIO / promovendus verbonden"
+        )
+        self.proposal.supervisor = self.supervisor
+        self.proposal.save()
```
miggol, line 83: I don't like that this method implicitly changes the relation. I would rather keep these steps explicit (and in the test) so that if those tests fail, it's easier to see.

miggol (changes requested):
I understand that this PR is branched from a different PR, but once again this is a PR that addresses various tangentially related things without any issues or stated goals.

As far as I can tell the following distinct things are going on here:

1. Added fixtures rather than explicit Proposal/user creation.
2. Merger of Proposal and Review base classes
3. Other changes/refactoring in SetUp, such as new helper methods
4. New tests added
5. Fixes for auto review test and test_long_route (?)
6. Remove generate PDF

Don't get me wrong, I'm a big fan of the merger of the Proposal an Review testcases in particular. But this borders on impossible to review. For me to Approve this, it will have to be in smaller chunks.

* (1) and (2) should be separate PRs into develop
* (3) can be dropped for now
* (4) should be in this PR, because it adds a test for `Feat/pre-assessment_get_revisions` which the branch will merge into. That is a good stated goal for this specific PR, stick to that.
* If (5) contains fixes for bugs or failing tests, please describe them and make issues for them before fixing in a separate PR. If these fixes are absolutely required for merging (4), I guess they can be in this PR, but in that case please precisely describe what you're doing.
* (6) can also be dropped for now, but could be chucked together with (3) in a future PR

RoordinkFrank:
> I understand that this PR is branched from a different PR, but once again this is a PR that addresses various tangentially related things without any issues or stated goals.
> 
> As far as I can tell the following distinct things are going on here:
> 
>     1. Added fixtures rather than explicit Proposal/user creation.
> 
>     2. Merger of Proposal and Review base classes
> 
>     3. Other changes/refactoring in SetUp, such as new helper methods
> 
>     4. New tests added
> 
>     5. Fixes for auto review test and test_long_route (?)
> 
>     6. Remove generate PDF
> 
> 
> Don't get me wrong, I'm a big fan of the merger of the Proposal an Review testcases in particular. But this borders on impossible to review. For me to Approve this, it will have to be in smaller chunks.
> 
>     * (1) and (2) should be separate PRs into develop
> 
>     * (3) can be dropped for now
> 
>     * (4) should be in this PR, because it adds a test for `Feat/pre-assessment_get_revisions` which the branch will merge into. That is a good stated goal for this specific PR, stick to that.
> 
>     * If (5) contains fixes for bugs or failing tests, please describe them and make issues for them before fixing in a separate PR. If these fixes are absolutely required for merging (4), I guess they can be in this PR, but in that case please precisely describe what you're doing.
> 
>     * (6) can also be dropped for now, but could be chucked together with (3) in a future PR

After feedback that  this pull request is to big I will try to split this up into different branches. This branch will not change but as other branches are slowly be merged into develop this one should become smaller.

Up to date status is at the first comment of this pull request.

reviews/tests.py
RoordinkFrank: moved to #1016

RoordinkFrank:
This PR has been resolved in the branches mentioned at the top. This is now an empty husk and will be closed.

PR #1009, opened 2025-10-24
feat/proposal tests additional fixtures

RoordinkFrank:
Branched from pre assessment tests. Replaces a part of #1007 

Goal of this pr is to make the proposal tests faster while keeping logic unchanged.

1. Moved all users to the fixtures including group assignment. Tests go from 24 to 18s with this change.
2. Applicants have been moved to the proposal fixtures. Tests go from 18s to 18s with this change.
3. The first proposal now calls the WMO by fixture instead of creating an object. Tests go from 18s to 18s with this change.
4. Removed the .save() functions in setup_proposal as there is nothing to save left. Tests go from 18s to 18s with this change.

Affected code:

1. Affected code setup_users() and test_users.json. 
2. Affected code is setup_proposal.
3. Affected code is setup_proposal and test_wmos.json
4. Affected code is setup_proposal

Proposal names stay as pk 
image
As @EdoStorm96 has conveyed verbally that this is his preference over the Feat/faster_tests setup of
image
The reasoning is that the pk is less likely to change then the title or reference number.
Pull requests after this will also have this change so long there is information what you are pulling from the test db. In this case you can still see that information with the self.proposal or self.pre_assessment.

Edit:
After verbal contact with @miggol we have to come to the conclusion to use names/description instead of pk's. Pkś are indeed less likely to change but they have a lot of downsides. The names/description will becomes a constant so that changing the name/description once it needs to change is way easier. This way we will hopefully have both of best worlds.


proposals/tests/proposal_submission_tests.py
```diff
@@ -44,54 +46,23 @@ def setUp(self):
...
-
-        Please note, this currently uses a mish-mash of both fixtures
-        and programmatically created users and objects.
+        Load our test proposals from a fixture.
         """
-        self.wmo = Wmo(
-            status=0,
-            metc="N",
-            is_medical="N",
+        self.proposal = Proposal.objects.get(reference_number=PROPOSAL_REF_NUMBER)
```
RoordinkFrank: ref_number is the unique identifier but not really readable by itself so it is combined with the title, if ref number changes by accident (just as pk) we will now know.

Without the ref_number I do not know what happens if we have two proposals with the same title.

proposals/tests/proposal_submission_tests.py
```diff
@@ -44,54 +46,23 @@ def setUp(self):
...
-        )
-
-        self.secretary.groups.add(
-            Group.objects.get(name=settings.GROUP_PRIMARY_SECRETARY)
-        )
-        self.secretary.groups.add(Group.objects.get(name=settings.GROUP_SECRETARY))
-
-        self.c1.groups.add(Group.objects.get(name=settings.GROUP_LINGUISTICS_CHAMBER))
-        self.c2.groups.add(Group.objects.get(name=settings.GROUP_LINGUISTICS_CHAMBER))
+        self.secretary = User.objects.get(username=SECRETARY)
```
RoordinkFrank, line 49: username and pk both are an unique identifier. It is consistent with the rest to use constants here but it feels a bit redunant as username was already very readable and declared only once in the setup_users.

EdoStorm96 (approved):
Looks good to me :)

PR #1010, opened 2025-10-24
Feat/review tests added fixtures

RoordinkFrank:
Branched from #1009 . Replaces a part of #1007 

Goal of this pr is to make the review tests faster while keeping logic unchanged.

1. Moved all users to the fixtures including group assignment. Tests go from 18s to 8s with this change.
2. Moved review proposal to fixture. Tests go from 8s to 8s with this change.
3. Moved study to fixture. The study now has an agegroup as a study without agegroup should not exist. Tests results remain unchanged with this. Small changes as well (see below). Test go from 8s to 8s with this change.

Affected code:

1. setup_users(). Original logic was a copy paste of proposal tests so fixtures can be reused.
2. setup_proposal(). Fixture is a copy of fixture from pk=1 except it has a supervisor and relation is 4
3. I got the "AssertionError: 6 != 7" once again when moving the study to a fixture so I also added the Asserts in autoreview to get more clarity, it does make changing reasons later on harder but without it you are not actually asserting if you actually added the right reason.
Solution to the AssertionError was changing self.study.proposal to self.proposal. Still no idea why this works. This is something that only appeared when moving the study to the fixture so it should be in this pr.

image

---------------------------------------------------------------------------------------------------------------
WMO will not be moved to  fixtures in this PR. Moving WMO to fixtures should be part of the pr but I have some technical problems. I have an error referring a WMO with pk=5.
image
Except there is no pk=5 in WMO and it also shouldn´t be called.
That said I accidentally fixed this problem in #1007 so WMO will be added to fixtures later on.


reviews/tests.py
```diff
@@ -330,42 +297,66 @@ def test_auto_review(self):
 
         reasons = auto_review(self.proposal)
         self.assertEqual(len(reasons), 1)
+        self.assertEqual(
```
RoordinkFrank, line 300: I made these asserts trying to find out why some tests were failing. The problem didn't end up being here but at least it's more clear this way.

That said it is less refactor able this way while they are not exactly critical. If you want these asserts removed then that is totally fine.

---------

also I have no idea why but this change is absolutely necessary. Even though it should be the exact same code.

Image

proposals/fixtures/testing/test_proposals.json
```diff
@@ -211,5 +211,76 @@
             1, 5
         ]
     }
+},
+{
```
RoordinkFrank, line 215: Separate fixture added so the PR remains simple. The next PR #1011 merges the two together. Both PR's remain smaller this way. 

pk=4 is exactly the same as pk=1 except it has a supervisor and study.

--------------

Pre-Assignment fixture for review is missing. That should be automatically fixed in # 1011 when merging. It doesn´t add anything to make a `pk=5 pre-assessment proposal with supervisor` right now.

PR #1011, opened 2025-10-27
feat/merge_proposal&review_test_data

RoordinkFrank:
Branched from #1010 . Replaces a part of #1007. This PR is also the reason why the branches are dependent on each other.

Goal of this pr is to merge the test data together as they have massive overlap.
Note: from this point on test data will change when it is necessary to merge the data together. It is not a pure refactor like the previous two branches.

All users are already the same. The proposals differ in that the proposal in the proposal_submission_tests start out without a supervisor and add a supervisor if it is needed for the test. The review tests do the opposite, they start with a supervisor and remove the supervisor when necessary for the test. To merge the data review tests need to have the same behavior as proposal tests.

Also review tests do have a study. Something that every proposal should have as far as I understand but the proposal_submission_tests currently do not have one. The merged proposal will start with a study.

This may be a bit hard to follow so we are going commit by commit.

1. BaseReviewTestCase inherits from BaseProposalTestCase now. Goal is to merge them as much together as possible. we are at: https://github.com/DH-IT-Portal-Development/ethics/pull/1011/commits/ee0eda3581cc843b131dc95f892c41507e29d619
2. Proposal=pk4 now no longer has a supervisor. Code that did remove the supervisor before starting the test has been remove and code has been added to add a supervisor where needed. We are at https://github.com/DH-IT-Portal-Development/ethics/pull/1011/commits/418f90e24fa1173728a38de528793e8865491d58
3. The review Proposal=pk4 has been removed. self.proposal is now pk=1 in both the review tests and setup proposal tests. We are at https://github.com/DH-IT-Portal-Development/ethics/pull/1011/commits/83fc99429f25fd16d55f995b67870d0e4baec2d2
4. So far every data has been merged except studies and refresh(). review tests.Refresh() can not be merged and remains necessary. Studies can be merged but doing so causes two methods in proposal_submission_tests to need to have a delete studies line because they don´t like having a study for some reason. We are at https://github.com/DH-IT-Portal-Development/ethics/pull/1011/commits/c6cb3cf20b40bdfff92f672b99c66bba8a114000
5.  And finally methods have been moved up to ProposalTestCase whenever possible. Makes them reuseable for all tests. 

affected code:

1. BaseReviewTestCase. setup_users has been removed as the method is the same in BaseReviewTestCase. setup_proposals and refresh call the super() method first before doing their thing.
2. Every method with `self.proposal.relation = Relation.objects.get(pk=5)`. Every method that adds a supervisor. The proposal pk=4 in the fixture.
3. setup_proposal and proposal fixtures. The WMO problem in #1010 has been removed as well. Once again by accident.
4. BaseReviewTestCase (removed fixtures as they are no longer needed, everything is loaded in Proposal tests). `test_submission_unsupervised(self)` and `test_proposal_supervised(self):` needed to have data with deleted studies.
5. `BaseReviewTestCase`, `ProposalSubmitTestCase.test_proposal_supervised`

------------------
image

Method is now called 
image

It was more correct to say that the relation changed instead of the supervisor. The argument is there to make it clear the supervisor is involved. Are you ok with this change or should the entire method be inlined?

------------------------
image
Changed


proposals/tests/proposal_submission_tests.py
```diff
@@ -113,6 +133,7 @@ def test_submission_unsupervised(self):
         - Because there is no supervisor, a new review is created
           in the assignment stage.
         """
+        self.proposal.study_set.all().delete()
         # Sanity checks to start
```
RoordinkFrank, line 146: kom ik op terug

reviews/tests.py
```diff
@@ -17,94 +14,27 @@
...
-        "fundings",
-        "agegroups",
-        "groups",
-        "institutions",
-        "testing/test_users",
-        "testing/test_proposals",
-        "testing/test_studies",
-    ]
+class BaseReviewTestCase(BaseProposalTestCase):
     relation_pk = 1
```
EdoStorm96, line 26: Can you remove this?

EdoStorm96 (approved):
One tiny comment :)

PR #1012, opened 2025-10-28
Feat/test cleanup

RoordinkFrank:
branches from #1011, part of #1007

Small cleanup that didn´t belong in any of the other pull requests. Biggest one is the removal of pdf generation. It is logical for proposals to generate pdf's but it wasn't actually being used. If it is ever needed I propose to generate it in the setup of one of the test child classes, not in the Proposal or Review setup.
Removing this speeds up the tests by about two seconds.

Refresh is now refreshes all proposals as you would expect the method to do.


reviews/tests.py
```diff
@@ -23,11 +23,7 @@
 
 
 class BaseReviewTestCase(BaseProposalTestCase):
-    relation_pk = 1
-
-    def setup_proposal(self):
-        super().setup_proposal()
-        self.proposal.generate_pdf()
+    review = None
```
RoordinkFrank: mag eruit

reviews/tests.py
```diff
@@ -23,11 +23,7 @@
 
 
 class BaseReviewTestCase(BaseProposalTestCase):
-    relation_pk = 1
-
-    def setup_proposal(self):
-        super().setup_proposal()
-        self.proposal.generate_pdf()
```
RoordinkFrank, line 30: check if pdf works test needed.

EdoStorm96 (approved):

Issue #966, opened 2025-07-10
double this field is required


RoordinkFrank:
<img width="959" height="709" alt="Image" src="https://github.com/user-attachments/assets/b64a8add-41c5-415c-b1af-8abb437f49fa" />
at step

<img width="959" height="709" alt="Image" src="https://github.com/user-attachments/assets/46eb0f68-655b-4817-9010-9f6f6b658f0f" />

RoordinkFrank:
I came across this in another Issue, the second field is required is caused by mark_soft_required. I still do not fully understand where the first required field comes from and why mark_soft_required only sometimes gives the double error but removing 
`self.mark_soft_required(cleaned_data, "student_justification")` in proposals/forms.py would remove this particular double error.

PR #1015, opened 2025-11-07
Added Detail Management Details

### LINKED ISSUE #1014: Changes to Question about consultation data manager.
New requests about the question that was added in #887.
> It turns out that researchers sometimes did discuss their application and documents with the p.o. and/or d.m., but not the versions that were submitted. To make the questions more specific, could these be changed to:
> 
> O De documenten voor deelnemers die ik hieronder indien zijn besproken met en gezien door de [privacy officer](privacy.gw@uu.nl) van de faculteit Geesteswetenschappen. Als dat niet gebeurd is, zou je dan willen toelichten waarom niet? Bijv. omdat vergelijkbare documenten van een eerdere aanvraag al zijn besproken met de privacy officer (vermeld in dat geval s.v.p. het referentienummer).
> 
> [comments field, free text, max. 500 words?]
> 
> O Het Data Management Plan (DMP) dat ik hieronder indien is besproken met en gezien door de [datamanager](datamanagement.gw@uu.nl) van de faculteit Geesteswetenschappen.
> 
> Als dat niet gebeurd is, zou je dan willen toelichten waarom niet? Bijv. omdat vergelijkbare documenten van een eerdere aanvraag al zijn besproken met de datamanager (vermeld in dat geval s.v.p. het referentienummer).
> 
> [comments field, free text, max. 500 words?]
> 
> O Het Data Management Plan (DMP) dat ik hieronder indien is besproken met en gezien door iemand van Research Data Management Support. Als dat niet gebeurd is, zou je dan willen toelichten waarom niet? Bijv. omdat (vergelijkbare) documenten van een eerdere aanvraag al zijn besproken met Research Data Management Support of de datamanager (vermeld in dat geval s.v.p. het referentienummer).
> 
> [comments field, free text, max. 500 words?]
------------------------
> Suggestion to replace 'hieronder' with 'in de volgende stap'. 

Let's discuss any further information regarding this here.

RoordinkFrank:
Fixes #1014

TODO List:
- [x] Add details to the models
- [x] Correct translations
- [x] Communicate redraft of questions, links only working in help_text
- [x] Front end conditional checks to have text field appear.
- [x] Back end conditional checks on text field.

This i my first time adding a condition field so I am going to have to do some research.

--------------------------------------------
### Front end checks
It seems that for the front end we need to add javascript to check if a form needs to appear or not.
Adding these functions removes the detail fields
image

The only problem is that they will not appear again after selecting no or yes so it works but only on loading and doen´t seem to do the second step, Whether it loads with a yes or no doesn't seem to matter either.

Edit: Just needed to change N to False.

--------------------------------------------
### Back end checks

In the clean() function there are multiple background checks, copied from other similar forms. The methods themself have no documentation. I was hoping they would add front end conditions but that doesn't seem the case. They are likely not_empty or right values checks but it is unlikely I will need these. 
image

Edit: removed these as they are not needed.


proposals/forms.py
RoordinkFrank: Only the conversation trigger on this but not the conversation_details. The conversation_details are supposed to be optional so that is good. Is it because a textfield is "" and not None?

RoordinkFrank:
Do note that make message generate a fuzzy on study start. This does not have to do with this pull request but it is changed now as well.

EdoStorm96 (commented):
Good job on figuring this out! It looks pretty good in my opinion, but I just would like to discuss this in person with the secretary. She says she wants this, but I really don't like what this adds to the portal and I kindoff want to protect out users from encountering this every time they fill in a form. So I'll give a full review when we discussed this again with the secretary.

RoordinkFrank:
> Good job on figuring this out! It looks pretty good in my opinion, but I just would like to discuss this in person with the secretary. She says she wants this, but I really don't like what this adds to the portal and I kindoff want to protect out users from encountering this every time they fill in a form. So I'll give a full review when we discussed this again with the secretary.

I will put this on hold until the meeting.

locale/en/LC_MESSAGES/django.po
```diff
@@ -2016,39 +2019,69 @@ msgstr ""
...
-#: proposals/models.py:459
+#: proposals/models.py:460
 msgid ""
-"Ik heb mijn aanvraag en de documenten voor deelnemers besproken met de "
-"privacy officer."
+"De documenten voor deelnemers die ik in de volgende stap indien zijn "
+"besproken met en gezien door de privacy officer van de Faculteit "
+"Geesteswetenschappen."
+msgstr ""
+"The documents for participants that will be submitted in the next step has "
```
EdoStorm96:
```suggestion
"The documents for participants that will be submitted in the next step have "
```

locale/en/LC_MESSAGES/django.po
```diff
@@ -2016,39 +2019,69 @@ msgstr ""
...
-"Ik heb mijn Data Management Plan (DMP) besproken met de data manager van de "
-"faculteit Geesteswetenschappen."
+"Zou je dan willen toelichten waarom niet? Bijv. omdat (vergelijkbare) "
+"documenten van een eerdere aanvraag al zijn besproken met Research Data "
+"Management Support of de datamanager (vermeld in dat geval s.v.p. het "
+"referentienummer)."
 msgstr ""
-"I discussed my application and Data Management Plan (DMP) with the data "
-"manager of faculty of Humanities."
+"Would you like to explain why not? For example because (simular) documents "
```
EdoStorm96:
```suggestion
"Would you like to explain why not? For example because (similar) documents "
```

locale/en/LC_MESSAGES/django.po
```diff
@@ -2016,39 +2019,69 @@ msgstr ""
...
+"from an earlier application have already been discussed with the Research "
+"Data Management Support or the Datamanager. In that case please write down "
```
EdoStorm96:
```suggestion
"Data Management Support or the Datamanager. In that case please mention "
```

proposals/forms.py
EdoStorm96: This needs to be updated. It does work but it is more so because of luck than intention ... I would keep and error check like this for the fields we already had and then check that the new fields are not empty if the applicant answered the original questions with "no".

EdoStorm96: I'd change this to:
```suggestion
        for field in ["privacy_officer_conversation",  "data_manager_conversation", "research_data_management_conversation"]:
```
or something ...

EdoStorm96 (changes requested):
This looks good. I would like you to improve the form validation. You have not made validation in the backend, but this should be added as it is used for validating the entire proposal.

Also, you forgot to add these new fields in the pdf, with the right conditionality. You could do something similar to line 512-515 of StudyOverviewSection if you want to have some fun, but maybe make it more readable then what I did there ...

proposals/forms.py
RoordinkFrank: Fair enough, it already functions right now but to a future coder it's not clear what the intentions are.

RoordinkFrank:
> This looks good. I would like you to improve the form validation. You have not made validation in the backend, but this should be added as it is used for validating the entire proposal.
> 
> Also, you forgot to add these new fields in the pdf, with the right conditionality. You could do something similar to line 512-515 of StudyOverviewSection if you want to have some fun, but maybe make it more readable then what I did there ...

Good call on the pdf.

TODO:

- [x] Change clean method, this already should be the backend validation right?
- [x] add to pdf
- [x] commit textual changes
- [x] additionally: fix the three additional questions being the same.

proposals/utils/pdf_diff_sections.py
```diff
@@ -628,10 +629,26 @@ class DMPSection(PageBreakMixin, BaseSection):
...
         "data_manager_conversation",
+        "data_manager_conversation_details",
         "research_data_management_conversation",
+        "research_data_management_conversation_details",
     ]
 
+    def get_row_fields(self):
+        return self.remove_empty_row_fields(self.obj, self.row_fields)
+
+    @staticmethod
```
RoordinkFrank: Image

Is this what you roughly wanted? I have opted to leave de details field empty if the user has access to them but still opted to leave them empty. Logic is reusable in 512 I think?

proposals/forms.py
EdoStorm96, line 773: So, you are still not validating that the details fields have been filled in, if the boolean question have been answered with False, which is what we want right?. That is what I meant when I said add validation to the backend. Have a look at `StudyEndForm` for a (sortoff) nice way to do this with a loop.

proposals/utils/pdf_diff_sections.py
EdoStorm96: It does not really seem to work the same as the logic in 512. That logic is admittedly pretty hard to read ... But what it is meant to do, is to remove the details field, if the base question has been answered in a way where the details question does not need to be filled in. There are other ways to achieve this. I guess this logic will be hard to map 1-to-1 for these two section, as here we are working with booleans and the other one works with string, but you could apply a similar principle. Of course the easier (and more legible way) to do this is just:
```
if not getattr(self.obj, "privacy_officer_conversation":
    rows.remove("privacy_officer_conversation_details")
etc ...
```
This is just a bit repetitive, but also totally fine. I was just having some fun with that other one.

I am a bit confused why we would check for empty strings, as if the details need to be filled in, we won't allow them to be left empty right? It is possible now (due to the lack of backend validation for the details-fields) but I'm pretty sure the secretary wants something to get filled in, if the person answers "No".

proposals/utils/pdf_diff_sections.py
```diff
@@ -628,10 +628,26 @@ class DMPSection(PageBreakMixin, BaseSection):
...
+    def remove_empty_row_fields(obj, row_fields: list[str]) -> list[str]:
+        """removes empty string fields and returns the list"""
+        for field in row_fields:
```
EdoStorm96: I am assuming you haven't tested this? Looping over a list that you are modifying in the loop is tricky business. You can try running this example to see why this is a bad idea:
```
l = [1,2,3,4,5,6,7,8,9]

for num in l:
    print(num)
    if num in [2,3,8]:
        l.remove(num)
        
print(l)
```
It's best to create an empty list, and fill that with your desired result, so that the loop will keep working as expected.

proposals/utils/pdf_diff_sections.py
```diff
@@ -628,10 +628,26 @@ class DMPSection(PageBreakMixin, BaseSection):
...
+            value = getattr(obj, field)
+            if isinstance(value, str):
+                if len(value) == 0:
```
EdoStorm96: `if not value` or `if value == ""` would be nicer, though again, I think checking for empty strings is not what we want here.

EdoStorm96 (changes requested):
This is still not really what I meant. From how I understand it, if the users answers "No" on one of the yes-no questions, they need to fill in the details right? Now, you don't check for this yet. This is also the logic that I think think you should use for the pdf. Not checking for empty string (as they won't be possible if we implement the validation properly).

proposals/forms.py
RoordinkFrank: Details is an optional field.

proposals/utils/pdf_diff_sections.py
```diff
@@ -628,10 +628,38 @@ class DMPSection(PageBreakMixin, BaseSection):
...
+
+    def remove_unused_conditional_fields(self, conditions: list) -> list[str]:
+        """assumes all conditional fields end with '_details'.
+        @condition: removes details when non _details field is condition value
+        """
+        rows_to_remove: list[str] = []
+        for index, conditional_field in enumerate(self.conditional_fields):
+            row_field: str = conditional_field.removesuffix("_details")
+            value = getattr(self.obj, row_field)
+            if isinstance(value, bool):
```
RoordinkFrank: isinstance is not needed but it does gives a bit of clarity what we are working with.

proposals/utils/pdf_diff_sections.py
```diff
@@ -628,10 +628,38 @@ class DMPSection(PageBreakMixin, BaseSection):
...
+        field for field in row_fields if field.endswith("_details")
+    ]
+
+    # right now all boolean conditions are true in DMP
+    conditions: list = [True for field in conditional_fields]
+
+    def get_row_fields(self):
+        return self.remove_unused_conditional_fields(self.conditions)
+
+    def remove_unused_conditional_fields(self, conditions: list) -> list[str]:
```
RoordinkFrank: I already went way overboard. Changing this method so it can handle optional fields in other classes as well really needs to be a separate issue assuming we even want this.

I am also seriously considering reverting this method to

Image

so the code is consistent everywhere and a new person does not need to learn the code twice. Current code can be used in a new issue if we want that. Current code does have some advantages but not enough compared to consistency?

proposals/utils/pdf_diff_sections.py
```diff
@@ -628,10 +628,25 @@ class DMPSection(PageBreakMixin, BaseSection):
...
     ]
 
+    def get_row_fields(self):
+        return self.remove_unused_details_fields()
+
+    def remove_unused_details_fields(self) -> list[str]:
+        """assumes all conditional fields end with '_details'."""
+        rows_to_remove: list[str] = []
+        for row_field in self.row_fields:
+            # getattr always needs to be a bool that is True to be removed in DMP.
```
RoordinkFrank: This now gets more input than just bool btw, such as a _details field which is not a bool so never True. Text and Int field just get ignored. I am fairly certain without the isinstance check it will crash with other fields. May matter in case this code gets used in other places?

proposals/utils/pdf_diff_sections.py
RoordinkFrank: No longer relevant.

proposals/utils/pdf_diff_sections.py
RoordinkFrank, line 201: I have been looking here as well. registration_details and registration_kinds are both details of registrations.

solution would be something like:
```
rows = self.remove_unused_details_fields(self.row_fields, "Anders, namelijk")
return self.remove_unused_details_fields(self.row_fields, "fysiodinges", "registration_kinds")
I can probably work around or but if this were an and it will be harder.
```

I have opted to not refactor this as the chance of this getting out of control is quite high. I do consider 
`self.remove_unused_details_fields(self.row_fields, True)` finished unless you think otherwise.

proposals/utils/pdf_diff_utils.py
```diff
@@ -66,6 +66,19 @@ def render(self, context):
     def get_sub_title(self, obj, sub_title_type):
         return SubTitle(obj, sub_title_type).sub_title
 
+    def remove_unused_details_fields(
```
RoordinkFrank, line 69: The single condition argument makes this reuseable in three methods.
It now looks like rows = self.remove_unused_details_fields(self.row_fields, True).
Also can handle things like "N". No need to check for isinstance.

proposals/utils/pdf_diff_sections.py
EdoStorm96: No I think this is a good call!

proposals/forms.py
```diff
@@ -762,27 +762,32 @@ class ProposalDataManagementForm(SoftValidationMixin, ConditionalModelForm):
...
     def clean(self):
         cleaned_data = super(ProposalDataManagementForm, self).clean()
 
-        for field in self.fields:
-            if cleaned_data[field] is None:
-                self.add_error(
-                    field,
-                    _("Dit veld is verplicht om verder te gaan."),
-                )
+        for field in self.base_fields:
```
RoordinkFrank, line 766: code from:

Image

When adding the soft_validation_required you actually get the double field required bug from issue #966. I can see why soft_validation_required is needed as the field needs to be nullable for the pre-assessment? In any case I am not entirely sure where the first required comes from.

proposals/utils/pdf_diff_sections.py
```diff
@@ -626,12 +609,18 @@ class DMPSection(PageBreakMixin, BaseSection):
...
+    row_fields: list[str] = [
         "privacy_officer_conversation",
+        "privacy_officer_conversation_details",
         "data_manager_conversation",
+        "data_manager_conversation_details",
         "research_data_management_conversation",
+        "research_data_management_conversation_details",
     ]
 
+    def get_row_fields(self):
```
RoordinkFrank: todo: make this an array if I have the time.

proposals/utils/pdf_diff_utils.py
EdoStorm96: Came out very nicely!

proposals/utils/pdf_diff_utils.py
```diff
@@ -66,6 +66,19 @@ def render(self, context):
...
+        self, row_fields: list[str], condition
+    ) -> list[str]:
+        """
+        assumes all conditional fields end with '_details'.
+        @condition: when True remove the details field
+        """
+        rows_to_remove: list[str] = []
+        for row_field in row_fields:
+            if getattr(self.obj, row_field) == condition:
```
EdoStorm96: If make sure the condition is always within a list, you can check for multiple conditions at once like so:
```
if getattr(self.obj, row_field) in condition_list:
```

proposals/utils/pdf_diff_sections.py
```diff
@@ -626,12 +609,17 @@ class DMPSection(PageBreakMixin, BaseSection):
...
-        "research_data_management_conversation",
+    row_fields: list[str] = [
+        "privacy_choice",
+        "privacy_choice_details",
+        "DMPChoice",
+        "DMPChoice_details",
     ]
 
+    def get_row_fields(self) -> list[str]:
+        rows: list[str] = self.remove_unused_details_fields(self.row_fields, 0)
```
EdoStorm96: Doing this twice can be done away with, if you make sure the function accepts conditions as a list. See my other comment.

proposals/models.py
```diff
@@ -474,39 +474,54 @@ def date_start_within_two_weeks(self):
...
+    class DMPChoices(models.IntegerChoices):
+        HUMANITIES_CONVERSATION = 0, _(
+            "Het Data Management Plan (DMP) dat ik in de volgende stap zal indienen is besproken met en gezien door de datamanager van de faculteit Geesteswetenschappen."
+        )
+        RESEARCH_DATA_MANAGEMENT_CONVERSATION = 1, _(
+            "Het Data Management Plan (DMP) dat ik in de volgende stap zal indienen is besproken met en gezien door iemand van Research Data Management Support."
+        )
+        OTHERWISE = 2, _("Anders (licht s.v.p. toe)")
+
+    DMPChoice = models.PositiveIntegerField(
```
EdoStorm96: I would prefer `dmp_choice` for the field name.

proposals/models.py
```diff
@@ -474,39 +474,54 @@ def date_start_within_two_weeks(self):
         blank=True,
     )
 
-    privacy_officer_conversation = models.BooleanField(
-        _(
-            "Ik heb mijn aanvraag en de documenten voor deelnemers besproken met de privacy officer."
-        ),
-        default=None,
+    class PrivacyChoices(models.IntegerChoices):
+        PRIVACY_CONVERSATION = 0, _(
```
EdoStorm96, line 478: I think it would be better to use StringChoices here. Then the actual value will still be somewhat legible.

locale/en/LC_MESSAGES/django.po
```diff
@@ -2021,39 +2021,65 @@ msgstr ""
...
-"Ik heb mijn Data Management Plan (DMP) besproken met de data manager van de "
-"faculteit Geesteswetenschappen."
+"Het Data Management Plan (DMP) dat ik in de volgende stap zal indienen is "
+"besproken met en gezien door de datamanager van de faculteit "
+"Geesteswetenschappen."
 msgstr ""
-"I discussed my application and Data Management Plan (DMP) with the data "
-"manager of faculty of Humanities."
+"The Data Management Plan (DMP) that will be submitted in the next step has "
+"been discussed with and shown to the manager of Faculty of Humanities."
```
EdoStorm96:
```suggestion
"been discussed with and shown to the datamanager of Faculty of Humanities."
```

EdoStorm96 (changes requested):
It's looking very nice! I just have a few comments :)

proposals/models.py
RoordinkFrank: Will change.

proposals/utils/pdf_diff_utils.py
RoordinkFrank: This should be possible yes.

PR #1016, opened 2025-11-12
Pre-Assessment tests

RoordinkFrank:
Wants to merge into Feat/Pre-assessment_get_revisions. These tests were originally part of the pullrequest Feat/faster tests #1007 

This pullrequest adds pre-assessment tests for review in the same way the test for proposal were added. Some very samll qol on assertions as well.

------------------
After this pull request is merged #1010 and #1011 will have merge problems. If these two were reviewed before this merge they will have to be reviewed. #1009 is fine to review

------------------
Feat/faster tests now wants to merge into this branch. That way I can see if I forgot any tests is what i want to say but the diff did not get updated when changing the branch you want to merge into so that was quite useless.


reviews/tests.py
```diff
@@ -275,7 +312,13 @@ def setUp(self):
...
         )
 
+    def test_auto_review_pre_assessment_short_route(self):
+        reasons = auto_review(self.proposal)
+        self.assertEqual(len(reasons), 0)
+
     def test_auto_review(self):
+        if self.proposal.is_pre_assessment:
+            return
         reasons = auto_review(self.proposal)
```
RoordinkFrank, line 325: Copied from Feat/Faster tests:

return statement is not great here as the pre assessment tests still does a whole setup for these tests that are basically immediately returned as these tests are only meant for base proposal.

Only solution I can think of is by letting the base proposal inherit in the same way as pre assessment now inherits AutoReviewTests. And adding the normal proposal specific tests there.

reviews/tests.py
```diff
@@ -409,6 +474,9 @@ class ReviewCloseTestCase(
 
     def setUp(self):
         super().setUp()
+        self.startReview()
+
```
RoordinkFrank, line 497: Copied from Feat/faster tests:

setUp BaseReviewTestCase -> setUp ReviewCloseTestCase -> setUp PreAssessmentReviewCloseTestCase is the order that triggers on the setUp inheritance and it has to remain this way. Yet I do want BaseReviewTestCase -> PreAssessmentReviewCloseTestCase -> ReviewCloseTestCase which is impossible in current setup so i have to override startReview here.

Possible solution is the same as in one of my previous comments about proposal inheriting the same was as preAssessment is doing now.

reviews/tests.py
```diff
@@ -76,6 +79,16 @@ def setup_proposal(self):
         )
         self.proposal.generate_pdf()
 
+        self.pre_assessment = Proposal.objects.get(pk=2)
```
RoordinkFrank: Loads in an existing fixture and adds data the fixture is missing.

This does mean that more then just Proposal.objects.get(pk=2) is loaded in as all the current proposal fixtures are loaded in. This does not affect the logic but  loading in fixtures you are not being used is not ideal as it makes things slightly slower. Something that gets fixed in other pull requests that are already on their way.

edit: Looking at the test time it actually greatly slows things down.

reviews/tests.py
```diff
@@ -76,6 +79,16 @@ def setup_proposal(self):
...
+        self.pre_assessment.applicants.add(self.user)
+        self.pre_assessment.wmo = Wmo.objects.create(
+            proposal=self.pre_assessment,
+            metc=YesNoDoubt.NO,
+        )
+        self.pre_assessment.supervisor = self.supervisor
+        self.pre_assessment.relation = Relation.objects.get(pk=4)
```
miggol: Try to avoid just setting an attribute by PK like this, it tells the future programmer nothing. A comment indicating which relation we're getting can help, but even better would be:

```python3
self.phd_student_relation = Relation.objects.get(pk=4)
self.pre_assessment.relation = self.phd_student_relation
```

This is informative without needing a comment, and allows future tests to reuse this variable.

reviews/tests.py
miggol: I'm a bit confused by this line in general. When would `self.proposal` ever be a preassessment? That if statement should never be true unless we specifically set a preassessment. So for now, that should only happen `PreassessmentReviewCloseTestCase`.

Could you remove this from the other test cases?

reviews/tests.py
```diff
@@ -476,6 +544,8 @@ def test_decision(self):
         )
 
     def test_long_route(self):
+        if self.proposal.is_pre_assessment:  # pre-assessment has no long route
+            return
```
miggol: So I can see why this is necessary in this case, but I would much rather you disabled this test for preassessments like this:

```python3
class PreAssessmentReviewCloseTestCase(ReviewCloseTestCase):

    ...

    def test_long_route(self):
        pass
```

In this way we can keep things relating to preassessments within the relevant testcase. No need to mess with the base class!

reviews/tests.py
```diff
@@ -493,6 +563,7 @@ def test_long_route(self):
         self.assertEqual(
             self.review.stage,
             self.review.Stages.CLOSED,
+            f"'{self.review.get_stage_display()}' does not match",
```
miggol: This error message isn't very good. Better to say something like "Review stage is not closed"

When this happens, we will otherwise get something along the lines of "Afsluiting door secretaris does not match", which is a bit confusing.

miggol (changes requested):
Thank you so much for breaking this out of the PR skyscraper! Must have taken some work, but really this is the only way to give this a solid review and get it merged. I hope that in the future we'll be seeing more PRs that look like this one.

Still, some minor remarks, but nothing relating to the test sttructure or logic. It will be a relief to have adequate tests for preassessments going forward. And well done in making use of all the existing base classes.

reviews/tests.py
RoordinkFrank: Without an error message you get 1 != 0 or something similar to that. As for the message itself I agree it should be more clear that the stage is supposed to be closed.

reviews/tests.py
RoordinkFrank: This is a good idea.

Edit: Looks way cleaner this way.

reviews/tests.py
RoordinkFrank: After verbal contact with @miggol we have to come to the conclusion to use names/description instead of pk's. Pkś are indeed less likely to change but they have a lot of downsides. The names/description will becomes a constant so that changing the name/description once it needs to change is way easier. This way we will hopefully have both of best worlds.

reviews/tests.py
RoordinkFrank: Edit: I would have to merge #1009, #1010 and #1011 to a single pr if I had kept this a a fixture. So I have made this a create Object

reviews/tests.py
```diff
@@ -563,7 +586,7 @@ def test_long_route(self):
         self.assertEqual(
             self.review.stage,
             self.review.Stages.CLOSED,
-            f"'{self.review.get_stage_display()}' does not match",
+            f"{self.review.get_stage_display()} is supposed to be the CLOSED stage.",
```
miggol: This does not resolve the issue. The error message may now read:

"Afsluiting door secretaris is supposed to be in the CLOSED stage."

Which still doesn't make much sense. If you were to change it to, for example:

```python3
f"Pre-assessment proposal stage is incorrect. We expect CLOSED but instead we got {self.review.get_stage_display()}.",
```

You would get 

"Review stage is incorrect. We expect CLOSED but instead we got Afsluiting door secretaris."

Which is still not ideal, rather we would use `dict(Review.Stages.Choices)[review.stage]` which would provide the string CLOSING in this example.

reviews/tests.py
```diff
@@ -465,6 +460,36 @@ def setUp(self):
             ),
         )
 
+    def test_auto_review_pre_assessment_short_route(self):
+        # pre-assessment is supposed to go short route
+        reasons = auto_review(self.proposal)
+        self.assertEqual(len(reasons), 0)
```
miggol: Why not replace `test_auto_review` with this test?

reviews/tests.py
```diff
@@ -465,6 +460,36 @@ def setUp(self):
...
+        pass
+
+    def test_auto_review_adults_to_short_route(self):
+        # pre-assessment has no age group
+        pass
+
+    def test_auto_review_session_time(self):
+        # pre-assessment actually passes session time if you let this test trough,
+        # most likely because of self.study.has_sessions = True
+        # which can not happen in a real scenario for pre-assessment
```
miggol: I like that you added comments to all these pass statements! However, please keep the future programmer in mind when writing comments. This specific comment is likely to cause more confusion than enlightenment.

miggol (approved):
Alright, this is approved! I would appreciate it if you took a look at my remaining comments though

reviews/tests.py
RoordinkFrank: Fair enough.

reviews/tests.py
miggol: Wait, no, that last bit is incorrect. But you get my gist

reviews/tests.py
RoordinkFrank: Hmm true this was better as a github comment on this pullrequest. I will simply remove the comment here, the fact that it accidentally passes does not really matter for future people.

PR #1018, opened 2025-11-17
Revision now always starts on the first page.

### LINKED ISSUE #995: Revision - start at 'Documents page' intended?
I was helping a researcher submit their revised application and noticed that if you choose 'revision', you immediately arrive at the Documents page. Is that true in all instances of a revision? If so, why is that? Researchers now often forget to adapt the intended start date for instance.

RoordinkFrank:
Fixes #995

### DIFF
--- a/proposals/models.py
+++ b/proposals/models.py
@@ -708,7 +708,7 @@ Als dat wel moet, geef dan hier aan wat de reden is:"
 
     def continue_url(self):
         stepper = self.stepper
-        if self.is_revision: # if revision always start on the first item.
+        if self.is_revision:  # if revision always start on the first item.
             return stepper.items[0].get_url()
         for item in stepper.items:
             if item.get_errors():
--- a/proposals/models.py
+++ b/proposals/models.py
@@ -195,6 +195,12 @@ gegeven worden; de FETC-GW geeft in die gevallen een post-hoc advies."
         null=True,
     )
 
+    def date_start_within_two_weeks(self):
+        if not self.date_start:
+            return False
+        return date.today()+timedelta(days=14) >= self.date_start
+
+
     expected_end_date = models.DateField(
         _(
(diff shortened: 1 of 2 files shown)


EdoStorm96 (changes requested):
Ok, so we had a nice talk about this. This is fine but you and I both don't love this. As I explained to you, a wrong start date does currently not take the user to the first page. I think we should make it work like that, at least for revisions. 

What I would suggest is to refactor the method `check_start_date()` of `ProposalSubmit`. Make this into a method of the Proposal model. Maybe rename it too, cause the current name is a little confusing. Maybe `start_date_within_two_weeks()` Then re-use that in ProposalSubmit and for continue url. I would make line 711 like this then:
```
if self.is_revision and self.start_date_within_two_weeks():
```
and in `ProposalSubmit.get_context_data()` you can do
```
context["start_date_warning"] = self.object.start_date_within_two_weeks()
```
Should work fine I think :) Good luck!

RoordinkFrank:
> Ok, so we had a nice talk about this. This is fine but you and I both don't love this. As I explained to you, a wrong start date does currently not take the user to the first page. I think we should make it work like that, at least for revisions.
> 
> What I would suggest is to refactor the method `check_start_date()` of `ProposalSubmit`. Make this into a method of the Proposal model. Maybe rename it too, cause the current name is a little confusing. Maybe `start_date_within_two_weeks()` Then re-use that in ProposalSubmit and for continue url. I would make line 711 like this then:
> 
> ```
> if self.is_revision and self.start_date_within_two_weeks():
> ```
> 
> and in `ProposalSubmit.get_context_data()` you can do
> 
> ```
> context["start_date_warning"] = self.object.start_date_within_two_weeks()
> ```
> 
> Should work fine I think :) Good luck!

Good to know the method location. I changed it to date_start instead of start_date to match the model but other then that no changes. Also as discussed earlier the continue_url will have an or instead of an and.

EdoStorm96 (approved):

PR #1026, opened 2026-01-30
Fix/greedy elif

miggol:
Fixes the issue with numbers not showing up in the PDF. The `elif` checking for integers added in the latest merge prevents integers from being rendered as normal values (unless relating to privacy)

In retrospect, none of these conditionals should have been an `elif`, so I changed them all.


RoordinkFrank (approved):
The Issue is indeed fixed. The double isinstance of int was the problem. And elif can indeed be removed here as there is a return statement after each if so the elif doesn´t do anything.
Don´t forget black before merging.

